PR #19, opened 2024-07-23
Fixing `CaptureServiceBuilder` interface

NachoEmbrace:
* The `add` and `remove` methods were not returning self which made the class not function like a normal builder.


atreat (approved):
You should add tests for this. Would expect the the same instance is returned

PR #22, opened 2024-07-23
Publish releases using GitHub instead of AWS S3

ArielDemarco:
# Overview
The purpose of this PR is to remove the use of AWS S3 for storing the release bundle (that later could be use e.g. by CocoaPods) and instead use GitHub Releases. It also automates the process of creating a GitHub Release with the specified version tag.

## Other
- Remove the usage of `EmbraceIO-DEV` cocoapods, as we're going to support prereleases using `EmbraceIO` and using specific tag formatting (e.g. `x.y.z-rcN`)
- Divided behavior when the release is "production ready" vs when it is prerelease.
- Now the create release candidate job can be triggered by pushing tags.
- Improved the way we push the podspec to cocoapods (e.g. we check if version already exists, we ignore output that is unnecessary)


.github/workflows/create-release-candidate.yaml
```diff
@@ -237,29 +229,51 @@ jobs:
...
+          if [ "$(gh config get prompt)" = "enabled" ]; then
+            gh config set prompt disabled
+          fi
+
+      - name: Create (Pre) Release
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          gh release create --repo embrace-io/embrace-apple-sdk ${{ needs.rc_version_number.outputs.rc_version }} cocoapods/embrace_${{ needs.rc_version_number.outputs.rc_version }}.zip -t "Release ${{ needs.rc_version_number.outputs.rc_version }}" -p
+
```
vitaliyf, line 269: This of course will fail if release already exists. In go repo, I ended up doing something like:

```
if ! gh release list .. | grep ..; then
  gh release create..
fi
gh release upload --clobber ..
```

but I don't know if you expect the workflow to re-run for same RC version.

vitaliyf (approved):
🎉

.github/workflows/create-release-candidate.yaml
```diff
@@ -237,29 +229,51 @@ jobs:
...
+        run: |
+          if [ "$(gh config get prompt)" = "enabled" ]; then
+            gh config set prompt disabled
+          fi
+
+      - name: Create (Pre) Release
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          gh release create --repo embrace-io/embrace-apple-sdk ${{ needs.rc_version_number.outputs.rc_version }} cocoapods/embrace_${{ needs.rc_version_number.outputs.rc_version }}.zip -t "Release ${{ needs.rc_version_number.outputs.rc_version }}" -p
```
vitaliyf, line 268:
```suggestion
          gh release create --repo embrace-io/embrace-apple-sdk ${{ needs.rc_version_number.outputs.rc_version }} cocoapods/embrace_${{ needs.rc_version_number.outputs.rc_version }}.zip --title "${{ needs.rc_version_number.outputs.rc_version }}" --prerelease
```

.github/workflows/create-release-candidate.yaml
ArielDemarco: I've done a modification to check if it already exists. Check if it works

.github/workflows/create-release-candidate.yaml
ArielDemarco: Seems outdated based on my other commit; gonna change it

.github/workflows/create-release-candidate.yml
```diff
@@ -270,23 +290,30 @@ jobs:
...
+            echo "COCOAPODS_VERSION_EXISTS=true" >> $GITHUB_ENV
           else
-            echo "Version doesn't exist at all, so it shouldn't be removed"
+            echo "COCOAPODS_VERSION_EXISTS=false" >> $GITHUB_ENV
           fi
+
+      - name: Delete and recreate podspec
+        if: ${{ inputs.override_version }} && env.COCOAPODS_VERSION_EXISTS == 'true'
+        run: |
+            echo "y" | pod trunk delete EmbraceIO-DEV $RC_VERSION
```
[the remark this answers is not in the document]
ArielDemarco: You mean, it should be something like this?
```suggestion
            echo "y" | pod trunk delete EmbraceIO-DEV $RC_VERSION || true
```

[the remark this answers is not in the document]
vitaliyf: Right - my `|| echo some message` has the same effect, but will print "some message" if it ever actually happens.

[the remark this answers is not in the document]
ArielDemarco: I see, so this will be the suggested one
```suggestion
            echo "y" | pod trunk delete EmbraceIO-DEV $RC_VERSION || echo "Ignoring HTML Output"
```

.github/workflows/create-release-candidate.yml
```diff
@@ -237,29 +228,58 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          RC_VERSION: ${{ needs.rc_version_number.outputs.rc_version}}
+        run: |
+          if gh release view --repo embrace-io/embrace-apple-sdk $RC_VERSION > /dev/null 2>&1; then
+            echo "Release $RC_VERSION already exists; editing..."
+            gh release upload $RC_VERSION cocoapods/embrace_$RC_VERSION.zip --clobber --repo embrace-io/embrace-apple-sdk
+          else 
+            echo "Creating Release $RC_VERSION in Github"
+            gh release create $RC_VERSION cocoapods/embrace_$RC_VERSION.zip --title "Release $RC_VERSION" --prerelease --repo embrace-io/embrace-apple-sdk
```
vitaliyf, line 273: nitpick - use just the actual version as release name, it looks slightly nicer in github UI:
```suggestion
            gh release create $RC_VERSION cocoapods/embrace_$RC_VERSION.zip --title "$RC_VERSION" --prerelease --repo embrace-io/embrace-apple-sdk
```

.github/workflows/create-release-candidate.yml
```diff
@@ -237,29 +228,58 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          RC_VERSION: ${{ needs.rc_version_number.outputs.rc_version}}
+        run: |
+          if gh release view --repo embrace-io/embrace-apple-sdk $RC_VERSION > /dev/null 2>&1; then
+            echo "Release $RC_VERSION already exists; editing..."
+            gh release upload $RC_VERSION cocoapods/embrace_$RC_VERSION.zip --clobber --repo embrace-io/embrace-apple-sdk
+          else 
+            echo "Creating Release $RC_VERSION in Github"
+            gh release create $RC_VERSION cocoapods/embrace_$RC_VERSION.zip --title "$RC_VERSION" --prerelease --repo embrace-io/embrace-apple-sdk
```
vitaliyf, line 273: I just saw they also have `--generate-notes` which will automatically create release notes based on PRs that were merged since last release.

ArielDemarco: I'm probably having to generate two releases with two different tags, so i'll have to see how to achieve this and which of those two releases is going to have the release notes (aka diff).

atreat (approved):

.github/workflows/create-release-candidate.yml
```diff
@@ -26,45 +31,86 @@ permissions:
...
+            fi
           fi
-
-          if [[ "$RC_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-            echo "Using RC_VERSION of $RC_VERSION"
+          
+          # Production-ready versions should always follow a "definitive" version format:
+          # - xx.yy.zz
+          # But, non-production-ready versions support having pre-release versions, like:
+          # - x.y.z-alpha
```
ArielDemarco, line 81:
```suggestion
```

.github/workflows/create-release-candidate.yml
```diff
@@ -1,12 +1,13 @@
-name: Create Release Candidate (6.x)
+name: Create Release Candidate
```
vitaliyf, line 1:
```suggestion
name: Create Release
```

and also rename this file to `create-release.yml`?

.github/workflows/create-release-candidate.yml
```diff
@@ -1,12 +1,13 @@
...
 
 env:
   ARCHIVE_QUIET_MODE: 1
   AWS_DEFAULT_REGION: "${{ vars.AWS_DEFAULT_REGION }}"
```
vitaliyf, line 5:
```suggestion
```

.github/workflows/create-release-candidate.yml
```diff
@@ -184,124 +230,92 @@ jobs:
...
       COCOAPODS_TRUNK_TOKEN: ${{ secrets.COCOAPODS_TRUNK_TOKEN }}
-      RC_VERSION: ${{ needs.rc_version_number.outputs.rc_version }}
+      RC_VERSION: ${{ needs.extractor.outputs.rc_version }}
+      IS_PRODUCTION_READY: ${{ needs.extractor.outputs.is_production_ready }}
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v4
         with:
+          fetch-depth: 0
           path: embrace-apple-core-internal
```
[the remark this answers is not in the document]
ArielDemarco: Oh, good catch; should be embrace-apple-sdk

.github/workflows/create-release-candidate.yml
ArielDemarco: Yep! Why not

PR #25, opened 2024-07-25
Adds example of session personas to UserInfo view in BrandGame

atreat:
This PR also fixes a bug when removing metadata in the MetadataHandler. 

The `lifespanId` was not being passed in correctly so only `permanent` records could be removed.

### DIFF
--- a/Sources/EmbraceCore/Public/Metadata/MetadataHandler.swift
+++ b/Sources/EmbraceCore/Public/Metadata/MetadataHandler.swift
@@ -66,28 +66,14 @@ public class MetadataHandler: NSObject {
             throw MetadataError.invalidKey("The key length can not be greater than \(Self.maxKeyLength)")
         }
 
-        // get lifespan id
-        var id: String = ""
-
-        // session lifespan
-        if lifespan == .session {
-            guard let sessionId = sessionController?.currentSession?.id.toString else {
-                throw MetadataError.invalidSession("Can't add a session property if there's no active session!")
-            }
-
-            id = sessionId
-
-            // process lifespan
-        } else if lifespan == .process {
-            id = ProcessIdentifier.current.hex
-        }
+        let lifespanContext = try currentContext(for: lifespan.recordLifespan)
 
         let record = try storage.addMetadata(
             key: key,
             value: validateValue(value),
             type: type,
             lifespan: lifespan.recordLifespan,
-            lifespanId: id
+            lifespanId: lifespanContext
         )
 
         if record == nil {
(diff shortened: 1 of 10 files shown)


Sources/EmbraceCore/Public/Metadata/MetadataHandler.swift
```diff
@@ -66,28 +66,14 @@ public class MetadataHandler: NSObject {
...
-        }
+        let lifespanContext = try currentContext(for: lifespan.recordLifespan)
 
         let record = try storage.addMetadata(
             key: key,
             value: validateValue(value),
             type: type,
             lifespan: lifespan.recordLifespan,
-            lifespanId: id
+            lifespanId: lifespanContext
```
NachoEmbrace, line 76:
```suggestion
            lifespanId: lifespanId
```

Sources/EmbraceCore/Public/Metadata/MetadataHandler.swift
```diff
@@ -185,6 +179,22 @@ extension MetadataHandler {
     }
 }
 
+extension MetadataHandler {
+    private func currentContext(for lifespan: MetadataRecordLifespan) throws -> String {
```
NachoEmbrace, line 183:
```suggestion
    private func lifespanId(for lifespan: MetadataRecordLifespan) throws -> String {
```

Sources/EmbraceCore/Public/Metadata/MetadataHandler.swift
```diff
@@ -144,11 +130,19 @@ public class MetadataHandler: NSObject {
...
+    ///  - lifespan: The lifespan of the metadata to remove.
+    ///
+    ///  - Throws: `MetadataError.invalidSession` if a metadata with a `.session` lifespan is removed when there's no active session.
     func remove(key: String, type: MetadataRecordType, lifespan: MetadataLifespan = .session) throws {
         try storage?.removeMetadata(
             key: key,
             type: type,
-            lifespan: lifespan.recordLifespan
+            lifespan: lifespan.recordLifespan,
+            lifespanId: try currentContext(for: lifespan.recordLifespan)
```
NachoEmbrace, line 145:
```suggestion
            lifespanId: try lifespanId(for: lifespan.recordLifespan)
```

NachoEmbrace:
Seems that `lifespanContext` is just the id.
Left some suggestions so we just use `lifespanId` so it's less confusing.

ArielDemarco (approved):
Given that most of the comments are related to naming, i'll proceed to merge this so we can unblock 6.2 release.
However, it'd be good to address the PR comments in another PR

PR #27, opened 2024-07-29
Adding `EmbraceSemantics` module

NachoEmbrace:
* Main purpose of this module is to not duplicate constants through out the code.
* This being a public module would also allow users of the SDK to generate OTel data using Embrace's semantics.
* Removed the usage of constants in tests (we do want hardcoded values there).


ArielDemarco (approved):

atreat:
Like the separation of these constants and how it's organized into well name-spaced constants. It may be useful to differentiate OpenTelemetry semantic conventions from those that are Embrace specific (until we can get those pushed upstream). 

It may be enough to reference these semantic conventions directly in [SemanticAttributes](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/OpenTelemetryApi/Trace/SemanticAttributes.swift) This would let people follow the definition through to the source.

atreat:
@NachoEmbrace added a commit to this branch to make the `EmbraceType` implementation be in one spot. It also allows us to attach these to new telemetry types ( I also added `SpanEventType`).

Hoping this means we can have telemetry specific declarations and interfaces, but a common implementation to what the value of `emb.type` means.

PR #28, opened 2024-07-29
[WIP] Fix: Adding missing custom properties to session span

ArielDemarco:


Sources/EmbraceCore/Session/SessionSpanUtils.swift
```diff
@@ -111,6 +117,17 @@ fileprivate extension SpanPayload {
             ))
         }
 
+        var mutableProperties = properties
+        mutableProperties.removeAll(where: { $0.key.contains("emb.user") })
+        attributeArray.append(
+            contentsOf: mutableProperties.map {
```
atreat, line 123: Would recommend using `compactMap` here, then inside guard against the `emb.user` prefix and return `nil`.  Will remove an iteration through the loop.

atreat: Actually, would best read using the `filter` method.

ArielDemarco: done!

atreat (approved):

ArielDemarco:
Merging; Test didn't fail; failed the parsing and upload of code coverage

PR #29, opened 2024-07-30
Adds check for activeSpan's context in EmbraceLogRecordBuilder

### LINKED ISSUE #16: Log record missing spancontext when using OpenTelemetry exporter
I've been trying out the Embrace iOS SDK 6.1.1 and seem to have run into a problem with the log API not saving the active span context to the LogRecord. When I use the Otel log exporter StdoutLogExporter, the spancontext is nil.

What I'm calling after setup/start:
```
Embrace.client?.log("log message", severity: LogSeverity.info)
```

When looking at the code, spanContext doesn't seem to be set by default:
https://github.com/embrace-io/embrace-apple-sdk/blob/1eccfb1c2ceb835e11ac978cc30f34262d4bef85/Sources/EmbraceOTelInternal/Logs/EmbraceLogRecordBuilder.swift#L62

I haven't been able to find where `spanContext` is actually set.

If you look at the OTel implementation, they check for nil spanContext and retrieve it from the active span context:
https://github.com/open-telemetry/opentelemetry-swift/blob/7bad8ae7f230e7a1b9ec697f36dcae91a8debff9/Sources/OpenTelemetrySdk/Logs/LogRecordBuilderSdk.swift#L74

Any help here would be great :)

atreat:
If activeSpan is set will be included in LogRecord

Fixes #16

### DIFF
--- a/Sources/EmbraceOTelInternal/Logs/EmbraceLogRecordBuilder.swift
+++ b/Sources/EmbraceOTelInternal/Logs/EmbraceLogRecordBuilder.swift
@@ -57,6 +57,11 @@ class EmbraceLogRecordBuilder: EventBuilder {
 
     func emit() {
         let resource = sharedState.resourceProvider.getResource()
+
+        if spanContext == nil {
+            spanContext = OpenTelemetry.instance.contextProvider.activeSpan?.context
+        }
+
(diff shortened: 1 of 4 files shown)


ArielDemarco (approved):
LGTM!

PR #32, opened 2024-08-05
Adds `W3C.traceparent` helper methods

atreat:
These helper methods are intended to be used when manually instrumenting network requests.


ArielDemarco (approved):
This means there's no need to expose `Config`/features throught the public API, isn't it?

PR #39, opened 2024-08-08
Adds ability to initialize SDK without Embrace AppId

NachoEmbrace:
This PR adds a new setup method that doesn't take an appId.
The goal of this setup method is to have the SDK running without it interacting with Embrace's back end.
For this reason this setup method also doesn't take endpoints definitions and it requires that the implementer passes a valid instance of `OpenTelemetryExport`.

This results in the `EmbraceUpload` instance being nil as well as the `EmbraceConfig` instance being nil in the `Embrace` client.

Flagging this as a draft/PoC for now.

### DIFF
--- a/Sources/EmbraceCore/Embrace.swift
+++ b/Sources/EmbraceCore/Embrace.swift
@@ -110,8 +110,7 @@ To start the SDK you first need to configure it using an `Embrace.Options` insta
                 return client
             }
 
-            try options.validateAppId()
-            try options.validateGroupId()
+            try options.validate()
 
             client = try Embrace(options: options)
             if let client = client {
--- a/Sources/EmbraceCore/Embrace.swift
+++ b/Sources/EmbraceCore/Embrace.swift
@@ -60,7 +60,7 @@ To start the SDK you first need to configure it using an `Embrace.Options` insta
     /// Returns the current `MetadataHandler` used to store resources and session properties.
     public let metadata: MetadataHandler
 
-    let config: EmbraceConfig
+    let config: EmbraceConfig?
     let storage: EmbraceStorage
     let upload: EmbraceUpload?
     let captureServices: CaptureServices
@@ -189,7 +189,7 @@ To start the SDK you first need to configure it using an `Embrace.Options` insta
                 return
             }
 
-            guard config.isSDKEnabled else {
+            guard config == nil || config?.isSDKEnabled == true else {
                 Embrace.logger.warning("Embrace can't start when disabled!")
                 return
             }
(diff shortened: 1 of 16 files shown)


Sources/EmbraceCore/Embrace.swift
```diff
@@ -60,7 +60,7 @@ To start the SDK you first need to configure it using an `Embrace.Options` insta
     /// Returns the current `MetadataHandler` used to store resources and session properties.
     public let metadata: MetadataHandler
 
-    let config: EmbraceConfig
+    let config: EmbraceConfig?
```
atreat, line 63: Would recommend we keep this as non-optional and implement this as a `protocol` to retrieve the values within the SDK. I'd prefer to have a simple object that lists out the defaults instead of having these default values live at the call site.

I'd like to avoid code like this:

```
if config.isSDKEnabled ?? false {
```

ArielDemarco: Maybe we should provide a way to allow users to inject a config object themselves (this is the same way Android works when no appId is present). This might need some changes, in particular how we define the `EmbraceConfig`.

Sources/EmbraceCore/Options/Embrace+Options.swift
```diff
@@ -54,22 +54,50 @@ extension Embrace {
...
+        ///   - logLevel: The `LogLevel` to use for console logs.
+        ///   - export: `OpenTelemetryExport` object to export telemetry outside of the Embrace backend.
+        @objc public init(
+            platform: Platform = .default,
+            captureServices: [CaptureService],
+            crashReporter: CrashReporter?,
+            logLevel: LogLevel = .default,
+            export: OpenTelemetryExport
+        ) {
+            self.appId = "no_id"
```
atreat, line 75: We need to audit the side effects of this change. I think it may be better to declare `appId` as Optional such that we have a clear understanding of the state of the SDK. 

There are some areas in the SDK that we should rename to update our assumptions. For instance, the `EmbraceFilePathProvider` takes an app_group and app_id to build the file path where data is stored. We should update this parameter to mark it as an opaque storage "partition"

Sources/EmbraceCore/Embrace.swift
atreat: Going to leave the injection of config up for a separate PR. That will require a bit of an overhaul to the structure of `EmbraceConfig`

Sources/EmbraceIO/Options/Options+CaptureService.swift
```diff
@@ -67,6 +67,25 @@ public extension Embrace.Options {
             crashReporter: EmbraceCrashReporter()
         )
     }
+
+    /// Initializer for `Embrace.Options` that does not require an appId but used the
```
ArielDemarco, line 71:
```suggestion
    /// Initializer for `Embrace.Options` that does not require an appId.
    
```

ArielDemarco (approved):

PR #42, opened 2024-08-20
Fixed crash on being with another player proxying `NSURLSession`/`URLSession`

ArielDemarco:
# Overview
This PR addresses an issue that arises when our SDK is initialized either before or after other SDKs that proxy `NSURLSession` / `URLSession` (specifically, those that replace or wrap the delegate provided upon calling NSURLSession.init).

The goal is to detect if this substitution has occurred in order to avoid forwarding calls to this proxy, thus preventing infinite recursion.


Tests/EmbraceCoreTests/TestSupport/TestDoubles/Dummy.swift
```diff
@@ -0,0 +1,67 @@
+//
```
atreat, line 1: Can we change this filename to match `DummyURLSessionInitWithDelegateSwizzler`

Sources/EmbraceCommonInternal/Protocols/SwizzleCache.swift
```diff
@@ -16,22 +16,29 @@ public class SwizzleCache {
     struct OriginalMethod: Hashable {
         let method: Method
         let baseClass: AnyClass
+        let swizzlerClass: String
```
atreat, line 19: Why is this change necessary? Can you document what this String means?

Tests/EmbraceCoreTests/TestSupport/TestDoubles/Dummy.swift
ArielDemarco: Oh yes! Dunno why It was saved as dummy 🤔  so weird!
Gonna change it 👍

Sources/EmbraceCommonInternal/Protocols/SwizzleCache.swift
ArielDemarco: Sure, I can.
The change was needed because the cache is just saving the "BaseClass" (aka `NSURLSession`) and the method that we get from selector. To prevent any kind of problems when multiple classes swizzle the same selector, im also adding the name of the swizzler (for example, `DummyURLSessionInitWithDelegateSwizzler` or `URLSessionInitWithDelegateSwizzler`).

NachoEmbrace (approved):

atreat (approved):

PR #43, opened 2024-08-20
Network Payload Capture

NachoEmbrace:
This is a new feature that allows the users of the SDK to create rules through the admin dashboard to capture data from their app's http requests. The SDK receives these rules through the remote configuration and applies them when necessary.

The data is encrypted in the SDK before being converted to OpenTelemetry logs.
This implementation uses hybrid encryption to secure the data. For this the users will also need to provide an asymmetrical public key for each rule.

Some details:
* Each rule can only be triggered once per session
* Rules must have an expiration date
* Rules can use regular expressions to match certain urls
* Rules can filter requests by http method and response status codes

Technical details:
* This feature was implemented inside the `URLSessionCaptureService`. Didn't make sense to swizzle all the `URLSession` stuff in 2 places.
* A new `NetworkPayloadCaptureRule` struct was added to `EmbraceConfigInternal` that is used to decode the remote configuration object.
* `URLSessionTaskCaptureRule` is the actual implementation of the rules that has an underlying `NetworkPayloadCaptureRule` from the remote config.
* `NetworkPayloadCaptureHandler` is the new class that handles all the logic and lives inside `URLSessionTaskHandler`.
* `EncryptedNetworkPayload` is the struct that handles the format for the encrypted payload that is included in the OTel logs.
* All the encryption logic also lives inside `EncryptedNetworkPayload`.
* It encrypts the entire payload by generating a 256 bits symmetrical key and using the `aes-256-cbc` encryption method provided by `CommonCrypto`.
* This symmetrical key is then encrypted using `RSA.PKCS1` and the public key provided by the user.
* The OTel log contains attributes that defines they type of encryption used for both the payload and the key.


Sources/EmbraceConfigInternal/NetworkPayloadCaptureRule.swift
```diff
@@ -0,0 +1,28 @@
...
+    public let publicKey: String
+
+    public var expirationDate: Date {
+        return Date(timeIntervalSince1970: expiration)
+    }
+
+    enum CodingKeys: String, CodingKey {
+        case id
+        case urlRegex = "url"
+        case statusCodes = "status_code"
```
atreat, line 23: Should the raw string be plural here (`status_codes`)?

Sources/EmbraceCore/Capture/Network/NetworkPayloadCapture/EncryptedNetworkPayload.swift
```diff
@@ -0,0 +1,202 @@
...
+            return nil
+        }
+
+        // generate symmetric key
+        let symmetricKey = SymmetricKey(size: .bits256)
+
+        // encrypt payload using aes
+        var encryptedData: Data?
+        do {
+            encryptedData = try AES.GCM.seal(data, using: symmetricKey).combined
```
atreat, line 133: I think this will read a little better if you move the logic to get the `data` var into this do/catch block for `encryptedData`

Sources/EmbraceCore/Capture/Network/NetworkPayloadCapture/EncryptedNetworkPayload.swift
```diff
@@ -0,0 +1,202 @@
...
+struct EncryptedNetworkPayload: Encodable {
+
+    let url: String
+    let httpMethod: String
+
+    let startTime: Int?
+    let endTime: Int?
+
+    let matchedUrl: String
+    let sessionId: String?
```
atreat, line 18: Would let this type be `SessionIdentifier` instead of `String`

Sources/EmbraceCore/Capture/Network/NetworkPayloadCapture/URLSessionTaskCaptureRule.swift
```diff
@@ -0,0 +1,106 @@
...
+                return false
+            }
+        }
+
+        // check status codes
+        if let statusCodes = rule.statusCodes {
+
+            // if -1 is passed as a status code for the rule
+            // it means we should capture any requests that errors
+            if statusCodes.contains(-1) && error != nil {
```
atreat, line 70: Is the `error` here what URLSession gives back? If so, it's likely that there is nothing to capture. These errors mean that no connection to the server was made, or the network request never even left the device. There isn't much an application developer can do when these errors are received, other than including retry logic.

For instance, if a `404` status code is returned there would not be an `error`. Is it intended for a value of `-1` be used to capture this situation?

Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
```diff
@@ -121,14 +127,30 @@ final class DefaultURLSessionTaskHandler: URLSessionTaskHandler {
 
     func finish(task: URLSessionTask, data: Data?, error: (any Error)?) {
         queue.async {
+            // save end time for payload capture
+            task.embraceEndTime = Date()
+
+            // process payload capture
+            self.payloadCaptureHandler.process(
```
atreat, line 134: I wonder if we should pass in the OTel SpanContext to this captured payload to link directly to the span we recorded?

If we do that the startTime/endTime seems redundant

Sources/EmbraceCore/Capture/Network/NetworkPayloadCapture/URLSessionTaskCaptureRule.swift
NachoEmbrace: This was a definition from product:
"-1 is used to capture errors that prevented the requests from leaving the device"

ArielDemarco (approved):
PR LGTM.
One observation is that I would document specific parts of the code as we might forget about the inner details of this feature in the future. In particular, related to the encryption mechanisms. For example the expected values & formats on `EncryptedPayloadResult` (e.g. `iv` in hex), the expected format of the public key `RSA.createKey(for publicKey: String)`, the usage of the `hybrid` key as `mechanism`, etc.

NachoEmbrace:
> PR LGTM. One observation is that I would document specific parts of the code as we might forget about the inner details of this feature in the future. In particular, related to the encryption mechanisms. For example the expected values & formats on `EncryptedPayloadResult` (e.g. `iv` in hex), the expected format of the public key `RSA.createKey(for publicKey: String)`, the usage of the `hybrid` key as `mechanism`, etc.

Comment added in `EncryptedNetworkPayload`.

PR #46, opened 2024-08-21
Fix recordCompletedSpan to set Span.Status consistently with other `end` methods

atreat:
Updates recordCompletedSpan to use the `end(errorCode:time:)` method and properly set `Span.Status`

Status will be `OK` if `errorCode` is `nil`, otherwise it will be `ERROR`.

Fixes duplication of `SpanErrorCode` and `ErrorCode` definitions

### DIFF
--- a/Sources/EmbraceCore/Public/Embrace+OTel.swift
+++ b/Sources/EmbraceCore/Public/Embrace+OTel.swift
@@ -60,6 +60,7 @@ extension Embrace: EmbraceOpenTelemetry {
         let builder = otel
             .buildSpan(name: name, type: type, attributes: attributes)
             .setStartTime(time: startTime)
+
         if let parent = parent { builder.setParent(parent) }
         let span = builder.startSpan()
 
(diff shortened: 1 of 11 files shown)


ArielDemarco (approved):
PR LGTM. 
One question: if somebody creates the span using the `SpanBuilder` (aka `buildSpan` public API), would they be able to use the `end(errorCode:time:)` or shall we suggest they set the `status` manually before ending?

NachoEmbrace (approved):

atreat:
> if somebody creates the span using the SpanBuilder (aka buildSpan public API), would they be able to use the end(errorCode:time:) w/o importing EmbraceOTelInternal or shall we suggest they set the status manually before ending?

I would not expect a caller to need to import `EmbraceOTelInternal`. I'll make sure that is not required. We can also look into setting the status in the SpanProcessor instead to move this logic from the interface layer

atreat:
> I would not expect a caller to need to import EmbraceOTelInternal

This is necessary right now. Going to see what I can do to prevent this from being necessary

ArielDemarco:
> > I would not expect a caller to need to import EmbraceOTelInternal
> 
> 
> 
> This is necessary right now. Going to see what I can do to prevent this from being necessary

It's good to know. If you see the solution would take longer, we can address it in another PR

atreat:
Going to address the `EmbraceOtelInternal` import requirement in another PR.

PR #48, opened 2024-08-23
Updated GRDB to latest version

ArielDemarco:
# Overview
This PR updates `GRDB` to use version `6.29.1`. This is needed to fix a crash on GRDB for xcode version 16.

>[!NOTE]
> For more information around this, [here's the reported issue](https://github.com/groue/GRDB.swift/issues/1588).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -36,7 +36,7 @@ let package = Package(
         ),
         .package(
             url: "https://github.com/groue/GRDB.swift",
-            exact: "6.16.0"
+            exact: "6.29.1"
         ),
         .package(
             url: "https://github.com/realm/SwiftLint",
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -36,7 +36,7 @@ let package = Package(
         ),
         .package(
             url: "https://github.com/groue/GRDB.swift",
-            exact: "6.16.0"
+            exact: "6.29.1"
```
atreat, line 39: We should update this to allow for a range of versions. Would start small with anything in `6.29.X`. 

This does open up the possibility of bugs occurring on specific versions of GRDB, so we may want to look into retrieving the finalized version when an application is built

ArielDemarco:
It might be better to make the decision regarding this PR when Xcode 16 is no longer in beta. Additionally, there was a discussion about allowing the minor version not to be fixed, which would require changes in our CI (when creating the xcframeworks). Given that this context is not sufficiently clear at the moment, I am closing this PR.

atreat (approved):

PR #49, opened 2024-08-26
Fixing async/await network capture

NachoEmbrace:
Not sure when exactly we broke the logic but the fix was very simple.
The issue was we were not setting the proxy delegate so we were not getting the callback when the request finished.

### DIFF
--- a/Sources/EmbraceCore/Capture/Network/URLSessionCaptureService.swift
+++ b/Sources/EmbraceCore/Capture/Network/URLSessionCaptureService.swift
@@ -138,7 +138,7 @@ struct SessionTaskResumeSwizzler: URLSessionSwizzler {
                 return { [weak handler = self.handler] task in
                     let handled = handler?.create(task: task) ?? true
 
-                    // if the task wasn't captured by other swizzlers
+                    // if the task was handled by this swizzler
                     // by the time resume was called it probably means
                     // it was an async/await task
                     // we set a proxy delegate to get a callback when the task finishes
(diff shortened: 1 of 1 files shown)


Sources/EmbraceCore/Capture/Network/URLSessionCaptureService.swift
```diff
@@ -136,13 +136,13 @@ struct SessionTaskResumeSwizzler: URLSessionSwizzler {
         if #available(iOS 15.0, tvOS 15.0, macOS 12, watchOS 8, *) {
             try swizzleInstanceMethod { originalImplementation -> BlockImplementationType in
                 return { [weak handler = self.handler] task in
-                    let captured = handler?.create(task: task) ?? true
+                    let handled = handler?.create(task: task) ?? true
 
                     // if the task wasn't captured by other swizzlers
```
atreat, line 141: Does this comment need to be updated? It looks like it references the `!captured` that was flipped.

NachoEmbrace: Comment still makes sense

atreat:
Do we have any tests around async/await network requests? If not, we should add some to make sure this works as intended

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService.swift
atreat: Can we flip the language to match the updated code?

```suggestion
                    // if the task was handled by this swizzler
```

atreat (approved):

PR #50, opened 2024-09-03
New Public API: Append crash info

ArielDemarco:
# Overview
The goal of this PR is to add functionality to extend crash reports by adding extra information.

## Description
A new public API has been added: `appendCrashInfo(key: String, value: String) throws`. This API will allow supported crashes (aka. those that implement `ExtendableCrashReporter`) to add _extra_ associated information to the next crash that occurs.

## Extra
Both **BrandGame** and **BombApp** were modified to test this functionality. For **BombApp**, to avoid exposing credentials, the `GoogleService-Info.plist` was completely removed. Everyone will need to use their own locally when wanting to use this demo app. If not provided, app _wont build_ (note: even though in the past the app was able to build correctly, it crashed at runtime as we don't submit the api_key)


Sources/EmbraceCrash/EmbraceCrashReporter.swift
```diff
@@ -49,10 +55,16 @@ public final class EmbraceCrashReporter: NSObject, CrashReporter {
             return
         }
 
-        ksCrash.userInfo = [
-            UserInfoKey.sdkVersion: sdkVersion ?? NSNull(),
-            UserInfoKey.sessionId: currentSessionId ?? NSNull()
-        ]
+        if ksCrash.userInfo == nil {
```
atreat, line 58: May want to build up a temporary dictionary here and then set that on `ksCrash.userInfo` in one go. Hoping to make this interaction more atomic in case a crash occurs when this is occurring.

```swift
var info = extraInfo
info[UserInfoKey.sdkVersion] = sdkVersion ?? NSNull()
info[UserInfoKey.sessionId] = currentSessionId ?? NSNull()

ksCrash.userInfo = info
```

Sources/EmbraceCore/Public/Embrace+CrashHelper.swift
```diff
@@ -18,4 +50,26 @@ extension Embrace {
...
+    /// This method allows the addition of a key-value pair as an attribute to the next occurring
+    /// crash that is reported during the lifetime of the process. This can be useful for adding context
+    /// or debugging information that may help in analyzing the crash when exported.
+    ///
+    /// - Important: Throws an exception if no proper `CrashReporter` is configured.
+    ///
+    /// - Parameters:
+    ///   - key: The key for the attribute.
+    ///   - value: The value associated with the given key.
+    @objc public func appendCrashInfo(key: String, value: String) throws {
```
atreat, line 64: Would pull this and the error class out into a different file to separate it from the integration helper `crash(type:)`.

Sources/EmbraceCore/Public/Embrace+CrashReporter.swift
```diff
@@ -0,0 +1,74 @@
...
+
+        guard let extendableCrashReporter = crashRporter as? ExtendableCrashReporter else {
+            throw EmbraceCrashReportError.noCrashReporterAvailable
+        }
+
+        extendableCrashReporter.appendCrashInfo(key: key, value: value)
+    }
+
+    /// Returns the last run end state.
+    @objc public func lastRunEndState() -> LastRunEndState {
```
atreat, line 66: :+1: for moving this in here. Been meaning to organize this

Sources/EmbraceCrash/EmbraceCrashReporter.swift
```diff
@@ -55,16 +55,20 @@ public final class EmbraceCrashReporter: NSObject, CrashReporter {
...
+        var crashInfo: [AnyHashable: Any] = [:]
+
+        if ksCrash.userInfo != nil {
+            crashInfo = ksCrash.userInfo
         }
 
-        ksCrash.userInfo[UserInfoKey.sdkVersion] = self.sdkVersion ?? NSNull()
-        ksCrash.userInfo[UserInfoKey.sessionId] = self.currentSessionId ?? NSNull()
+        crashInfo[UserInfoKey.sdkVersion] = self.sdkVersion ?? NSNull()
+        crashInfo[UserInfoKey.sessionId] = self.currentSessionId ?? NSNull()
```
atreat, line 65: Can we move these to after the extraInfo values are set? Would want these keys to not be overridable

Sources/EmbraceCrash/EmbraceCrashReporter.swift
```diff
@@ -55,16 +55,20 @@ public final class EmbraceCrashReporter: NSObject, CrashReporter {
             return
         }
 
-        if ksCrash.userInfo == nil {
-            ksCrash.userInfo = [:]
+        var crashInfo: [AnyHashable: Any] = [:]
+
+        if ksCrash.userInfo != nil {
+            crashInfo = ksCrash.userInfo
```
atreat, line 61: What is this line trying to achieve? Is this to defend against some other addition to the ksCrash.userInfo outside of this class?

If necessary, this would be shorthand:
```
var crashInfo = ksCrash.userInfo ?? [:]
```

Sources/EmbraceCrash/EmbraceCrashReporter.swift
ArielDemarco: This is the way it did behave in the past. Maybe, we can split `sdkVersion` and `currentSessionId` to their own methods instead

atreat: No I just mean I don't want anyone to be able to call `Embrace.client.appendCrashInfo` and override the sdkVersion or sessionId.

This is still possible if you reach directly into KSCrash but hoping that we could prevent accidents.

ArielDemarco: Oh I see; ok, I can do it to prevent this. Plus, i'll add a test to check that possibility

atreat (approved):

PR #51, opened 2024-09-05
Updating OTel dependencies to v1.10.1

NachoEmbrace:
* I removed all of the `typealias` we had in `EmbraceOTelInternal` in the process as well.
* `SpanData` did not change at all so this didn't require any kind of changes / migrations on our database.


Examples/BrandGame/BrandGame/OpenTelemetry/Export/Log/StdoutLogExporter.swift
```diff
@@ -40,10 +40,10 @@ class StdoutLogExporter: LogRecordExporter {
         return .success
     }
 
-    func forceFlush() -> OpenTelemetrySdk.ExportResult {
+    func forceFlush(explicitTimeout: TimeInterval?) -> OpenTelemetrySdk.ExportResult {
         return .success
     }
 
-    func shutdown() { }
+    func shutdown(explicitTimeout: TimeInterval?) { }
```
ArielDemarco, line 47: This was done because `LogRecordExporter` changed?
Probably this is something we should communicate in the release notes of `6.5.0`. Mostly because this changes that were necessary to be done on the BrandGame app, will probably be also necessary for any customer using a custom `LogRecordExporter`

Sources/EmbraceOTelInternal/Logs/EmbraceLogRecordBuilder.swift
```diff
@@ -55,6 +56,10 @@ class EmbraceLogRecordBuilder: EventBuilder {
         return self
     }
 
+    func setData(_ attributes: [String: OpenTelemetryApi.AttributeValue]) -> Self {
+        return setAttributes(attributes)
+    }
+
```
ArielDemarco, line 62: Seems like implementing this right now is kinda useless. Probably this is needed because when I've created the `EmbraceLogRecordBuilder` I've made the class conformant to `EventBuilder` instead of implementing directly `LogRecordBuilder` (note: `EventBuilder` inherits from `LogRecordBuilder`.
I think we could be able to get rid of it

ArielDemarco (approved):
This will require changing also the reference to the OTel library in `Tuist/Package.swift` and `Tuist/Package.resolved`. I'll try to fix that inconsistency between the _real_ `Package.swift`/`Package.resolved` and the one that `Tuist` uses in another PR.

In the meantime, this change is necessary so we don't have issues like [this one](https://github.com/embrace-io/embrace-apple-sdk/pull/57)

Sources/EmbraceOTelInternal/Logs/EmbraceLogRecordBuilder.swift
NachoEmbrace: What would we return in `EmbraceLogger` `eventBuilder(name: String)` then?

ArielDemarco: Mmm I see why we did it then 😂 . You're right. We should leave it as is then.
Changing the `EmbraceLogger.eventBuilder(name: String)` to use other thing (like the default `LogRecordBuilderSdk`) would require other changes and seems that is not worth it.

PR #65, opened 2024-09-18
Updates EmbraceSemantics to move definitions under `SpanEventType`

atreat:
These were incorrectly defined under `SpanType`. They are created/added to items that we record as SpanEvents.


ArielDemarco:
Would this PR generate any breaking changes on the public API? If that's the case I'd:
a. Keep both old and new code (marking old as deprecated)
b. Specify this in someway.
I understand `EmbraceSemantics` is meant to be exposed and used publicly.

ArielDemarco (approved):

PR #67, opened 2024-09-23
Fixed crash with third party library `gtm-session-fetcher`

NachoEmbrace:
This is an internal library used by Firebase: [gtm-session-fetcher](https://github.com/google/gtm-session-fetcher/).
They are incorrectly assuming the type of the `URLSession` which leads to a crash due to them calling an internal selector our proxy doesn't / can't implement.

This issue was raised a long time ago by other parties, but they seem pretty dismissive about the problem:
https://github.com/google/gtm-session-fetcher/issues/190 .

This PR addresses the issue by not initializing our `URLSession` proxy when the original delegate comes from this library.
Doing this fixes the crashes like: `*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[EmbraceCore.URLSessionDelegateProxy setFetcher:forTask:]: unrecognized selector sent to instance`

**This will result on Embrace not capturing requests done by this library.**

### DIFF
--- a/Sources/EmbraceCore/Capture/Network/URLSessionCaptureService.swift
+++ b/Sources/EmbraceCore/Capture/Network/URLSessionCaptureService.swift
@@ -102,6 +102,12 @@ struct URLSessionInitWithDelegateSwizzler: URLSessionSwizzler {
         try swizzleClassMethod { originalImplementation -> BlockImplementationType in
             return { urlSession, configuration, delegate, queue -> URLSession in
                 let proxiedDelegate = (delegate != nil) ? delegate : EmbraceDummyURLSessionDelegate()
+
+                // check if we support proxying this type of delegate
+                guard isDelegateSupported(proxiedDelegate) else {
+                    return originalImplementation(urlSession, Self.selector, configuration, delegate, queue)
+                }
+
                 let newDelegate = URLSessionDelegateProxy(originalDelegate: proxiedDelegate, handler: handler)
                 let session = originalImplementation(urlSession, Self.selector, configuration, newDelegate, queue)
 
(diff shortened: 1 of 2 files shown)


Sources/EmbraceCore/Capture/Network/URLSessionCaptureService.swift
```diff
@@ -116,6 +122,32 @@ struct URLSessionInitWithDelegateSwizzler: URLSessionSwizzler {
...
+        // We can't solve this on our side in a clean way so we'll just not
+        // capture any requests from this library until the issue is solved
+        // on their side.
+        //
+        // Library: https://github.com/google/gtm-session-fetcher/
+        // Issue: https://github.com/google/gtm-session-fetcher/issues/190
+        "GTMSessionFetcher"
+    ]
+
+    func isDelegateSupported(_ delegate: URLSessionDelegate?) -> Bool {
```
atreat, line 143: Only concern is that if someone is using the workaround in https://github.com/google/gtm-session-fetcher/issues/190 we will still not be able to support these requests.

An alternative solution would be that we provide the workaround and ask people running into this issue to explicitly make this call before they start Embrace? We'd provide the snipper in some `ThirdPartySupport` target where they could call [`GTMSessionFetcherService. swizzleDelegateDispatcherForFetcher()`](https://github.com/google/gtm-session-fetcher/issues/190#issuecomment-607398226)?

NachoEmbrace: People have reported that the workaround doesn't work, that's why we had to do this in the first place.

ArielDemarco (approved):

PR #69, opened 2024-09-24
Release/6.4.1

NachoEmbrace:
This release contains:
* OTel dependencies bumped to v1.10.1
* Fixed logs not having the resources from the previous session when recovering unsent data during startup.
* Fixed crash with the [gtm-session-fetcher](https://github.com/google/gtm-session-fetcher/) library.
* Updates KSCrash dependency to fix compilation issue that occurs in Xcode 16.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -189,6 +189,7 @@ jobs:
           path: xcframeworks.zip
       
       - name: Tag the release candidate version
+        if: env.IS_PRODUCTION_READY == 'false'
         run: |
           git tag $RC_VERSION
           git push origin $RC_VERSION


bin/dependencies/build_opentelemetry-swift.sh
```diff
@@ -1,6 +1,16 @@
...
 BUILD_DIR=$2
 
+# We replace the `package` keyword because it's not supported by the `create-xcframework` command.
+# We tried using "OTHER_SWIFT_FLAGS" and setting the package and module names, but it ends up creating problems
+# since we are building two frameworks at the same time.
+# Until the `package` keyword is supported directly, we'll replace the `package` keyword with just `public`
+# so the frameworks compile.
+# We realize this exposes something that is not meant to be exposed, we think it's harmless.
+
+grep -rl 'package static' "$REPO_DIR/Sources/" | xargs sed -i '' 's/package static/public static/g'
```
atreat, line 11: We should add a SwiftLint rule that prevents `package` access until we can support it directly. This leads to an inconsistent interface depending on if we distribute the xcframework or we distribute using source. That will lead to confusion in the future.

For now, we should change/commit the code changes that is using the `package` keyword so our interfaces are consistent, even if they're more public than we would hope

atreat: Just realized this is specifically for the `opentelemetry-swift` dependency. How has this worked for previous releases?

NachoEmbrace: This is affecting the OpenTelemetryApi and OpenTelemetrySdk libraries, not Embrace.
The only alternative to this that we could think of was to fork them.

ArielDemarco (approved):

atreat (approved):

PR #72, opened 2024-09-25
Fixing background sessions logic in `SessionController`

NachoEmbrace:
Even though we were parsing the background sessions remote config, we were never actually using it in the session logic.
This PR adds the necessary validations into `SessionController` to drop background sessions if the config is disabled.

There's one special case that we have to deal with when the app is starting. In this scenario we can't know right away if the app is actually starting in the background or not.
Due to this, the cold start session is always created even if the app is in the background and the config is disabled.
Once the session ends it will be dropped unless the session state was updated to foreground.

We also updated the logic for "coldStart" sessions. Due to the changes in iOS 15 where apps can be pre-warmed by the OS, our calculations for the process start time are not accurate. For this reason we can't really know for sure how much time has passed since the process started and the first session is created.
For this reason we will always flag the very first session as the "coldStart" session, regardless on how much time has passed since the process started.

Additionally, this PR adds 2 new resources: process start time and process pre warm.


ArielDemarco (approved):

Sources/EmbraceCore/Capture/OneTimeServices/AppInfoCaptureService.swift
```diff
@@ -61,5 +61,20 @@ class AppInfoCaptureService: ResourceCaptureService {
             key: AppResourceKey.processIdentifier.rawValue,
             value: .string(ProcessIdentifier.current.hex)
         )
+
+        // process start time
+        if let uptime = ProcessMetadata.uptime() {
```
atreat, line 66: This should be `ProcessMetadata.startTime` instead of the `uptime`. We want the absolute time the process started and not the length of time until the AppInfoCaptureService has run.

The `emb-process-launch` span already measures the duration from process launch to the end of the `Embrace.setup` method.

Tests/EmbraceCoreTests/Session/SessionControllerTests.swift
```diff
@@ -89,30 +89,6 @@ final class SessionControllerTests: XCTestCase {
         XCTAssertTrue(controller.currentSession!.coldStart)
     }
 
-    func test_startSession_ifStartAtMatchesAllowedColdStartInterval_marksSessionAsColdStartTrue() throws {
```
atreat, line 95: Can you add tests for the new cold start logic? Should verify that the first session only is marked as `cold_start = true`

Sources/EmbraceCore/Capture/OneTimeServices/AppInfoCaptureService.swift
NachoEmbrace: good catch

atreat (approved):

PR #78, opened 2024-10-01
Fix `iOS 17.5` < simulators with cocoapods

ArielDemarco:
# Overview
Fixed an issue that would cause a crash when running Embrace with devices with versions lower than iOS 17.5 after installing the SDK with Cocoapods.

### DIFF
--- a/Project.swift
+++ b/Project.swift
@@ -9,6 +9,7 @@ let project = Project(
             destinations: .iOS,
             product: .framework,
             bundleId: "com.embraceio.EmbraceIO",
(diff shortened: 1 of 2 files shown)


atreat (approved):

atreat:
Do we need other platforms defined here too?

ArielDemarco:
> Do we need other platforms defined here too?

Yes, we'd have to define something for tvOS at least at some point; but we don't have enough tests in place to really test this

NachoEmbrace (approved):

PR #81, opened 2024-10-03
Adds initial set of storage Benchmarks

atreat:
Initial integration and testing of the ordo-one/package-benchmark package to measure performance

Compare baselines with: 
```bash
swift package benchmark baseline compare alpha
```

See [documentation for Benchmarks](https://swiftpackageindex.com/ordo-one/package-benchmark/1.25.0/documentation/benchmark/creatingandcomparingbaselines)


Package.swift
```diff
@@ -16,7 +16,7 @@ if ProcessInfo.processInfo.environment["IS_XCTEST"] != nil ||
 let package = Package(
     name: "EmbraceIO",
     platforms: [
-        .iOS(.v13), .tvOS(.v13), .macOS(.v13), .watchOS(.v6)
+        .iOS(.v16), .tvOS(.v13), .macOS(.v13), .watchOS(.v6)
```
NachoEmbrace, line 19: I assume this means the Benchmark stuff is only available in iOS 16+?
Not sure we can actually go through with this change...

ArielDemarco:
Looks that this is something we should deprioritize right now, considering the lack of support of Benchmarks for lower iOS versions

PR #83, opened 2024-10-09
Using OpenTelemetry's enums in `EmbraceSemantics`

NachoEmbrace:


ArielDemarco (changes requested):
Add to `Project.swift` the new dependency of `EmbraceSemantics`

ArielDemarco (approved):

PR #87, opened 2024-10-10
Setting status as `error` for crashed session spans

NachoEmbrace:
This affects the session span itself and all the open spans within it.
This does NOT affect spans that had already ended successfully before the crash.

### DIFF
--- a/Sources/EmbraceCore/Payload/Spans/SpanPayload.swift
+++ b/Sources/EmbraceCore/Payload/Spans/SpanPayload.swift
@@ -38,7 +38,7 @@ struct SpanPayload: Encodable {
         self.spanId = span.spanId.hexString
         self.parentSpanId = span.parentSpanId?.hexString
         self.name = span.name
-        self.status = failed ? Status.sessionCrashedError().name : span.status.name
+        self.status = span.status == .ok ? span.status.name : (failed ? Status.sessionCrashedError().name : span.status.name)
         self.startTime = span.startTime.nanosecondsSince1970Truncated
         self.events = span.events.map { SpanEventPayload(from: $0) }
         self.links = span.links.map { SpanLinkPayload(from: $0) }
--- a/Sources/EmbraceCore/Payload/Spans/SpanPayload.swift
+++ b/Sources/EmbraceCore/Payload/Spans/SpanPayload.swift
@@ -5,6 +5,7 @@
 import Foundation
 import EmbraceOTelInternal
 import EmbraceSemantics
+import OpenTelemetryApi
 import OpenTelemetrySdk
 
 struct SpanPayload: Encodable {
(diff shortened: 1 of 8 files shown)


Sources/EmbraceCore/Session/SessionSpanUtils.swift
```diff
@@ -128,3 +128,9 @@ fileprivate extension SpanPayload {
         }
     }
 }
+
+internal extension OpenTelemetryApi.Status {
+    static func sessionCrashedError() -> Status {
```
atreat, line 133: This is not a valid status. Only 3 values are allowed (`Unset`, `Error`, and `Ok`)
https://opentelemetry.io/docs/concepts/signals/traces/#span-status

Sources/EmbraceCore/Payload/Spans/SpanPayload.swift
```diff
@@ -37,7 +38,7 @@ struct SpanPayload: Encodable {
         self.spanId = span.spanId.hexString
         self.parentSpanId = span.parentSpanId?.hexString
         self.name = span.name
-        self.status = span.status.name
+        self.status = span.status == .ok ? span.status.name : (failed ? Status.sessionCrashedError().name : span.status.name)
```
atreat, line 41: This will need to be broken out, 2 tertiary statements on the same line makes me cross-eyed.

Sources/EmbraceCore/Session/SessionSpanUtils.swift
NachoEmbrace: The string passed is just an internal description, the status is still an error (it's `name` will be `error`).
From OpenTelemetryApi:

```
public enum Status: Equatable {
    /// The operation has been validated by an Application developers or Operator to have completed successfully.
    case ok
    /// The default status.
    case unset
    /// The operation contains an error.
    case error(description: String)

    /// True if this Status is OK
    public var isOk: Bool {
        return self == .ok
    }

    /// True if this Status is an Error
    public var isError: Bool {
        if case .error = self {
            return true
        }
        return false
    }

    public var name: String {
        switch self {
            case .ok:
                return "ok"
            case .unset:
                return "unset"
            case .error(description: _):
                return "error"
        }
    }

    public var code: Int {
        switch self {
            case .ok:
                return 0
            case .unset:
                return 1
            case .error(description: _):
                return 2
        }
    }
}

extension Status: CustomStringConvertible {
    public var description: String {
        if case let Status.error(description) = self {
            return "Status{statusCode=\(name), description=\(description)}"
        } else {
            return "Status{statusCode=\(name)}"
        }
    }
}
```

atreat: 👍 misread the `.error` there

ArielDemarco (approved):
LGTM

PR #90, opened 2024-10-15
Updating release process

NachoEmbrace:
Made some updates to the release process to prevent triggering multiple release workflows at the same time and also prevent version discrepancies that could occur.

* Release candidates are still triggered automatically in every push on a release branch
* Production releases will now be triggered manually through GitHub's workflow dispatch using the `Create Release` workflow.
  - This requires us to indicate the branch, final version and set the "production ready" flag to true. All of which can be done when triggering the workflow.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -5,8 +5,6 @@ env:
 
 on:
   push:
-    tags:
-      - '[0-9]+.[0-9]+.[0-9]+'
     branches:
       - "release/*"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release.yml
```diff
@@ -48,17 +46,6 @@ jobs:
...
-      - name: Exctract Production Readiness
-        run: |
-          if [ -z "$IS_PRODUCTION_READY" ]; then
-            if [[ "${{ github.event_name }}" == "push" && "${{ github.ref_type }}" == "tag" ]]; then
-              echo "IS_PRODUCTION_READY=true" >> $GITHUB_ENV
-            else
-              echo "IS_PRODUCTION_READY=false" >> $GITHUB_ENV
-            fi
-          fi
-          echo "Production readiness determined: $IS_PRODUCTION_READY"
```
ArielDemarco, line 60: I think we might need this for latest steps; we should keep that in mind if the rc job fails

ArielDemarco (approved):

PR #101, opened 2024-10-22
Improvements during startup

NachoEmbrace:
* Improves `Embrace.setup` by 10ms-20ms
  - `EmbraceConfig` was unintentionally doing some stuff on the main thread. It is now being dispatched async on a different queue.

* Improves `Embrace.start` by 10ms-20ms
  - I split `CaptureServices.start` into 2 methods. Only the `install` needed to be done on the main thread.

* Fixes the remote config being fetched twice on startup.

### DIFF
--- a/Sources/EmbraceConfigInternal/EmbraceConfig.swift
+++ b/Sources/EmbraceConfigInternal/EmbraceConfig.swift
@@ -20,6 +20,8 @@ public class EmbraceConfig {
 
     let configurable: EmbraceConfigurable
 
+    let queue: DispatchQueue
+
     public init(
         configurable: EmbraceConfigurable,
         options: Options,
@@ -30,6 +32,7 @@ public class EmbraceConfig {
         self.notificationCenter = notificationCenter
         self.logger = logger
         self.configurable = configurable
+        self.queue = DispatchQueue(label: "com.embrace.config", attributes: .concurrent)
 
         update()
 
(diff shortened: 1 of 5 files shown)


ArielDemarco (approved):

Sources/EmbraceConfigInternal/EmbraceConfig.swift
```diff
@@ -54,21 +57,24 @@ public class EmbraceConfig {
...
+        self.queue.async { [weak self] in
+            self?.configurable.update { [weak self] didChange, error in
+                if let error = error {
+                    self?.logger.error(
+                        "Failed update in EmbraceConfig",
+                        attributes: [ "error.message": error.localizedDescription ]
+                    )
+                }
+
+                self?.lastUpdateTime = Date().timeIntervalSince1970
```
atreat, line 73: If `updateIfNeeded` is called for the first time many times in quick succession its possible that we kick off multiple jobs onto this async queue.

Should we guard against this with the `lastUpdateTime` or should we introduce a variable to track that an update is currently in progress?

Sources/EmbraceConfigInternal/EmbraceConfig.swift
```diff
@@ -30,6 +32,7 @@ public class EmbraceConfig {
         self.notificationCenter = notificationCenter
         self.logger = logger
         self.configurable = configurable
+        self.queue = DispatchQueue(label: "com.embrace.config", attributes: .concurrent)
 
         update()
```
atreat, line 37: We should move this call to update as it is a side effect of the `init` method.

Think it makes more sense if we call this in the `Embrace.start` method in the [async job here](https://github.com/embrace-io/embrace-apple-sdk/pull/101/files#diff-b90b0a3a8c473dbc673660227e7cab7328e5201df828de2a1e60c7b2431c4edaR210)

Tests/TestSupport/Mocks/MockEmbraceConfigurable.swift
```diff
@@ -92,11 +92,11 @@ public class MockEmbraceConfigurable: EmbraceConfigurable {
         }
     }
 
-    public let updateExpectation = XCTestExpectation(description: "update called")
+    public var updateCallCount = 0
```
atreat, line 95: Curious why you moved away from the XCTestExpectation. Was this causing test failures?

PR #120, opened 2024-11-07
Delete extra EmbraceObjCUtilsInternal module.modulemap file

lucasmarcal-faire:
## Checklist

Remove extra modulemap file.

Swift Package Manager only searches for modulemap files for library targets inside the `include/` folder.

This file is not being used anywhere and fails to build when using auxiliary tools like `rules_swift_package_manager` that translates SPM target description into Bazel BUILD files

### DIFF
--- a/Sources/EmbraceObjCUtilsInternal/module.modulemap
+++ b/Sources/EmbraceObjCUtilsInternal/module.modulemap
@@ -1,4 +1,0 @@
-module EmbraceObjCUtilsInternal {
-    umbrella "./include"
-    export *
-}


lucasmarcal-faire:
@NachoEmbrace @ArielDemarco

ArielDemarco:
Hey @lucasmarcal-faire! thanks for the contribution! Unfortunately, we cannot remove the `module.modulemap` from `EmbraceObjCUtilsInternal` as it's being used by the `Project.swift` file, which defines our project for `Tuist` during the `.xcframework`s creation process.

To give a bit of context: Generating `.xcframeworks` directly from a `Package.swift` can be nearly impossible in many cases (we invested many hours trying to figure out an easier way to achieve this using the package definition or the underlying `.xcodeproj` that SPM creates). To work around this, we generate an `.xcodeproj` that closely resembles the `Package.swift`. However as you may know, SPM automates certain steps, like generating module maps, which we need to manually replicate using `Tuist`.

That being said, one potential solution is to adjust the script used for creating `.xcframeworks` (you can find it in `./bin/build_xcframeworks`). You could modify it to generate the `module.modulemap` before running both `tuist install` and `tuist generate`, and then either remove it post-process or exclude it by adding the `.modulemap` to `.gitignore`.

lucasmarcal-faire:
Actually, in the `Project.swift` you are using the other one that is placed in the `include/` folder
https://github.com/embrace-io/embrace-apple-sdk/blob/main/Project.swift#L221C55-L221C79

All the other references to the ObjcUtils target to include as a search path are making reference to the include folder

lucasmarcal-faire:
Just in case, I ran the `./bin/build_xcframeworks`, which succeeded with these PR changes.

Screenshot 2024-11-07 at 12 27 59 PM

bstpierr (approved):

ArielDemarco:
> Actually, in the `Project.swift` you are using the other one that is placed in the `include/` folder https://github.com/embrace-io/embrace-apple-sdk/blob/main/Project.swift#L221C55-L221C79
> 
> All the other references to the ObjcUtils target to include as a search path refer to its `include/` folder.
> 
> Or is that something that happens under the hood?

You're absolutely right! I'm not sure why there was duplication, but the module references are indeed as you mentioned.

I cloned the fork and tested the generated `.xcframeworks` in one of our test apps, and everything seems to be functioning perfectly. LGTM!

ArielDemarco:
It seems the tests passed, but for some reason, the `xcresulttool` action (which is then used to upload the coverage) failed. Issue isn't related to the PR and/or changes in it (besides, the tests, as I mentioned, passed). I'll merge it. 
Thanks @lucasmarcal-faire for the contribution!

PR #122, opened 2024-11-08
Fixing network payload rules parsing

NachoEmbrace:
It seems that the remote config API was never updated with the suggested changes for network payload capture.
This resulted in network capture rules not being parsed correctly on this version of the SDK.

This PR contains the necessary changes for us to parse the old payload.


ArielDemarco (approved):
Considering the actual API wasn't updated... is there any value that were considering that could arrive to the SDK and we don't know how to handle? (the only one I remember is the status_code `-1`, but I think we handle that one)

NachoEmbrace:
> Considering the actual API wasn't updated... is there any value that were considering that could arrive to the SDK and we don't know how to handle? (the only one I remember is the status_code `-1`, but I think we handle that one)

The only big difference was the http method being a single string instead of an array.
Other than that there are keys that could be present but will be ignored.
Overall there should be no issues.

PR #126, opened 2024-11-20
Improving instrumentation around `UIViewController`

NachoEmbrace:
This is a refactor on `ViewCaptureService`.

It still retains the old functionality: It creates `emb-screen-view` spans when `viewDidAppear` is called on a `UIViewController`. These spans are ended when `viewDidDisappear` is called.
This feature can be enabled/disabled through `ViewCaptureService.Options.instrumentVisibility`.

Added a new feature to automatically instrument the time-to-render on `UIViewControllers`.
This will create a parent span starting from `viewDidLoad` and ending on `viewDidAppear`.
The parent span will contain child spans for each step of the loading process (`viewDidLoad`. `viewWillAppear` and `viewDidAppear`).
The user can enable/disable this through `ViewCaptureService.Options.instrumetTimeToRender`.

If the `UIViewController` implements the `InteractableViewController` protocol, the service will create the time-to-interactive span instead.
This is a parent span starting from `viewDidLoad` and ending when the view is deemed as ready to be interacted or the `UIViewController` disappears.
The implementers must call `InteractableViewController.onInteractionReady()` to flag the view as ready.
The parent span will contain child spans for each step of the loading process (`viewDidLoad`. `viewWillAppear` and `viewDidAppear`), and an extra `ui-ready` span that starts at the end of `viewDidAppear` and ends when the view is ready for interaction.

Furthermore, if the `UIViewController` implements the `InstrumentableViewController` protocol, it can use the `buildChildSpan` and `recordCompletedChildSpan` methods to add custom child spans to the time-to-render/time-to-interactive parent spans.


Sources/EmbraceCore/Capture/UX/View/UIViewControllerHandler.swift
```diff
@@ -0,0 +1,361 @@
...
+
+    func onViewDidLoadEnd(_ vc: UIViewController) {
+        queue.async {
+            guard let id = vc.emb_identifier,
+                  let span = self.viewDidLoadSpans[id] else {
+                return
+            }
+
+            span.end()
+            self.viewDidLoadSpans[id] = nil
```
ArielDemarco, line 129: Nitpick: to prevent from accessing twice the dictionary (which also has the locking mechanism) maybe you can do something like:
```suggestion
            guard let id = vc.emb_identifier,
                  let span = self.viewDidLoadSpans.removeValue(forKey: id) else {
                return
            }

            span.end()
```

ArielDemarco (approved):

PR #130, opened 2024-11-26
Fixed a bug preventing crashes from having session properties

ArielDemarco:
# Overview
Crash Reports should have the associated custom properties to the session span. However due tu a misconfiguration of the object that creates the attributes on the crash log, this was not happening.
This PR addresses that issue.

### DIFF
--- a/Sources/EmbraceCore/Internal/Logs/DefaultInternalLogger.swift
+++ b/Sources/EmbraceCore/Internal/Logs/DefaultInternalLogger.swift
@@ -126,7 +126,6 @@ class DefaultInternalLogger: InternalLogger {
 
         let attributes = attributesBuilder
             .addLogType(.internal)
-            .addApplicationProperties()
             .addApplicationState()
             .addSessionIdentifier()
             .build()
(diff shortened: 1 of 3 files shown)


Sources/EmbraceCore/Internal/Logs/DefaultInternalLogger.swift
```diff
@@ -126,7 +126,6 @@ class DefaultInternalLogger: InternalLogger {
 
         let attributes = attributesBuilder
             .addLogType(.internal)
-            .addApplicationProperties()
```
NachoEmbrace, line 129: Is this correct?

NachoEmbrace (approved):

Sources/EmbraceCore/Internal/Logs/DefaultInternalLogger.swift
ArielDemarco: Yes, the `addApplicationProperties` method does nothing because the `DefaultInternalLogger` doesn't contain an instance of `EmbraceStorage`. Consequently, an `EmbraceLogAttributesBuilder` is created without `EmbraceStorage`, making the call to `addApplicationProperties` useless. Here is the relevant code inside that method:
```swift
    @discardableResult
    func addApplicationProperties() -> Self {
        guard let sessionId = currentSession?.id,
              let storage = storage else {
            return self
        }
        /* 
           More code...
        */
    }
```
In fact, the only _useful_ method to call in this case is `addLogType`, because the rest of the attributes will be added later. Further down, this method calls `otel.log`, which internally also creates another builder and adds the "default" information to the logs (which also includes application properties, state, etc.).

PR #135, opened 2024-12-04
Upload module improvements

ArielDemarco:
# Overview
This PR addresses some issues and also does some improvements on the way the `UploadModule` works. 

# Details
- Reduced the complexity of `EmbraceUploadCache` to clean/purge it only based on date.
- Added exponential backoff to all retry logics (more info on `ExponentialBackoff` struct).
- Fixed a bug that would prevent from retrying requests.
- Fixed a bug that caused the `attemptCount` to not be properly included in the headers of retried requests
- Improved logic to understand which requests should be retried based on status codes (`429`, `500`-`599`) and errors. . Some requests (like the ones with statusCode `400: Bad Request` or the ones with specific errors like `URLError.badURL`) are going to be dropped. 
- Some improvements were made to the request recovery and retry process (which occurs after a request fails multiple times and is stored in the cache):
   - Queue the different retry operations (up to two) to avoid overwhelming the backend
   - Prevents this process from being executed multiple times (it'll wait until the first "retry" of all requests finishes)
   - `maximumAmountOfRetries` was added to `RedundancyOptions` to prevent requests from retrying indefinitely (which fixes a bug that would use an unnecessary amount of memory and data).
- Added the possibility to use the `Retry-After` header in combination with the Expontential Backoff logic.


ArielDemarco:
All warnings are from Tests. We should remove that :) 
Also... [Run Tests](https://github.com/embrace-io/embrace-apple-sdk/actions/runs/12161106911/job/33914905798?pr=135) seems to have worked properly. Don't know what happened with `Run Danger` that failed (while running tests). 
Merging as everything seems fine

PR #136, opened 2024-12-04
Disabling memory termination feature from KSCrash

NachoEmbrace:
Also bumped KSCrash to v2.0.7 which adds sanity checks for the reports timestamps


naftaly:
Hey @NachoEmbrace I'd love to know what kind of issues you've run into with the KSCrash memory termination feature.

PR #156, opened 2025-01-10
OTel Logs with file attachments

NachoEmbrace:
* Added new public APIs to create logs with attachments (`Embrace+OTel.swift`).
* Moved log creating logic inside of `LogController` (so we can test the logic without needing to initialize the Embrace client).
* Added new type of data to the `EmbraceUpload` module:
  - Uploaded attachments are retried a few times but never cached
  - If the upload fails, the data is discarded and the log is sent with the appropriate attachment error

### DIFF
--- a/Sources/EmbraceCore/Utils/URL+Embrace.swift
+++ b/Sources/EmbraceCore/Utils/URL+Embrace.swift
@@ -18,4 +18,8 @@ extension URL {
     static func logsEndpoint(basePath: String) -> URL? {
         return endpoint(basePath: basePath, apiPath: "/v2/logs")
     }
+
+    static func attachmentsEndpoit(basePath: String) -> URL? {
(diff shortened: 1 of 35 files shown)


Sources/EmbraceCore/Utils/URL+Embrace.swift
```diff
@@ -18,4 +18,8 @@ extension URL {
     static func logsEndpoint(basePath: String) -> URL? {
         return endpoint(basePath: basePath, apiPath: "/v2/logs")
     }
+
+    static func attachmentsEndpoit(basePath: String) -> URL? {
```
ArielDemarco, line 22: Typo: `attachmentsEndpoit` as `attachmentsEndpoint`

ArielDemarco (approved):

ArielDemarco (approved):

PR #159, opened 2025-01-17
Fix crash "cannot form weak reference to instance of viewcontroller"

ArielDemarco:
# Overview
There's a crash happening lately 
> Cannot form weak reference to instance (xxxxxxxx) of class AController. It is possible that this object was over-released, or is in the process of deallocation.
Seems to happen mostly on SwiftUI views (in particular, short-lived `UIHostingController`s and the ones from the navigation/presentation stack).

To prevent this, we got rid of every mention/usage of `UIViewController` in background threads to avoid race conditions.


ferdraghi (approved):

ferdraghi:
In the future it might be worth exploring capturing a weak reference to the VC instead of removing it from the background threads.

ArielDemarco:
> In the future it might be worth exploring capturing a weak reference to the VC instead of removing it from the background threads.
I agree; it's worth trying out as seems simpler to mantain (as long as we don't end up loosing data when those get deleted/`deinit`ed).

NachoEmbrace (approved):

PR #171, opened 2025-02-04
URLSessionDelegateProxy self recursion crash

sergiothinks2:
Puttshack team experiencing a hang/crash noticed in URLSessionDelegateProxy

added checks using !(originalDelegate is URLSessionDelegateProxy) to prevent self-delegation
added similar check for the session delegate: !(sessionDelegate is URLSessionDelegateProxy)
combined the guard conditions to make the code more robust


Sources/EmbraceCore/Capture/Network/Proxy/URLSessionDelegateProxy.swift
```diff
@@ -88,13 +90,10 @@ class URLSessionDelegateProxy: NSObject {
         }
 
         // if session delegate also responds to selector, we must call it
-        if let sessionDelegate = session.delegate,
-           sessionDelegate.responds(to: selector) {
```
ArielDemarco, line 92: I wouldn't remove this line

PR #176, opened 2025-02-06
Core Data Migration

NachoEmbrace:
* Removes GRDB dependency and replaces it with CoreData.
* Added protocols in `EmbraceCommonInternal` to define our data model (ie `EmbraceSession`, `EmbraceSpan`, etc).
  - This was necessary to stop manipulating the `SessionRecord`, `SpanRecord`, etc objects directly
  - These objects are now `NSManagedObjects` and can't be initialized without a CoreData stack
  - Using these protocols instead of the objects allows for easier testing.


sergiothinks2 (commented):
@NachoEmbrace great work here. A lotttt of work. One thought, regarding thread safety, given CoreData requires careful handling of contexts. Do we have something that does explicit handling of background contexts vs main context?

Thinking something like this
`public class CoreDataContextManager {
    let mainContext: NSManagedObjectContext
    let backgroundContext: NSManagedObjectContext
    private let logger: InternalLogger
    
    init(persistentContainer: NSPersistentContainer, logger: InternalLogger) {
        self.logger = logger
        
        // Main context for UI operations
        mainContext = persistentContainer.viewContext
        mainContext.automaticallyMergesChangesFromParent = true
        
        // Background context for async operations
        backgroundContext = NSManagedObjectContext(concurrencyType: .privateQueueConcurrencyType)
        backgroundContext.parent = mainContext
    }
    
    func performBackgroundTask(_ block: @escaping (NSManagedObjectContext) -> Void) {
        backgroundContext.perform { [weak self] in
            guard let self = self else { return }
            block(self.backgroundContext)
            
            do {
                try self.backgroundContext.save()
                self.mainContext.performAndWait {
                    try? self.mainContext.save()
                }
            } catch {
                self.logger.error("Failed to save context: \(error)")
            }
        }
    }
}`

which I know would be a good amount of added work but it could be used in some way like 

`public func cleanUpSpans(date: Date? = nil) {
    contextManager.performBackgroundTask { context in
        let request = SpanRecord.createFetchRequest()
        
        if let date = date {
            request.predicate = NSPredicate(format: "endTime != nil AND endTime < %@", date as NSDate)
        } else {
            request.predicate = NSPredicate(format: "endTime != nil")
        }
        
        let spans = try? context.fetch(request)
        spans?.forEach { context.delete($0) }
    }
}`

ArielDemarco (approved):
💥

PR #177, opened 2025-02-07
Refactor Proxies using NSProxy and Objc

ArielDemarco:


Sources/EmbraceObjCUtilsInternal/include/.swift
ferdraghi, line 1: was this added by accident? it's empty

ferdraghi (approved):

PR #182, opened 2025-02-19
 Fix CoreData Thread Safety in retryCachedData

sergiothinks2:
### Problem

Some customers are experiencing crashes when the upload was not successful and the data was cached. The crash occurs on app relaunch when retryCachedData() is called, specifically when trying to access CoreData managed object properties on a different thread than the managed object context.

### Changes

**Modified retryCachedData() method to handle CoreData objects safely**

- Removed direct property access of CoreData managed objects across threads
- Added conversion of managed objects to value types before processing
- Maintained existing thread safety mechanisms (semaphore, queue handling)
- Verified that test_fetchAllUploadData() is passing 

### Before

The original implementation accessed CoreData managed object properties (uploadData.type, uploadData.data, etc.) directly while iterating through them on a different thread than where they were fetched.

### After
The new implementation:
- Uses existing fetchAllUploadData() method which properly handles CoreData context
- Immediately converts managed objects to value types using map
- Processes these thread-safe value types in the upload loop

### DIFF
--- a/Sources/EmbraceUploadInternal/EmbraceUpload.swift
+++ b/Sources/EmbraceUploadInternal/EmbraceUpload.swift
@@ -81,23 +81,26 @@ public class EmbraceUpload: EmbraceLogUploader {
                 // clear data from cache that shouldn't be retried as it's stale
                 self.clearCacheFromStaleData()
 
-                // get all the data cached first, is the only thing that could throw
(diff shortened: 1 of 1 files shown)


NachoEmbrace:
Were you able to reproduce the crash?

PR #184, opened 2025-02-20
Allows to add a custom stacktrace through the log API

ArielDemarco:
# Overview
This PR adds a new stack trace case to the `StackTraceBehavior` enum: `custom(_ value: EmbraceStackTrace)`.

This allows SDK clients to add their own stack traces to logs, which can be useful, for example, when dealing with past events.

`EmbraceStackTrace` includes all necessary validations to ensure the stack trace (and its frames) adhere to the required format. By design, it `throws` an error (`EmbraceStackTraceError`) when attempting to create an invalid stack trace that cannot be used later by Embrace.

**Limits & Format:**
- Each frame should follow the format: `<index> <image> <memory address> <symbol> [ + <offset> ]`. The `"+ <offset>"` part is optional.
- Each frame cannot be longer than 10.000 characters.
- We'll trim stacktraces with more than 200 frames.


Sources/EmbraceCommonInternal/Models/EmbraceStackTrace.swift
```diff
@@ -0,0 +1,55 @@
...
+    ///
+    /// Example format of _a single frame_:
+    /// ```
+    /// 2   EmbraceApp    0x0000000001234abc  -[MyClass myMethod] + 48
+    /// ```
+    ///
+    /// - Parameter frames: An array of frames strings, following the format of `Thread.callStackSymbols`.
+    /// - Throws: An `EmbraceStackTraceError.invalidFormat` if any of the frames are not in the expected format.
+    public init(frames: [String]) throws {
+        guard frames.allSatisfy(EmbraceStackTrace.isValidStackFrame) else {
```
fnewberg, line 28: * should we have a limit on the length of any individual frame?
* should we have a limit on the max number of frames?
* if we have limits, should we then truncate rather than reject the stacktrace?

ArielDemarco: I don’t know if we should limit individual frames. In particular, with SwiftUI, you can now see really long unmangled symbols.
That said, I’d definitely apply limits to the number of frames we allow to capture, and I’d agree on truncating them by removing the ones with the highest frame numbers

fnewberg: > you can now see really long unmangled symbols.

i meant some limit that we never think we'd hit, like 10k characters, just to ensure that there is some limit. 

>  and I’d agree on truncating them by removing the ones with the highest frame numbers

if we truncate, probably makes sense to put the original frame count as an attribute

ArielDemarco: I like both ideas; i'll update the PR with them 👍

Sources/EmbraceCore/Public/Embrace+OTel.swift
```diff
@@ -119,6 +119,8 @@ extension Embrace: EmbraceOpenTelemetry {
     ///   - severity: `LogSeverity` for the log.
     ///   - attributes: Attributes for the log.
     ///   - stackTraceBehavior: Defines if the stack trace information should be added to the log
+    ///
+    /// - Important: `.info` logs will _never_ have stacktraces.
```
NachoEmbrace, line 123: These comments in the documentation are not clear enough.
It should say that only warnings and errors can have stacktraces right?

NachoEmbrace (approved):

Sources/EmbraceCore/Public/Embrace+OTel.swift
ArielDemarco: Exactly

PR #194, opened 2025-03-20
Unit test updates - non-iOS lifecycle, json encoding, and OS availability

blainemercado:
The following updates are intended to improve the testing performance for all OS types. Some tests were failing, and for some OS versions the project was unable to build.

- Updated endSession test that failed for non-iOS lifecycle devices (watchOS, tvOS) due to differing behavior of endSession method on iOS and non-iOS devices.
-  One JSON encoding test failed due to a mismatched coding key so updated to reference the coding key variable directly instead of hard coding the strings.
-  Updated several tests' OS availabilities that were using swift functions that are only available in later versions of tvOS so project can build.

## Checklist

- [X] PR Description to summarize changes
- [X] Add sufficient test coverage
- [X] Read [Contributing Guidelines](./CONTRIBUTING.md)
- [X] Agree to [Embrace CLA](https://docs.google.com/forms/d/e/1FAIpQLSct_OV9j5-yGCXkhMKOUKpwTxFWdwCQYTPeESk39lOM6k3uKA/viewform)


Tests/EmbraceCoreTests/Payload/ResourcePayloadTests.swift
```diff
@@ -49,31 +49,31 @@ class ResourcePayloadTests: XCTestCase {
...
+        XCTAssertEqual(json[ResourcePayload.CodingKeys.diskTotalCapacity.rawValue] as? Int, 494384795648)
+        XCTAssertEqual(json[ResourcePayload.CodingKeys.deviceArchitecture.rawValue] as? String, "arm64")
+        XCTAssertEqual(json[ResourcePayload.CodingKeys.deviceModel.rawValue] as? String, "arm64_model")
+        XCTAssertEqual(json[ResourcePayload.CodingKeys.deviceManufacturer.rawValue] as? String, "Apple")
+        XCTAssertEqual(json[ResourcePayload.CodingKeys.screenResolution.rawValue] as? String, "1179x2556")
+        XCTAssertEqual(json[ResourcePayload.CodingKeys.osVersion.rawValue] as? String, "17.0.1")
+        XCTAssertEqual(json[ResourcePayload.CodingKeys.osBuild.rawValue] as? String, "23D60")
+        XCTAssertEqual(json[ResourcePayload.CodingKeys.osType.rawValue] as? String, "darwin")
+        XCTAssertEqual(json[ResourcePayload.CodingKeys.osAlternateType.rawValue] as? String, "iOS_variant")
+        XCTAssertEqual(json[ResourcePayload.CodingKeys.osName.rawValue] as? String, "iPadOS")
```
ArielDemarco, line 76: The idea of using hardcoded strings in test instead of `ResourcePayload` is precisely to make it fail when there’s a change. The benefit is that it helps prevent unintentional typos or mistakes when adding or modifying code in `ResourcePayload` (it acts as a kind of double check).
So, if this test fails, I would simply update the key in question (which probably just wasn’t updated, or happened due to some bad merge conflict 😅 ).

ArielDemarco (changes requested):
Thanks a lot for your contribution!! Minor change needed. Aside from that, the PR looks great! If you update the `ResourcePayloadTests`, we can go ahead and merge 👍

Tests/EmbraceCoreTests/Payload/ResourcePayloadTests.swift
blainemercado: Thanks, that makes sense. I updated the PR with this change.

blainemercado:
Updated with the suggested change from the review by @ArielDemarco. Thanks

Sources/EmbraceCore/Payload/ResourcePayload.swift
```diff
@@ -56,7 +56,7 @@ struct ResourcePayload: Codable {
         case deviceManufacturer = "device_manufacturer"
         case screenResolution = "screen_resolution"
         case buildId = "build_id"
-        case bundleVersion = "build"
+        case bundleVersion = "bundle_version"
         case environmentDetail = "environment_detail"
```
ArielDemarco, line 60: As far as I can tell, this change shouldn’t be necessary:
- [On main](https://github.com/embrace-io/embrace-apple-sdk/blob/92aab5fd0d2b639e73cdd2379f012f079eea3c60/Sources/EmbraceCore/Payload/ResourcePayload.swift#L59) this is `build`.
- Tests shouldn't fail as [they show exactly the same value on main](https://github.com/embrace-io/embrace-apple-sdk/blob/92aab5fd0d2b639e73cdd2379f012f079eea3c60/Tests/EmbraceCoreTests/Payload/ResourcePayloadTests.swift#L55)

Have you synced your fork with the latest version of `main`?

blainemercado: You're right, mine was out of sync. I thought I pulled the latest, but it looks like I didn't sync correctly so I didn't have the latest. I fixed that now.

ArielDemarco (approved):

ArielDemarco:
Everything LGTM! Thanks @blainemercado for the contribution!

PR #197, opened 2025-03-31
Fix header injection crash

ArielDemarco:
# Overview
This PR fixes an issue where, in some subtasks, our logic for injecting the `traceparent` header was causing a crash.
This happens because creating Objective-C categories on `NSURLSessionTask` is not recommended, so this PR changes the approach to achieve the same result.


NachoEmbrace (approved):

sergiothinks2 (approved):
Fixes injection header crash, code looks good 👍

PR #198, opened 2025-03-31
Added EMBRACE_ENABLE_TUIST_OBJC_LINK to fix linking issues with tuist

ArielDemarco:
# Overview
When `EmbraceIO` is linked statically, Objective-C selectors from `@objc extensions` are sometimes stripped unless the `-ObjC` linker flag is explicitly passed. This flag is automatically included when using SPM or CocoaPods.

However, when using [Tuist](https://tuist.dev/), this linker setting is not added by default and should be set manually ([here're some suggestions](https://pepicrft.me/blog/2024/06/04/why-you-need-objc) from one of Tuist's creators). This affects targets like `EmbraceCore` and `EmbraceIO` which rely on such extensions to work properly (w/o the flag, some methods will simply crash).

For example: `MetadataHandler` is an `NSObject` that has one `@objc extension` that exposes a `public var userName` with a `getter` and `setter`. If the `ObjC` flag is not present in both the SDK and the app linking it statically, a crash like the following may occur:
```objc
-[EMBMetadataHandler setUserIdentifier:]: unrecognized selector sent to instance
```

## Description
This PR addresses the issue conditionally adding the `-ObjC` linker flag when the environment variable `EMBRACE_ENABLE_TUIST_OBJC_LINK` is set.

## How to enable
If you're using [Tuist](https://tuist.dev/), set the environment variable before building:
```
export EMBRACE_ENABLE_TUIST_OBJC_LINK=1
```

For example:
```
export EMBRACE_ENABLE_TUIST_OBJC_LINK=1 ; tuist install ; tuist generate
```

> [!TIP]
> `Tuist` does a great job caching things. If for some reason you’re not seeing the `-ObjC` flag applied under `Other Linker Flags`, try running `tuist clean`, then reinstall dependencies and regenerate the project.


sergiothinks2 (approved):
tested in 2 separate codebases. LGTM 👍

ferdraghi (approved):

naftaly (approved):

PR #203, opened 2025-04-03
Startup instrumentation

NachoEmbrace:


Sources/EmbraceCore/PrivacyInfo.xcprivacy
```diff
@@ -22,10 +23,10 @@
 		</dict>
 		<dict>
 			<key>NSPrivacyAccessedAPIType</key>
-			<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
```
ArielDemarco, line 25: I think this shouldn't be deleted because now we're going to use it

ArielDemarco (approved):

PR #211, opened 2025-04-10
Update SDK to use dependency source instead of pre-built xcframeworks

naftaly:
This PR points the SPM package to the main KSCrash repo and updates anything required based on differences between our fork and the main repro.

Due to how nicely modularized the SDK is, the pod needed to either be split into a pod per module, or simply not import each module when built from the spec. I decided on the latter due to its simplicity. This allows us to add a compiler directive only when built using the pod and define out the import, which is also the reason for so many file changes.

You'll see a lot of the following. It's the main reason for so many file changes.
```
import EmbraceCoolModule
#endif
```

TODO:
- [x] ensure all changes from the fork are either covered by the main repo or ported.
- [x] update pod and spa to use dependencies directly instead of pre-building them.
- [ ] ~~figure out issues with KSCrash license~~
- [ ] ~~Most of the scripts in `bin` can be updated to not deal with xcframeworks.~~
- [x] All tests pass.

## Checklist


EmbraceIO.podspec
```diff
@@ -10,133 +10,113 @@ Pod::Spec.new do |spec|
   spec.documentation_url                 = "https://embrace.io/docs/"
   spec.license                        = { :type => "Apache 2.0" }
   spec.author                         = "Embrace.io"
-  spec.source                         = { "http" => "https://github.com/embrace-io/embrace-apple-sdk/releases/download/#{spec.version}/embrace_#{spec.version}.zip" }
+  spec.source                         = { :git => "https://github.com/embrace-io/embrace-apple-sdk.git", :tag => spec.version }
```
naftaly, line 13: No need to zip everything up anymore with xcframeworks, it's all built from source. We can point directly to the tag.

EmbraceIO.podspec
```diff
@@ -10,133 +10,113 @@ Pod::Spec.new do |spec|
...
-    core.dependency "EmbraceIO/EmbraceSemantics"
-    core.dependency "EmbraceIO/EmbraceConfiguration"
-  end
-
-  spec.subspec 'EmbraceCommonInternal' do |common|
-    common.vendored_frameworks = "xcframeworks/EmbraceCommonInternal.xcframework"
-  end
+  ## Tell the Swift source code to not import subspecs as modules.
+  spec.pod_target_xcconfig = {
+    'SWIFT_ACTIVE_COMPILATION_CONDITIONS' => 'EMBRACE_COCOAPOD_BUILDING_SDK'
```
naftaly, line 21: In code, we can do the following:

```
import EmbraceCoolModule
#endif
```

Package.swift
```diff
@@ -40,8 +40,8 @@ let package = Package(
     ],
     dependencies: [
         .package(
-             url: "https://github.com/embrace-io/KSCrash.git",
-             exact: "2.0.7"
+             url: "https://github.com/kstenerud/KSCrash",
```
naftaly, line 43: KSCrash directly from the source.

Tischel:
Were the changes in `CrashReporter` necessary?
Those technically mean a breaking change to the public API which we don't really want to do right now.

NachoEmbrace:
Were the changes in `CrashReporter` necessary?
Those technically mean a breaking change to the public API and we don't really want to do that right now.

naftaly:
> Were the changes in `CrashReporter` necessary? Those technically mean a breaking change to the public API and we don't really want to do that right now.

I made that change due to how highly modular the SDK is. In SPM, the modules stay the same so we have a "namespace", but in the pod, we lose modularity unless we split each subspec up into pod specs. This leaves us with a different "namespace". All that being said, KSCrash has its own `CrashReport` so we end up conflicting with it unless we namespace ours.

Updating the name was the simplest change I could make that would have minimal repercussions. I'm happy to revisit this, or find a better solution.

Sources/EmbraceCommonInternal/CrashReporter/CrashReporter.swift
```diff
@@ -8,29 +8,29 @@ import Foundation
...
 ///
 /// Implementing this protocol is optional and should only be considered in cases where
 /// additional customization in error reporting is required.
-public protocol ExtendableCrashReporter: CrashReporter {
+public protocol EmbraceExtendableCrashReporterPlugin: EmbraceCrashReporterPlugin {
     func appendCrashInfo(key: String, value: String)
 }
 
-@objc public class CrashReport: NSObject {
+@objc public class EmbraceCrashReport: NSObject {
```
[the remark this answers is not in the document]
naftaly: Screenshot 2025-04-15 at 2 28 44 PM

ArielDemarco (approved):

PR #219, opened 2025-04-23
Fixed crash in DefaultURLSessionTaskHandler.addData

### LINKED ISSUE #217: Crash on `URLSessionDelegateProxy+DataDelegate.swift @ 41`
SDK version: `6.5.2`
Platform: iOS
Crash id: `836DD40777985DFDB5BB7C3FB331D381`

I am seeing a crash on `URLSessionDelegateProxy+DataDelegate.swift @ 41`: 

```swift
        if var previousData = dataTask.embraceData {
            previousData.append(data) <------------------- Crash is here. The array's access is not protected.
            dataTask.embraceData = previousData
        } else {
            dataTask.embraceData = data
        }
```

ArielDemarco:
# Overview
A crash was occurring when attempting to append data to a URLSessionTask’s associated `embraceData` because the associated_object was not done in a way that would ensure thread safety.

# Details
- Switched from `OBJC_ASSOCIATION_RETAIN` to `OBJC_ASSOCIATION_COPY` when associating `embraceData` to `URLSessionTask` to ensure data consistency and avoid thread/memory issues.
- Added a guard around the `addData` method to ensure it only performs operations if network body capture feature is enabled (which, by default, is off).

Fixes #217

### DIFF
--- a/Sources/EmbraceCore/Capture/Network/URLSessionTask+Extension.swift
+++ b/Sources/EmbraceCore/Capture/Network/URLSessionTask+Extension.swift
@@ -30,6 +30,12 @@ extension URLSessionTask {
         }
     }
 
+    /// Stores the accumulated response body data for a `URLSessionTask`.
+    ///
+    /// This is primarily used when network body capture is enabled.
+    ///
+    /// - Warning: Access to this property must be synchronized externally (e.g., using `DispatchQueue` or `UnfairLock`)
+    /// to ensure thread safety, since `URLSessionTask` is shared across threads.
     var embraceData: Data? {
         get {
             return objc_getAssociatedObject(self,
--- a/Sources/EmbraceCore/Capture/Network/URLSessionTask+Extension.swift
+++ b/Sources/EmbraceCore/Capture/Network/URLSessionTask+Extension.swift
@@ -39,7 +39,7 @@ extension URLSessionTask {
             objc_setAssociatedObject(self,
                                      &AssociatedKeys.embraceData,
                                      newValue,
-                                     .OBJC_ASSOCIATION_RETAIN)
+                                     .OBJC_ASSOCIATION_COPY)
         }
     }
 
--- a/Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
+++ b/Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
@@ -26,12 +26,15 @@ protocol URLSessionTaskHandlerDataSource: AnyObject {
 final class DefaultURLSessionTaskHandler: NSObject, URLSessionTaskHandler {
     private var spans: [URLSessionTask: Span] = [:]
     private let queue: DispatchableQueue
+    private let capturedDataQueue: DispatchableQueue
     private let payloadCaptureHandler: NetworkPayloadCaptureHandler
     weak var dataSource: URLSessionTaskHandlerDataSource?
 
     init(processingQueue: DispatchableQueue = DefaultURLSessionTaskHandler.queue(),
+         capturedDataQueue: DispatchableQueue = DefaultURLSessionTaskHandler.capturedDataQueue(),
          dataSource: URLSessionTaskHandlerDataSource?) {
         self.queue = processingQueue
+        self.capturedDataQueue = capturedDataQueue
         self.dataSource = dataSource
         self.payloadCaptureHandler = NetworkPayloadCaptureHandler(otel: dataSource?.otel)
     }
@@ -159,11 +162,13 @@ final class DefaultURLSessionTaskHandler: NSObject, URLSessionTaskHandler {
             }
 
             if let data = data {
-                let totalData = task.embraceData ?? data
-                span.setAttribute(
-                    key: SpanSemantics.NetworkRequest.keyResponseSize,
-                    value: totalData.count
-                )
+                self.capturedDataQueue.sync {
+                    let totalData = task.embraceData ?? data
+                    span.setAttribute(
+                        key: SpanSemantics.NetworkRequest.keyResponseSize,
+                        value: totalData.count
+                    )
+                }
             }
 
             if let error = error ?? task.error {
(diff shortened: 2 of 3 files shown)


ferdraghi (approved):

Sources/EmbraceCore/Capture/Network/URLSessionTask+Extension.swift
```diff
@@ -39,7 +39,7 @@ extension URLSessionTask {
             objc_setAssociatedObject(self,
                                      &AssociatedKeys.embraceData,
                                      newValue,
-                                     .OBJC_ASSOCIATION_RETAIN)
+                                     .OBJC_ASSOCIATION_COPY)
```
naftaly, line 42: Nice. There's a small catch however. `OBJC_ASSOCIATION_COPY` only copies when set. When the copied object is retrieved, no copy is made so you can end up with a shared object just like it was beforehand.

Due to that, I think a copy on `var embraceData: Data?` might be wise.

Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
```diff
@@ -191,11 +191,13 @@ final class DefaultURLSessionTaskHandler: NSObject, URLSessionTaskHandler {
...
 
     func addData(_ data: Data, dataTask: URLSessionDataTask) {
-        if var previousData = dataTask.embraceData {
-            previousData.append(data)
-            dataTask.embraceData = previousData
-        } else {
-            dataTask.embraceData = data
+        if payloadCaptureHandler.isEnabled() {
+            if var previousData = dataTask.embraceData {
+                previousData.append(data)
```
naftaly, line 196: As is, without a copy on return as mentioned above, you might still end up with a shared reference that gets mutated and could cause threading issues.

Sources/EmbraceCore/Capture/Network/URLSessionTask+Extension.swift
naftaly: However, Data is struct and uses COW, so it's hard to tell if you need to copy. Unless the data comes in from ObjC, then you might need to ensure the copies. Any thoughts?

ArielDemarco: > However, Data is struct and uses COW, so it's hard to tell if you need to copy. Unless the data comes in from ObjC, then you might need to ensure the copies. Any thoughts?

I thought about this; if this was the case, the only solution would be wrapping in a queue, cause the underlying data type could be either set as mutable (`NSMutableData`) or inmutable (`NSData`). But only we're calling this in Swift so I don't think the `COPY` would be a problem. That said, we could still wrap this solution around a queue if we have doubts

naftaly: If we're going to end up locking the resource, I'd like to understand where the usage is coming from that is causing the problem in the first place. Is this data shared around a lot and vended outside, or do we fully control it?

ArielDemarco: We fully control when the data is get/set. For reference, the only place where we `set` it is inside `EMBURLSessionDelegateProxy`, in the `didReceiveData` method; there, we call `DefaultURLSessionTaskHandler.addData`.

The `getter` is also called within the same method (before setting, we check whether to append or not). Additionally, whenever a captured `URLSessionTask` reaches `DefaultURLSessionTaskHandler.finish(task:data:error:)`, the getter is triggered as well.

naftaly: Ok, if this data is not public and we control 100% of its lifetime, then your original idea of the lock is better. Sorry about the churn.

Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
RuiAAPeres: I am unsure how this would prevent a data race: the `dataTask.embraceData` access is still unprotected.

Sources/EmbraceCore/Capture/Network/URLSessionTask+Extension.swift
RuiAAPeres: I don't think `objc_setAssociatedObject` is thread safe. And it can still crash when accessed from multiple threads. A quick google search shows many examples of that.

ArielDemarco: Yes! We know! PR is not updated with latest changes yet; associated_object will also be wrapped in a sync queue access 🙂

Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
ArielDemarco: It doesn't! PR description tries to clarify this piece of code:

> Added a guard around the `addData` method to ensure it only performs operations if network body capture feature is enabled (which, by default, is off).

This means that in most cases, it won’t crash because the piece of code won’t even be reachable.
Most of the time (and for most customers) the Network Payload Capture feature is turned off (and if it’s enabled, it’s usually just for a specific period of time to gather a specific amount of data to resolve a particular situation).

So, to be clear: this was simply something we noticed while reviewing the code. It prevents storing `Data` in memory unnecessarily and helps avoid wasting resources when it’s not needed; and, at the same time, to prevent from doing extra logic that is unnecessary.

ArielDemarco:
> Looks good. I have two notes.
> 
> 1- Do we ever use embraceData anywhere? Seems like we're only counting bytes. If so, maybe we just keep tabs on the count instead. 2- A queue is a bit heavy handed for this. We could instead us an unfair lock, or even an NSLock since its backed by a pthread_mutex. 3- Let's put in a comment to explain what embraceData is used for and how it's protected by the queue.

1. I'd say that 99% of the times the `embraceData` is used for the size. There's still one use case (Network Payload Capture) were we need the actual payload. That's why 1) I've added the `payloadCaptureHandler.isEnabled()` check prior to access the `embraceData` variable and 2) I've raised [this other PR](https://github.com/embrace-io/embrace-apple-sdk/pull/220), that'll not only ensure we always have the payload when we need it, but also don't add it whenever is not necessary.
2. We could try to use the `UnfairLock` (from `EmbraceCommonInternal`). I can do that on the other PR if needed.
3. I can do that pre-merge, no prob

NachoEmbrace (approved):

PR #220, opened 2025-04-23
Ensure payload data is always included when Network Payload Capture is enabled

ArielDemarco:
# Overview
This PR introduces the necessary changes to ensure that, whenever the **Network Payload Capture** functionality is enabled, the payload associated with a request is properly captured regardless of whether `async` APIs, block-based (`completionHandler`), or delegation patterns are used.

# Details
- Removed `embraceEndTime` as it's an unnecessary associated object to the `URLSessionTask`. 
- Make some changes to make the `endTime` that the **Network Payload Capture** uses to be more accurate.

### DIFF
--- a/Sources/EmbraceCore/Capture/Network/URLSessionTask+Extension.swift
+++ b/Sources/EmbraceCore/Capture/Network/URLSessionTask+Extension.swift
@@ -9,7 +9,6 @@ extension URLSessionTask {
         static var embraceCaptured: UInt8 = 0
         static var embraceData: UInt8 = 1
         static var embraceStartTime: UInt8 = 2
-        static var embraceEndTime: UInt8 = 3
     }
 
     var embraceCaptured: Bool {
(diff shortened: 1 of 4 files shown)


Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
```diff
@@ -125,69 +125,96 @@ final class DefaultURLSessionTaskHandler: NSObject, URLSessionTaskHandler {
...
-                data: data,
-                error: error,
-                startTime: task.embraceStartTime,
-                endTime: task.embraceEndTime
-            )
-
-            // stop if the service is disabled
-            guard self.dataSource?.state == .active else {
-                return
+            if self.payloadCaptureHandler.isEnabled() {
```
naftaly, line 134: Unless you need the serialization of `queue`, maybe run the check before the jump to the queue to save more cycles.

ArielDemarco: Aside of this, i have to run inside the `queue` the `self.handleTaskFinished`, so I don't think I can avoid it in this case

Sources/EmbraceCore/Capture/Network/URLSessionTask+Extension.swift
```diff
@@ -61,17 +60,4 @@ extension URLSessionTask {
...
-            return objc_getAssociatedObject(self,
-                                            &AssociatedKeys.embraceEndTime) as? Date
-        }
-        set {
-            objc_setAssociatedObject(self,
-                                     &AssociatedKeys.embraceEndTime,
-                                     newValue,
-                                     .OBJC_ASSOCIATION_RETAIN)
-        }
-    }
```
ArielDemarco, line 76: Removed as it's no longer being used

NachoEmbrace (approved):

naftaly (approved):
There seem to be a lot of whitespace changes or formatting unrelating to the PR, maybe these should be moved to another cleanup PR to keep things clear.

ArielDemarco:
> There seem to be a lot of whitespace changes or formatting unrelating to the PR, maybe these should be moved to another cleanup PR to keep things clear.

TBH, I don't know exactly _why_ it happened, as those files shouldn't have been modified by the linter in the way they did

PR #224, opened 2025-04-29
Improvements around data cleanup and fetches

NachoEmbrace:
This PR addresses 2 things:
* Changes the order of operations when sending (and deleting) sessions, so the spans and metadata are cleaned up before hand. This change was made to fix a potential race condition that we detected.
* Improves code around fetches with Core Data. The previous implementation was pretty lazy and caused a lot of jumps between threads that were not needed.

### DIFF
--- a/Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
+++ b/Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
@@ -110,6 +110,17 @@ public class CoreDataWrapper {
         return result
     }
 
+    public func fetchAndPerform<T>(
+        withRequest request: NSFetchRequest<T>,
+        block: (([T]) -> Void)) where T: NSManagedObject {
+        context.performAndWait {
+            do {
+                let result = try context.fetch(request)
+                block(result)
+            } catch { }
+        }
+    }
+
     /// Synchronously fetches the count of records that satisfy the given request
     public func count<T>(withRequest request: NSFetchRequest<T>) -> Int where T: NSManagedObject {
         var result: Int = 0
@@ -134,6 +145,24 @@ public class CoreDataWrapper {
             }
 
             try? self?.context.save()
+        }
+    }
+
+    public func deleteRecords<T>(withRequest request: NSFetchRequest<T>)where T: NSManagedObject {
+        context.performAndWait { [weak self] in
+            guard let strongSelf = self else {
+                return
+            }
+
+            guard let records = try? strongSelf.context.fetch(request) else {
+                return
+            }
+
+            for record in records {
+                strongSelf.context.delete(record)
+            }
+
+            try? strongSelf.context.save()
         }
     }
 }
--- a/Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
+++ b/Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
@@ -94,18 +94,25 @@ public class CoreDataWrapper {
                 try self?.context.save()
             } catch {
                 let name = self?.context.name ?? "???"
-                self?.logger.warning("Error saving CoreData \"\(name)\": \(error.localizedDescription)")
+                self?.logger.error("Error saving CoreData \"\(name)\": \(error.localizedDescription)")
             }
         }
     }
 
     /// Synchronously fetches the records that satisfy the given request
     public func fetch<T>(withRequest request: NSFetchRequest<T>) -> [T] where T: NSManagedObject {
+
         var result: [T] = []
-        context.performAndWait {
+        context.performAndWait { [weak self] in
+            guard let self else {
+                return
+            }
+
             do {
                 result = try context.fetch(request)
-            } catch { }
+            } catch { 
+                self.logger.error("Error fetching!!!:\n\(error.localizedDescription)")
+            }
         }
         return result
     }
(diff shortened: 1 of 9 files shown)


Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -110,6 +110,17 @@ public class CoreDataWrapper {
...
     }
 
+    public func fetchAndPerform<T>(
+        withRequest request: NSFetchRequest<T>,
+        block: (([T]) -> Void)) where T: NSManagedObject {
+        context.performAndWait {
+            do {
+                let result = try context.fetch(request)
+                block(result)
+            } catch { }
```
naftaly, line 120: Let's at least log something when an error is thrown. What do we usually do throughout the SDK with errors like this?

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -136,4 +147,22 @@ public class CoreDataWrapper {
             try? self?.context.save()
         }
     }
+
+    public func deleteRecords<T>(withRequest request: NSFetchRequest<T>)where T: NSManagedObject {
+        context.performAndWait { [weak self] in
+            guard let strongSelf = self else {
```
naftaly, line 153: You can simply do the following if you want, I find it cleaner.

```
guard let self else { return }
```

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -136,4 +147,22 @@ public class CoreDataWrapper {
...
+                return
+            }
+
+            guard let records = try? strongSelf.context.fetch(request) else {
```
naftaly, line 157: we shouldn't disregard anything that might get thrown here. If we have a good reason to, let's leave a comment about why for future .

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -136,4 +147,22 @@ public class CoreDataWrapper {
...
+                return
+            }
+
+            for record in records {
+                strongSelf.context.delete(record)
+            }
+
+            try? strongSelf.context.save()
```
naftaly, line 165: same.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -136,4 +147,22 @@ public class CoreDataWrapper {
...
+            guard let strongSelf = self else {
+                return
+            }
+
+            guard let records = try? strongSelf.context.fetch(request) else {
+                return
+            }
+
+            for record in records {
+                strongSelf.context.delete(record)
```
naftaly, line 162: Depending on our goals, it might be worth looking at batching our deletes. It can be much more performant with the caveats that you need to merge back into the context afterwards.

naftaly (commented):
There a lot of `fetchAndPerform` in order to simply return the first object, there might be something to optimize there. Maybe a a `fetchFirst(_:)` would work. Maybe the fact we only want the first can be encoded into the predicate/request to help CoreData optimize.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
NachoEmbrace: I think in general we're not doing a great job at dealing with internal errors like this.
I'll add some internal logs for now, but we should probably discuss if we need to do more.

NachoEmbrace:
> There a lot of `fetchAndPerform` in order to simply return the first object, there might be something to optimize there. Maybe a a `fetchFirst(_:)` would work. Maybe the fact we only want the first can be encoded into the predicate/request to help CoreData optimize.

I do set the `fetchLimit` to 1 for requests were only 1 result matter (though I might've missed some requests, to be fair).
I'll add a `fetchFirstAndPerform` as a QoL function still.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
NachoEmbrace: When I was building the feature I tried to do something like that, but ran into problems.
I wouldn't be opposed to trying again, but not on this PR.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
NachoEmbrace: Same as above, adding an internal log for now.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -144,25 +183,30 @@ public class CoreDataWrapper {
                 self?.context.delete(record)
```
naftaly, line 183: Reuse of `self` here can be dangerous since it's weak and could go nil at any point which would lead to data issues and we wouldn't know. Do the `guard let self else { return }`. 

This also makes me wonder if these functions should be throwing so we can handle data corruption and errors at a higher level. The new `catch` and logging is good, but we'll still lose data and not really have insight into it. This can probably wait for another PR however. Maybe we need a new feature at some point that adds observability to our SDK ;P

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -133,7 +183,30 @@ public class CoreDataWrapper {
...
             }
 
-            try? self?.context.save()
+            do {
+                try self?.context.save()
```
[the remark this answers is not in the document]
NachoEmbrace: Oops, yea.
Will be addressed on https://github.com/embrace-io/embrace-apple-sdk/pull/226

This PR is getting closed since it was made from the wrong branch.

PR #228, opened 2025-05-05
Updated create-release workflow to work with the new schema

ArielDemarco:
# Overview
Updated the `create-release` workflow to align with the changes introduced in [this PR](https://github.com/embrace-io/embrace-apple-sdk/pull/211).

# Description
- Removed the forced creation of `XCFrameworks` (did not delete `build_xcframework.sh` as we may still offer a way for customers to build them if needed). This also removes the installation of `mise` and `tuist`.
- The `.zip` containing all the `XCFrameworks` (for CocoaPods) is no longer required, so that step was removed.
- Creating a GitHub release now simply involves tagging, with or without the `--prerelease` flag.
- _Bonus_: Fixed the job that records the internal SDK Version History.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -132,112 +132,17 @@ jobs:
       - name: Select Xcode 15
         run: sudo xcode-select -switch /Applications/Xcode_15.4.app
 
-      - name: Install Mise
-        run: |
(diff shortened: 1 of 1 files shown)


naftaly (approved):
🤟

PR #230, opened 2025-05-05
[WIP] SwiftUI View Tracing

naftaly:


Sources/EmbraceCore/SwiftUI/EmbraceTracePhase.swift
```diff
@@ -0,0 +1,82 @@
...
+        guard let client = Embrace.client else { return nil }
+        
+        let sanitizedName = "\(name.lowercased())-\(next)"
+        
+        //print("\(logPrefix)[AC:START:\(CACurrentMediaTime())] \(function) - \(sanitizedName)")
+
+        let builder = client.buildSpan(name: sanitizedName, attributes: attributes ?? [:])
+        if let parent = spanStack.last {
+            builder.setParent(parent)
+        }
```
ArielDemarco, line 55: Could those spans be siblings instead of parent-child?

Sources/EmbraceCore/SwiftUI/RunLoopTracker.swift
```diff
@@ -0,0 +1,231 @@
+import Foundation
+import QuartzCore
+import EmbraceCommonInternal
+
+public func preconditionOnRunLoop(_ runLoop: @autoclosure () -> RunLoop) {
+    precondition({
+        runLoop() == RunLoop.current
+    }())
+}
```
ArielDemarco, line 9: Seems is not being used

naftaly: I'll remove it later if I end up not using it.

Sources/EmbraceCore/SwiftUI/EmbraceTracePhase.swift
naftaly: What would the benefit be of them being siblings vs. parent/child? This isn't complete yet, but I'd like to find a way to ensure that there's a hierarchy within the bodies being executed and the source of the change.

Sources/EmbraceCore/SwiftUI/EmbraceTrace+ViewModifier.swift
```diff
@@ -0,0 +1,8 @@
+import SwiftUI
+
+@available(iOS 13, macOS 10.15, tvOS 13, watchOS 6.0, *)
+public extension View {
+    func embraceTrace(_ viewName: String, attributes: [String: String]? = nil) -> some View {
+        EmbraceTraceView(viewName, attributes: attributes) { self }
+    }
```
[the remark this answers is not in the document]
naftaly: With the View, a user can make it their top level return from another view if they wanted.

```
var body: any View {
   EmbraceTraceView("my-view") {
      List {
         Text("1")
         Text("2")
      }
   }
}
```

Sources/EmbraceCore/SwiftUI/EmbraceTracePhase.swift
ArielDemarco: I meant that the relationship might not necessarily be parent-child. It could be siblings (or just a `Link`). It’s not so much about the benefits but rather the accuracy. From what I understand, using `embraceTrace` with examples like:
```swift
ParentView {
    ChildView { ... }.embraceTrace("child")
}.embraceTrace("parent")
```
and
```swift
VStack {
    FirstChildView { ... }.embraceTrace("parent")
    SecondChildView { ... }.embraceTrace("child")
}
```
With the current logic, it seems both cases would report the spans as parent and child, even though in the second example, they aren’t.

naftaly: it's not supposed to, but I'll double check the logic. I do need to split up the logic between bodies and states and how they interact.

naftaly: Cleaned this up and it works correctly now.

Sources/EmbraceCore/SwiftUI/EmbraceTracePhase.swift
```diff
@@ -0,0 +1,234 @@
...
+
+/// Default implementation for `isEmpty` based on `peek()`.
+fileprivate extension LinearCollection {
+    var isEmpty: Bool { peek() == nil }
+}
+
+/// FIFO queue implementation for cycle-based spans.
+fileprivate class Queue: LinearCollection {
+    var storage: [Span] = []
+    func pop() -> Span? {
```
fnewberg, line 211: does this function need synchronization given that we're both reading from and writing to `storage`?

naftaly: It doesn’t. Consider it to be just like a low level data structure, the sync will happen at a higher level depending on usage.

ArielDemarco (approved):

NachoEmbrace (approved):

PR #233, opened 2025-05-09
Hang Capture Service

naftaly:
This PR creates a service to capture hangs. Hangs are sent in the same manner as they are on Android as [described here](https://www.notion.so/embraceio/ANRs-1d77e3c9985281c58765d8c622443e2c). A similar span is also sent using iOS terminology as an interim in order for it to appear in the timeline.

- Currently stack traces are not sent, this is for a future PR.
- The iOS span appears as "emb_thread_blockage" in the Performance timeline.
- Limit is set to 200 hangs per session. Currently hard coded.
- Limit is set to 0 samples per hang. Currently hard coded.
- Add a session attribute named `emb-thread-blockage` with the count of hangs in the session.
- ~~Currently using `ux` for the type, I think it needs to be `perf.thread_blockage`.~~

**Client Work**
- [x] Add limits.
- [x] Figure out what spans to log (ANR or Hang), and get backend work done accordingly.
- [ ] If there's an un-terminated hang span on startup, we should determine that a crash occurred due to a watchdog event (SIGKILL 0x8badf00d), but still close the span and have it appear in the timeline.
- [x] Understand the overflow error that is currently commented out.

**Backend work**
- [x] Make sure the correct naming is used on each platform. (iOS => Hang)
- [x] Show `perf.thread_blockage` spans in timeline.
- [x] Add remote config limits, here's the structure used in the app:
```
hang_limits: {
    hang_per_session: Uint,
    samples_per_hang: Uint
}
```

# Gate
Set hang_per_session = 0 to ensure the watchdog doesn't run.

# Visuals
Screenshot 2025-09-03 at 1 36 48 PM

### DIFF
--- a/Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
+++ b/Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
@@ -1,0 +1,175 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+import OpenTelemetryApi
+
+#if !EMBRACE_COCOAPOD_BUILDING_SDK
+    import EmbraceCaptureService
+    import EmbraceCommonInternal
+    import EmbraceOTelInternal
+    import EmbraceSemantics
+    import EmbraceConfiguration
+#endif
+
+/// Service that generates OpenTelemetry span events for hangs.
+@objc(EMBHangCaptureService)
+public final class HangCaptureService: CaptureService {
+
+    public init(
+        limits: HangLimits = HangLimits()
+    ) {
+        dispatchPrecondition(condition: .onQueue(.main))
+        self.mainThread = pthread_self()
+        self.limitData = EmbraceMutex(LimitData(limits: limits))
+        super.init()
+    }
+
+    public override func onInstall() {
+        // Since we use `limits.hangPerSession` as a gate for the watchdog,
+        // we need to wait until the remote config is actually loaded from disk
+        // which happens just before this call.
+        watchdog = limits.hangPerSession > 0 ? HangWatchdog() : nil
+        watchdog?.hangObserver = self
+        watchdog?.logger = logger
+    }
+
+    public override func onSessionStart(_ session: any EmbraceSession) {
+        limitData.withLock { $0.hangsInSessionCount = 0 }
+    }
+
+    public override func onSessionWillEnd(_ session: any EmbraceSession) {
+        let value = limitData.withLock { $0.hangsInSessionCount }
+        try? Embrace.client?.metadata.updateProperty(key: "emb-thread-blockage", value: "\(value)")
+    }
+
+    public override func onConfigUpdated(_ config: any EmbraceConfigurable) {
+        self.limitData.withLock { $0.limits = config.hangLimits }
+    }
+
+    private var mainThread: pthread_t
+    private var watchdog: HangWatchdog?
+    private let queue = DispatchQueue(label: "io.embrace.hang.service")
+
+    private var span: OpenTelemetryApi.Span?
+
+    struct LimitData {
+        var limits: HangLimits = HangLimits()
+        var hangsInSessionCount: UInt = 0
+        var samplesInHangCount: UInt = 0
+    }
+    var limitData: EmbraceMutex<LimitData>
+
+    public var limits: HangLimits {
+        get {
+            limitData.withLock { $0.limits }
+        }
+        set {
+            limitData.withLock { $0.limits = newValue }
+        }
+    }
+}
+
+extension HangCaptureService: HangObserver {
+
+    // Hang span documented here:
+    // https://www.notion.so/embraceio/ANRs-1d77e3c9985281c58765d8c622443e2c
+
+    public func hangStarted(at: NanosecondClock, duration: NanosecondClock) {
+
+        logger?.debug("[Watchdog] Hang started, at \(at.date) after waiting \(duration.uptime.milliseconds) ms")
+
+        // Keep tabs on how many hang spans we've created
+        guard
+            limitData.withLock({
+                $0.samplesInHangCount = 0
+                $0.hangsInSessionCount += 1
+                return $0.hangsInSessionCount <= $0.limits.hangPerSession
+            })
+        else {
+            let limitData = self.limitData.withLock { $0 }
+            logger?.warning(
+                "[Watchdog] Dropping hang due to surpassing limit, \(limitData.hangsInSessionCount) of \(limitData.limits.hangPerSession)")
+            return
+        }
+
+        // build the span
+        guard
+            let builder = buildSpan(
+                name: "emb-thread-blockage",
+                type: SpanType(primary: .performance, secondary: "thread_blockage"),
+                attributes: [:]
+            )
+        else {
+            logger?.warning("[Watchdog] failed to create emb-thread-blockage span.")
+            return
+        }
+
+        queue.async { [self] in
+            span =
+                builder
+                .setStartTime(time: at.date)
+                .setAttribute(key: "last_known_time_unix_nano", value: .int(Int(at.realtime)))
+                .setAttribute(key: "interval_code", value: .int(0))
+                .setAttribute(key: "thread_priority", value: .int(0))
+                .startSpan()
+        }
+    }
+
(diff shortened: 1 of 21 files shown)


Package.swift
```diff
@@ -132,7 +136,8 @@ let package = Package(
             name: "EmbraceCaptureService",
             dependencies: [
                 "EmbraceOTelInternal",
-                .product(name: "OpenTelemetrySdk", package: "opentelemetry-swift")
+                .product(name: "OpenTelemetrySdk", package: "opentelemetry-swift"),
+                .product(name: "Atomics", package: "swift-atomics")
```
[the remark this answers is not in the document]
naftaly: Ok. I can get rid of Atomics and simply use a lock for now, or use built-in C atomics, this was just easier for the draft.

Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    // Hang span documented here:
+    // https://www.notion.so/embraceio/ANRs-1d77e3c9985281c58765d8c622443e2c
+
+    public func hangStarted(at time: UInt64, duration: UInt64) {
+
+        logger?.debug("[Watchdog] Hang started, at \(nanosecondsToMilliseconds(duration)) ms")
+        let startTime = Date(timeIntervalSinceNow: -nanosecondsToSeconds(duration))
+
+        // ANR span (ewww!)
```
ArielDemarco: 🤣

Sources/EmbraceCore/Capture/Hang/Watchdog/HangWatchdog.swift
```diff
@@ -0,0 +1,315 @@
...
+            precondition(Thread.current == self.watchdogThread)
+
+            let now = suspendingTimeInNanoseconds()
+            let (observer, startHang, enterTime, hangTime) = hangData.withLock {
+
+                let enterTime = $0.enterTime
+                let (hangTime, overflow) = now.subtractingReportingOverflow(enterTime)
+                if overflow {
+                    self.logger?.error("[Watchdog] overflow \(now) - \(enterTime)")
+                    // return (nil, false, enterTime, hangTime)
```
ArielDemarco: Should this be commented?

naftaly: For now yes, I'm still unsure why it happens. The hang is still valid without the overflow. I'll add it to the things I still need to finish up in the summary. Thanks for pointing it out.

Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
naftaly: Every chance I have. :)

Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
```diff
@@ -0,0 +1,130 @@
...
+            .setStartTime(time: startTime)
+            .setAttribute(key: "last_known_time_unix_nano", value: .int(Int(time)))  // this is not unix time, it's uptime raw
+            .setAttribute(key: "interval_code", value: .int(0))
+
+        anrSpan = builder.startSpan()
+
+        // Hang span :)
+        guard
+            let builder = buildSpan(
+                name: "hang",
```
ArielDemarco: Can we make this `emb-hang` or something like that? What do you think? Or do you prefer to avoid the `emb` prefix for some reason?

naftaly: I just avoid it because it shows up in the timeline. But we should have it in there. I'll add it.

Sources/EmbraceCore/Capture/Hang/Watchdog/HangWatchdog.swift
```diff
@@ -0,0 +1,315 @@
...
+/// Implement these methods to handle the various phases of a detected hang.
+///
+/// - hangStarted: Called when a hang is first detected on a background queue.
+/// - hangUpdated: Called periodically while a hang persists on a background queue.
+/// - hangEnded: Called when the hang ends on the hung thread.
+public protocol HangObserver: AnyObject {
+    func hangStarted(at nanoseconds: UInt64, duration nanoseconds: UInt64)
+    func hangUpdated(at nanoseconds: UInt64, duration nanoseconds: UInt64)
+    func hangEnded(at nanoseconds: UInt64, duration nanoseconds: UInt64)
+}
```
ArielDemarco: I have two doubts regarding these interfaces (_Side note_: this will probably affect `hangStarted` and `hangEnded`, since we’re not really using `hangUpdated` right now; so maybe my doubts are more related to those two methods).

As far as I can see, we’re not using the `at nanoseconds: UInt64` anywhere, and as far as I understand, that's because it's not a time we can map to a `Date` object, but rather a timestamp relative to the process startup.
Assuming I'm right (though I might have missed something), are we okay with removing the `at nanoseconds` parameter and leaving the interfaces with just the `duration`? For example:
```swift
public protocol HangObserver: AnyObject {
    func hangStarted(duration nanoseconds: UInt64)
    func hangUpdated(duration nanoseconds: UInt64)
    func hangEnded(duration nanoseconds: UInt64)
}
```

My other question is about the naming of the `duration` parameter in the `hangStarted` method. It seems like the value actually represents a `delay` since the hang was detected (which is also why we can call it a `duration`). Would it be better to rename it to `delay`? Seems like, if that's the case, is a bit easier to follow and understand. 
BTW, feel free to correct me if I’m misunderstanding its usage.

Sources/EmbraceCore/Capture/Hang/Watchdog/HangWatchdog.swift
ArielDemarco: Maybe it's something we can pass as an attribute of the span whenever it happens, so we can easily track if there was an overflow.
And by the way, did it ever happen while testing? Maybe I'm worrying too much about something that wouldn't actually happen

Sources/EmbraceCore/Capture/Hang/Watchdog/HangWatchdog.swift
naftaly: For the `at` field, I'm trying to slow get us to move to a monotonic clock. Using Date is simply a recipe for disaster since it can be moved around and is not good for performance work. You'll notice that when the times are used in this code we have a starting point which we move around based on a timing/clock that cannot be altered by any external factors. So I want to keep it, it's important.

As far as `duration` goes, it is actually the current duration of the hang, it isn't the a delay. If we used `delay`, it would mean the amount of time since the thread at hand has been stopped. In this case, we're looking at how long in time is the hang. So basically, `duration` is relative to the hang, and `delay` would be relative the _thing_ that is hanging. It's just semantics really.

naftaly: But if there are points to bring up around this PR, those are some really important ones that are worth talking about. Time is especially complicated when it comes to the different versions that exist.

naftaly: For the `at` field, we might add another parameter as a `Date` that we think is valid though. I think that's worthy.

ArielDemarco: Great, I think the `duration` name is the right one. Thanks for the clarification!

> For the `at` field, we might add another parameter as a `Date` that we think is valid though. I think that's worthy.

I think that would be good in terms of usage; but I trust your judgement in terms of understanding if its the right call or not to do so

naftaly: I've updated the timestamps to use a new clock that holds the 3 important type of timestamps required to do this right.

- uptime: always incrementing but stops during suspension.
- monotonic: always incrementing, never stops.
- realtime: unix epoch

With these 3 anything can be calculated correctly without losing precision.

Sources/EmbraceCore/Capture/Hang/Watchdog/HangWatchdog.swift
naftaly: cleaned it up and fixed it with the new clock.

PR #235, opened 2025-05-14
Add EmbraceMutex and use in a few areas

naftaly:
This pull request introduces EmbraceMutex<Value>, a backward-compatible alternative to iOS 18's Mutex<Value>, for thread-safe access to mutable data. The changes also replace existing thread-safety mechanisms with EmbraceMutex in several areas to improve concurrency handling.

--

NOTE: I've found that the use of `@ThreadSafe` is making access safe, much like the `atomic` property attribute in Objective-C. It's not however making access the contents safe which is what we usually want. In other words, it makes the access to the "pointer" safe, but not using that pointer since the usage point is outside of the scope of the lock.


ArielDemarco (approved):
I like this. Maybe there are two things we can discuss as a team regarding standardization:
1. How we want to name internal structs that “wrap” multiple properties/variables from an object. (For example, in this PR there’s `ProtectedData`, which is more generic, and elsewhere there’s `LoggerData`, which is more tightly scoped to the specific context).
2. Whether we want to use the `_` prefix pattern for those internal thread-safe variables (even if they’re already marked as `private`).

ArielDemarco (approved):

PR #237, opened 2025-05-14
Implementing custom critical logs

NachoEmbrace:
* Added an `OSLog` implementation in our internal logger.
* There's a single subsystem with 2 categories: `internal` and `custom-export`.
* All the existing `LogLevels` belong to the `internal` category.
* Added 2 new `LogLevels` (`startup` and `critical`) that go to the `custom-export` category.
* If any point during the current process there's a `critical` log, all the entries in that category get dumped into a file.
* The next time the SDK is started, the contents of that file are sent as a single fatal OTel log.
* This log won't have any resources or metadata (with the exception of the resources `app_bundle_id`, `os_name` and `sdk_version` which will aways have a value).

Note:
Since my goal was to have the SDK send these logs ASAP before initializing most things, I had to move stuff around in the constructor of the `Embrace` class.
`EmbraceUpload` needed the device identifier to be constructed, which meant we had to wait for `EmbraceStorage` to be initialized.
Since the storage is one of the mains reasons we are doing this, I removed this hard requirement by saving the device identifier in a file instead of our database.

With these changes the order of operations goes like this:
* The device identifier is retrieved from the new file, or the keychain.
* `EmbraceUpload` is initialized.
* The SDK checks if the critical logs file is present from the previous process.
* The SDK generates the log payload and sends it immediately.
* The SDK continues with the initialization.
* `EmbraceStorage` is initialized.
* The device identifier is fetched and then passed to `EmbraceUpload`.

Also note that this happens before the SDK is even fully setup or started.
I think we need to be pretty aggressive so we can get eyes on issues that prevent the SDK to work entirely.

Note 2:
On phones on iOS 14 and below, the export functionality doesn't work due to missing some necessary APIs.


Sources/EmbraceCommonInternal/InternalLogger.swift
```diff
@@ -19,4 +19,10 @@ import Foundation
 
     @discardableResult @objc func error(_ message: String, attributes: [String: String]) -> Bool
```
naftaly, line 20: While you're here, I think we could remove the `Bool` return value and get rid of the `@discardableResults`, unless they have some great meaning, they remove clarity from the API.

Also, could you only keep the full `log` in the protocol, and have an extension on the protocol that simply implements all sugar.

```
protocol InternalLogger: AnyObject {
 func log(_ LogLevel: level, _ message: String)
}

extension InternalLogger {
 func startup(_ message: String) {
  log(.startup, message)
 }
}
```

```

File: Sources/EmbraceCore/Embrace.swift [hunk 2]
```diff
@@ -148,23 +151,48 @@ To start the SDK you first need to configure it using an `Embrace.Options` insta
          logControllable: LogControllable? = nil,
          embraceStorage: EmbraceStorage? = nil) throws {
 
+        // use OSLog in iOS 15+
+        if #available(iOS 15.0, *) {
```
naftaly, line 155: I've always found it a bit odd here to have a static logger on embrace, why not just set it on the client? And for thread safety, set it only once in here, not directly on the static member.

Sources/EmbraceCore/Embrace.swift
```diff
@@ -148,23 +151,48 @@ To start the SDK you first need to configure it using an `Embrace.Options` insta
...
+
         self.options = options
         self.logLevel = options.logLevel
 
+        // initialize upload module
+        self.upload = Embrace.createUpload(options: options)
+
+        // send critical logs from previous session
+        if #available(iOS 15.0, *) {
+            UnsentDataHandler.sendCriticalLogs(fileUrl: EmbraceFileSystem.criticalLogsURL(), upload: upload)
```
naftaly, line 167: What do you think about having the UnsentDataHandler take care of its business internally. It's feels a bit odd for this to happen in the init of the main Embrace class. Maybe we need a type of task scheduler for jobs that need to be taken care of at startup.

Sources/EmbraceCore/Embrace.swift
```diff
@@ -271,6 +300,12 @@ To start the SDK you first need to configure it using an `Embrace.Options` insta
                     // retry any remaining cached upload data
                     self?.upload?.retryCachedData()
                 }
+
+                if let appId = options.appId {
```
naftaly, line 304: why wouldn't we have an appId here?

Sources/EmbraceCore/Internal/Logs/Loggers/BaseInternalLogger.swift
```diff
@@ -59,56 +59,78 @@ class DefaultInternalLogger: InternalLogger {
         currentSession = nil
     }
 
-    @discardableResult func log(level: LogLevel, message: String, attributes: [String: String]) -> Bool {
+    func output(_ message: String, level: LogLevel, customExport: Bool) {
+        print(message)
+    }
+
+    @discardableResult func log(
```
naftaly, line 66: I wonder what you think of doing something a bit more comprehensive here. Have you seen how swift-log works. It's basically a main logging interface to which you can add handlers for different kinds of logs (Otel, print, file, ...) and they get called in a loop and can do whatever they need to do.

Although was you have here works, it can get dirty pretty quick if we start adding more reasons or places to log.

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
```diff
@@ -0,0 +1,95 @@
...
+                .getEntries(at: position)
+                .compactMap { $0 as? OSLogEntryLog }
+                .filter { $0.subsystem == subsystem && $0.category == customExportCategory }
+                .map { "[\($0.date.formatted(.iso8601))] \($0.composedMessage)" }
+
+            // remove existing file, if any
+            try? FileManager.default.removeItem(at: fileUrl)
+
+            // write new log file
+            let log = entries.joined(separator: "\n")
```
naftaly, line 71: this could get very big in memory and cause OOMs, maybe write it entry by entry or see if there's a built in OSLog exporter to file.

Also, it's not clear what queue we're on, but make sure this is off the main queue.

Sources/EmbraceCore/Session/DataRecovery/UnsentDataHandler.swift
```diff
@@ -273,4 +274,43 @@ class UnsentDataHandler {
...
+
+        guard let upload = upload,
+              let fileUrl = fileUrl else {
+            return
+        }
+
+        // always remove the logs from previous session
+        defer { try? FileManager.default.removeItem(at: fileUrl) }
+
+        guard let logs = try? String(contentsOf: fileUrl) else {
```
naftaly, line 288: Same as my comment above. Simply loading all the contents like this can get very intense on resources and cause memory issues. Do we have the facilities to upload a file on disk and let the OS do the grunt work?

Sources/EmbraceCore/Session/DataRecovery/UnsentDataHandler.swift
```diff
@@ -273,4 +274,43 @@ class UnsentDataHandler {
...
+            body: logs,
+            attributes: attributes,
+            storage: nil,
+            sessionId: nil
+        )
+
+        // send log
+        do {
+            let payloadData = try JSONEncoder().encode(payload).gzipped()
+            upload.uploadLog(id: id, data: payloadData, completion: nil)
```
naftaly, line 311: Do we care if this fails or errors out in any way?

naftaly (commented):
Definitely love all the ideas and where this is headed. I think there are a few places that could use a bit more polishing.

naftaly:
fwiw, I just learnt we can support a lower OS version by using `os_log` like this:

```
os_log(.debug, log: log, "%{public}@", message)
```

https://iosdevelopers.slack.com/archives/C031X84F6/p1747264276800289?thread_ts=1747261368.323409&cid=C031X84F6

Sources/EmbraceCore/Embrace.swift
NachoEmbrace: I wouldn't be opposed to having a more robust system in place, however for this particular thing, it was done this way deliberately. 
I want this to happen as soon as possible before anything else is initialized. So with this in mind, I have a hard time thinking a way of doing this in a more elegant way.

Sources/EmbraceCore/Internal/Logs/Loggers/BaseInternalLogger.swift
NachoEmbrace: Not sure exactly what you mean here.
I don't think we need anything special for the way we use internal logs.
Do you have an example of how things can get dirty?

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
NachoEmbrace: I don't think there's a built in OSLog feature to do something like this, or at least I didn't find one.

Regarding the memory concern, you're right about that. I'll probably address it by writing the file line by line.
This along with dispatching the process to a queue should be good enough.

Sources/EmbraceCore/Session/DataRecovery/UnsentDataHandler.swift
NachoEmbrace: Not really, worst case we are in the same spot as today.

Sources/EmbraceCore/Embrace.swift
NachoEmbrace: There's not really a good reason, but to change it I'd like to have a better system.
I don't really want parts within `EmbraceCore` to start doing `Embrace.client?.logger...`.
We are already doing this in some places (to access the client) and we should to get rid of it (it's bad for tests).
Don't really disagree but I'd probably do this in a separate PR.

Sources/EmbraceCommonInternal/InternalLogger.swift
NachoEmbrace: It was done this way because `LogLevel` is public in `EmbraceCore` (the user can configure the verbosity through the Embrace client in code).
The protocol definition is in `EmbraceCoreInternal` since it's used by all/most other modules, and has no visibility on `LogLevel`.

naftaly: It's still a lot of repeated code. If we can, lets find a way to get around the problem you mentioned.

Sources/EmbraceCore/Embrace.swift
```diff
@@ -151,20 +154,40 @@ To start the SDK you first need to configure it using an `Embrace.Options` insta
         self.options = options
         self.logLevel = options.logLevel
 
+        // initialize upload module
+        self.upload = Embrace.createUpload(options: options)
+
+        // send critical logs from previous session
+        if #available(iOS 15.0, *) {
+            UnsentDataHandler.sendCriticalLogs(fileUrl: EmbraceFileSystem.criticalLogsURL, upload: upload)
```
naftaly, line 162: To clean this up a bit, I'd push the #available macro into `sendCriticalLogs`.

Sources/EmbraceCore/Internal/Logs/Loggers/BaseInternalLogger.swift
naftaly: I mean you have 3 different destinations now: Otel, stdout and critical. It seems like there might be an architectural decision to be made. Something around having classes for each destination and just iterating through them and pushing the data instead of specialized functions for each.

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
```diff
@@ -0,0 +1,169 @@
...
+
+class DefaultInternalLogger: BaseInternalLogger {
+
+    let subsystem: String = "com.embrace.logger"
+    let defaultCategory: String = "internal"
+    let customExportCategory: String
+    let customExportByteCountLimit: Int
+
+    @ThreadSafe
+    var defaultLogger: OSLog
```
naftaly, line 23: Does all of this really need the @ThreadSafe wrapper? It's all private and most cannot be mutated after init. Maybe the queue could be used for that instead if protection is needed.

Sources/EmbraceCore/Internal/Logs/Loggers/BaseInternalLogger.swift
NachoEmbrace: I think the main goal of this class is for us to print our logs without really caring what is being done to them.

OTel is just there in order to send the data to our servers, we don't really use this class to create OTel logs, we use this class to do our internal logging and it so happens that some of them get transformed into OTel data (with limits applied).

The critical logs can be considered a feature added on top that essentially adds new log levels and distributes those logs differently.

I don't really think we need to do much more in terms of architecture here. Also the features got split in between the `BaseInternalLogger` and `DefaultInternalLogger` which helps with readability. (it was split initially since the whole OSLog stuff was put behind iOS 15 being available, which is technically not needed anymore; but I figured it was fine anyways just to not have 1 huge class).

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
NachoEmbrace: Yea not needed really, I'll remove it.

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
```diff
@@ -0,0 +1,169 @@
...
+    @ThreadSafe
+    var customExportLogger: OSLog
+
+    let exportFilePath: URL?
+
+    @ThreadSafe
+    var exporting: Bool = false
+    @ThreadSafe
+    var exportByteCount: Int = 0
```
ArielDemarco, line 32: Seems this should be renamed to `exportByCount`

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
```diff
@@ -0,0 +1,169 @@
...
+import EmbraceConfigInternal
+import EmbraceConfiguration
+#endif
+
+class DefaultInternalLogger: BaseInternalLogger {
+
+    let subsystem: String = "com.embrace.logger"
+    let defaultCategory: String = "internal"
+    let customExportCategory: String
+    let customExportByteCountLimit: Int
```
ArielDemarco, line 20: Seems this should be renamed to `customExportByCountLimit`

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
```diff
@@ -0,0 +1,169 @@
...
+                // fetch log entries
+                let entries: [OSLogEntryLog] = try store
+                    .getEntries(at: position)
+                    .compactMap { $0 as? OSLogEntryLog }
+                    .filter { $0.subsystem == self.subsystem && $0.category == self.customExportCategory }
+
+                // create file if needed
+                if !FileManager.default.fileExists(atPath: fileUrl.path) {
+                    FileManager.default.createFile(atPath: fileUrl.path, contents: nil)
+                }
```
ArielDemarco, line 107: Do we know if, at this point, the "root" directory for embrace was created? if that's not the case, i'd also check if the folder exists, and if it doesn't, create it. As far as I know `createFile` fails if the intermediate folders do not exist

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
```diff
@@ -0,0 +1,169 @@
...
+    let exportFilePath: URL?
+
+    @ThreadSafe
+    var exporting: Bool = false
+    @ThreadSafe
+    var exportByteCount: Int = 0
+    @ThreadSafe
+    var exportLimitReached: Bool = false
+    @ThreadSafe
+    var lastExportDate: Date?
```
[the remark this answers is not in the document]
NachoEmbrace: That wasn't merged when I started this, if it was merged already then sure.

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
NachoEmbrace: It's the amount of bytes of the export file, the name seems correct no?

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
NachoEmbrace: Good catch

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
ArielDemarco: Oh I misunderstood it; I thought it was the count of the amount of times it was logged. Ignore my suggestion 😂

Sources/EmbraceUploadInternal/EmbraceUpload.swift
```diff
@@ -18,6 +18,10 @@ public class EmbraceUpload: EmbraceLogUploader {
     public private(set) var options: Options
     public private(set) var logger: InternalLogger
     public private(set) var queue: DispatchQueue
+
+    @ThreadSafe
+    public var deviceId: String = "0"
+
```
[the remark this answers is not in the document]
NachoEmbrace: Changed device id so its stored in an isolated file instead of our core data database.
This basically ensures we'll always have a device id before creating EmbraceUpload, which means I can revert all these changes here.

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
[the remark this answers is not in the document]
NachoEmbrace: Ended up just using the queue I had.

ArielDemarco (approved):

PR #252, opened 2025-05-30
[EMBR-7753][UI_TESTS_APP] Update Embrace Startup Spans tests

ferdraghi:
Startup procedure has been updated so this PR updates their tests for the UI Tests App


ferdraghi:
> @ferdraghi can we check for children spans? E.g Now the startup trace contains the emb-sdk-start. Could we check that the parent span contains all of those? Maybe should be part of a broader PR tho (so i'll approve as PR LGTM)

I'll create a ticket and get started on it.

PR #261, opened 2025-06-09
[EASY] Missing header in umbrella

naftaly:
There was a missing header in an umbrella header which might have been leading to requiring some hoops to import the module.


ArielDemarco:
This fix was already done when merged the `release/6.11.0` branch. Closing

PR #263, opened 2025-06-10
Background task wrapper for core data operations

NachoEmbrace:
We've seen a lot of app terminations due to a deadlock.
We suspect this happens when our core data operations are performed in the background and the OS suspends the app.

This PR wraps all of our core data operations inside background tasks which should prevent the OS from suspending the app until the operations are done.
This is probably not gonna fix 100% of the cases, but should fix the vast majority of them.

### DIFF
--- a/Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
+++ b/Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
@@ -18,15 +18,31 @@ class BackgroundTaskWrapper {
     private var taskID: UIBackgroundTaskIdentifier
 
     init?(name: String, logger: InternalLogger) {
+        // This should never be called from the main thread
+        // since we are checking for `UIApplication.shared.backgroundTimeRemaining`.
+        //
+        // Note: In the current context of things, this class is only used internally
+        // within this module, and it's always called from the core data wrapper context thread
+        // which should never be the main thread.
+        // Leaving the check anyways just in case.
+        if Thread.isMainThread {
+            return nil
+        }
+
+        // do not create task if there's not enough time until suspension
+        if UIApplication.shared.backgroundTimeRemaining <= 5 {
+            return nil
+        }
+
         self.name = name
         self.taskID = .invalid
 
-        let taskID = UIApplication.shared.beginBackgroundTask(withName: name) {
+        let taskID = UIApplication.shared.beginBackgroundTask(withName: name) { [weak self] in
             logger.critical("Background task \(name) cancelled!")
-            self.endTask()
+            self?.endTask()
         }
 
-        // handle weird case where the task can't be created
+        // handle case where the task can't be created
         if taskID == .invalid {
             return nil
         }
--- a/Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
+++ b/Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
@@ -6,7 +6,9 @@ import Foundation
 
 #if canImport(UIKit) && !os(watchOS)
 import UIKit
+#if !EMBRACE_COCOAPOD_BUILDING_SDK
 import EmbraceCommonInternal
+#endif
 
 /// This class is a wrapper around `UIApplication.shared.beginBackgroundTask`.
 /// Based off https://developer.apple.com/forums/thread/85066 and https://developer.apple.com/forums/thread/729335
--- a/Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
+++ b/Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
@@ -1,0 +1,78 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+#if canImport(UIKit) && !os(watchOS)
+import UIKit
+import EmbraceCommonInternal
+
+/// This class is a wrapper around `UIApplication.shared.beginBackgroundTask`.
+/// Based off https://developer.apple.com/forums/thread/85066 and https://developer.apple.com/forums/thread/729335
+class BackgroundTaskWrapper {
+
+    let name: String
+    private var taskID: UIBackgroundTaskIdentifier
+
+    init?(name: String, logger: InternalLogger) {
+        self.name = name
+        self.taskID = .invalid
+
+        let taskID = UIApplication.shared.beginBackgroundTask(withName: name) {
+            logger.critical("Background task \(name) cancelled!")
+            self.endTask()
+        }
+
+        // handle weird case where the task can't be created
+        if taskID == .invalid {
+            return nil
+        }
+        self.taskID = taskID
+    }
+
+    deinit {
+        self.endTask()
+    }
+
+    func finish() {
+        self.endTask()
+    }
+
+    private func endTask() {
+        guard self.taskID != .invalid else {
+            return
+        }
+
+        UIApplication.shared.endBackgroundTask(self.taskID)
+        self.taskID = .invalid
+    }
+}
+
+#else
+
+// TODO: Implement WatchOS Version
+class BackgroundTaskWrapper {
+
+    let name: String
+    private var taskID: UIBackgroundTaskIdentifier
+
+    init(name: String) {
+        self.name = name
+        self.taskID = .invalid
+    }
+
+    deinit {
+        self.endTask()
+    }
+
+    func finish() {
+        self.endTask()
+    }
+
+    private func endTask() {
+
+    }
+}
+
+#endif
--- a/Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
+++ b/Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
@@ -26,11 +26,13 @@ class BackgroundTaskWrapper {
         // which should never be the main thread.
         // Leaving the check anyways just in case.
         if Thread.isMainThread {
+            logger.critical("Failed to create background task \(name), called from main thread!")
             return nil
         }
 
         // do not create task if there's not enough time until suspension
         if UIApplication.shared.backgroundTimeRemaining <= 5 {
+            logger.critical("Failed to create background task \(name), not enough time!")
             return nil
         }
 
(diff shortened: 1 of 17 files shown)


Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
```diff
@@ -0,0 +1,78 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+#if canImport(UIKit) && !os(watchOS)
+import UIKit
+import EmbraceCommonInternal
```
ArielDemarco, line 10: Wrap this with `#if !EMBRACE_COCOAPOD_BUILDING_SDK` like this:
```suggestion
#if !EMBRACE_COCOAPOD_BUILDING_SDK
import EmbraceCommonInternal
#endif
```

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
```diff
@@ -0,0 +1,80 @@
...
+        self.name = name
+        self.taskID = .invalid
+
+        let taskID = UIApplication.shared.beginBackgroundTask(withName: name) {
+            logger.critical("Background task \(name) cancelled!")
+            self.endTask()
+        }
+
+        // handle weird case where the task can't be created
+        if taskID == .invalid {
```
ArielDemarco, line 48: Maybe add a `critical` log here?

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
```diff
@@ -0,0 +1,80 @@
...
+class BackgroundTaskWrapper {
+
+    let name: String
+    private var taskID: UIBackgroundTaskIdentifier
+
+    init?(name: String, logger: InternalLogger) {
+        self.name = name
+        self.taskID = .invalid
+
+        let taskID = UIApplication.shared.beginBackgroundTask(withName: name) {
```
naftaly: From my experience at Meta (and there's a Apple WWDC video out there with this info in it as well), if there <= 5 seconds of `backgroundTimeRemaining` and you create a task, you'll never get the expiration callback and you're 100% going to crash the app. So check for that. Also, if you call `backgroundTimeRemaining` on the main thread you're likely going to cause hangs, so don't, or at least cache the value and only update that cache if it's been N seconds since the last time you checked it.

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
naftaly: This is going to happen a lot more than you think.

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
```diff
@@ -0,0 +1,80 @@
...
+            logger.critical("Background task \(name) cancelled!")
+            self.endTask()
```
naftaly: In most cases, this will never be called, so make sure that by calling end from somewhere else it actually releases the block and the ref to `self` at the same time.

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    let name: String
+    private var taskID: UIBackgroundTaskIdentifier
+
+    init?(name: String, logger: InternalLogger) {
+        self.name = name
+        self.taskID = .invalid
+
+        let taskID = UIApplication.shared.beginBackgroundTask(withName: name) {
+            logger.critical("Background task \(name) cancelled!")
```
naftaly, line 43: This is actually "expiration". You have a very limited amount of time to cancel your work here and end the task. I don't see anything around ensuring the "task" that is being wrapped in this gets notified to cancel, so basically things will still crash in some (probably most) circumstances.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -17,9 +17,12 @@ public class CoreDataWrapper {
 
     let logger: InternalLogger
 
+    private let isTesting: Bool
+
     public init(options: CoreDataWrapper.Options, logger: InternalLogger) throws {
```
naftaly, line 22: Unrelated, but it would be great if you could pass in the type of store here instead of embedding testing information into the CoreData classes.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -104,17 +107,41 @@ public class CoreDataWrapper {
         }
     }
 
+    /// Synchronously performs the given block on the current context.
+    /// This will also create a background task to perform the operation.
+    /// If the background task can't be created, the block will be called without a context.
+    public func performOperation(name: String, _ block: (NSManagedObjectContext?) -> Void) {
+
+        if isTesting {
```
naftaly: You're not actually testing if you do things totally differently than what you do in prod.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -104,17 +107,41 @@ public class CoreDataWrapper {
...
+        } else {
+            context.performAndWait {
+                let taskName = options.storageMechanism.name + "_" + name
+                guard let task = BackgroundTaskWrapper(name: taskName, logger: logger) else {
+                    logger.critical("Failed to create background task \(taskName)!")
+                    block(nil)
+                    return
+                }
+
+                block(context)
```
naftaly, line 127: What happens if the background task assertion expires? How do you stop the CoreData task before ending the task?

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
NachoEmbrace: As discussed internally, due to the complexity of cancelling ongoing CoreData operations (we're not even sure if that's gonna be feasible 100%), we're gonna do a first pass where we basically "ignore" the expiration of the tasks and just hope that having them will be already good enough for the vast majority of cases.

We'll leave logs behind and monitor how this is working. Depending on what we see we might revisit this and try to figure out a fool proof way of dealing with the expiration callback.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
NachoEmbrace: That was just me doing it quickly to get the draft going.
I've added a new "enableBackgroundTasks" boolean to the appropriate configuration objects.

That being said, from what I've seen, background tasks are never created during tests, which means this is impossible to test automatically with XCTest anyways.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
NachoEmbrace: Addressed in https://github.com/embrace-io/embrace-apple-sdk/pull/263#discussion_r2155204310

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
NachoEmbrace: Added the check!

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
```diff
@@ -0,0 +1,99 @@
...
+        // do not create task if there's not enough time until suspension
+        if UIApplication.shared.backgroundTimeRemaining <= 5 {
+            logger.critical("Failed to create background task \(name), not enough time!")
+            return nil
+        }
+
+        self.name = name
+        self.taskID = .invalid
+
+        let taskID = UIApplication.shared.beginBackgroundTask(withName: name) { [weak self] in
```
naftaly, line 42: if self == nil, we won't call end which means we'll crash right then and there.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -104,17 +107,40 @@ public class CoreDataWrapper {
...
+
+        if options.enableBackgroundTasks == false {
+            context.performAndWait {
+                block(context)
+            }
+        } else {
+            context.performAndWait {
+                let taskName = options.storageMechanism.name + "_" + name
+                guard let task = BackgroundTaskWrapper(name: taskName, logger: logger) else {
+                    block(nil)
```
naftaly, line 123: what happens if it doesn't have a context?

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -104,17 +107,40 @@ public class CoreDataWrapper {
...
+                    return
+                }
+
+                block(context)
+                task.finish()
```
naftaly, line 128: prefer if you use defer for this instead.

naftaly (approved):

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
NachoEmbrace: Since we are calling `endTask` on `deinit`, wouldn't we always call it no matter what?

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
NachoEmbrace: We've decided to let things silently fail for now (not that silent since we are generating critical logs for these tasks).
Once we see how the SDK is behaving and have enough data we'll potentially look in changing this logic.

To answer the question directly, context being nil means the core data operations will return 0 objects or not do anything, essentially.

naftaly: Does that mean we could lose data if terminated soon afterwards?

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
naftaly: I guess so. That's an interesting thought though, could we still receive the callback even if the wrapper has been deinited and task ended.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
NachoEmbrace: Potentially yes.

PR #264, opened 2025-06-12
[EMBR-7958] Fix potential race condition on First Frame Time capture

ferdraghi:
There's a potential race condition when setting up the spans for startup instrumentation. If the first frame time was already captured when `buildMainSpans()` is called, then the spans would never terminate because the chain of events to call `endSpans()` would not trigger. This PR checks if the time already exists, it terminates the spans right after creating them.

### DIFF
--- a/Sources/EmbraceCore/Internal/Startup/StartupInstrumentation.swift
+++ b/Sources/EmbraceCore/Internal/Startup/StartupInstrumentation.swift
@@ -99,6 +99,10 @@ public class StartupInstrumentation: NSObject {
             $0.rootSpan = parent
             $0.firstFrameSpan = firstFrameSpan
         }
+
+        if let firstFrameTime = provider.firstFrameTime {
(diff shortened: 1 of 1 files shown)


Tests/EmbraceCoreTests/Internal/Startup/StartupInstrumentationTests.swift
```diff
@@ -223,6 +223,7 @@ class StartupInstrumentationTests: XCTestCase {
     }
 
     func test_addAttributesToTrace() {
+        provider.firstFrameTime = nil
```
ferdraghi, line 226: First Frame is captured as soon as the test starts so there is already a time set on the Provider, which means the span will be closed immediately after creation and thus the addition of attributes on this test would fail. The solution is to remove the time captured, build the spans and having them not close, and then add the attributes and set an arbitrary firstFrameTime on the provider and this will kick off the span closing function.

PR #269, opened 2025-06-18
[EMBR-7885][UI_TESTS_APP] Update Startup Tests for new Startup Instrumentation system.

ferdraghi:
Implements new tests for the new Startup Instrumentation system.


Examples/EmbraceIOTestApp/EmbraceIOTestApp/Tests/EmbraceInit/EmbraceInitScreen.swift
```diff
@@ -82,6 +108,9 @@ private extension EmbraceInitScreen {
                               export: .init(spanExporter: dataCollector.spanExporter, logExporter: dataCollector.logExporter))
                 ).start()
             viewModel.showProgressview = false
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+                NotificationCenter.default.post(name: NSNotification.Name("UIApplicationDidFinishLaunchingNotification"), object: nil)
+            }
```
ArielDemarco, line 113: One thing to mention: startup instrumentation will work without the `didFinishLaunching` notification. Obviously, some spans wont be there. Same thing goes with the first frame rendered

ferdraghi: yes, I forced those events in order to kick off the procedures that export those spans. In a regular app they are expected to happen automatically but since Embrace is being initialized in the "wrong place" on the testing app, some events have to be forced.

ferdraghi: I can add a switch for these events if you think it's worth it.

ArielDemarco: na, I think that if you considered it while coding it, everything's fine

PR #274, opened 2025-06-20
[6.12.patch] Fix a few bg task issue

naftaly:
Introduces BackgroundTimeTracker to cache backgroundTimeRemaining checks and prevent creating background tasks when insufficient time remains. This reduces expensive calls to UIApplication.shared.backgroundTimeRemaining and improves reliability of background task creation.


Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
```diff
@@ -12,79 +12,168 @@ import EmbraceCommonInternal
...
+    public static let timeout: UIBackgroundTaskIdentifier = UIBackgroundTaskIdentifier(rawValue: -1024)
+    
+    public static let noApp: UIBackgroundTaskIdentifier = UIBackgroundTaskIdentifier(rawValue: -1025)
+    
+    var stringValue: String {
+        if self == .invalid { return "bgtask: invalid" }
+        if self == .timeout { return "bgtask: timeout" }
+        if self == .noApp { return "bgtask: noApp" }
+        return "bgtask: \(self)"
+    }
```
NachoEmbrace: What is this for?

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
```diff
@@ -12,79 +12,168 @@ import EmbraceCommonInternal
...
+    private var knownTimeRemain: TimeInterval = 0
+    private var lastTimeRemainCheck: CFAbsoluteTime = 0
+    private let lock: UnfairLock = UnfairLock()
+    
+    func canStartTask() -> Bool {
+        
+        // Allow task starting when there's no app.
+        guard let app else {
+            return true
+        }
```
NachoEmbrace: This will never be executed since you are checking for app != nil before calling this method.

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
naftaly: logging

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
naftaly: but if we call it form somewhere else there's no guarantee we'll have checked beforehand. It's a part of the logic.

NachoEmbrace: But is that method intended to be called from anywhere else?

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
NachoEmbrace: Did I miss some logging or do you mean, potentially to log something?

NachoEmbrace (approved):

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
naftaly: Oh, right :) I was using it to be very descriptive about the tasks in the logging I had earlier but removed. so "potentially to log something". I can remove it if you prefer.

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
naftaly: Yes, it can be called from anywhere else as available. also, fwiw, if you look at the code for `var app`, it can easily go from nil to a valid app so it's definitely worth checking again if this is what you're worried about.

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
```diff
@@ -12,79 +12,152 @@ import EmbraceCommonInternal
...
-
+    
     func finish() {
         self.endTask()
     }
-
+    
     private func endTask() {
-        guard self.taskID != .invalid else {
+        guard taskID.rawValue > 0 else {
```
ArielDemarco: Shall we call `Self.taskProvider.endBackgroundTask` when `.noApp` is the `taskID`? the value will be `9223372036854775805`

naftaly: it doesn't actually matter, we can call it with any task that is valid or not and it'll just be a noop.

ArielDemarco: So, if I understand correctly: considering that we have two tasks, both of them `.noApp`. This code will end up calling `Self.taskProvider.endBackgroundTask("9223372036854775805")` which will end up calling the `UIApplication.shared.endBackgroundTask("9223372036854775805"). As the task doesn't actually exist, it wont cause any issue. Am I right?

naftaly: correct. But I can make sure we don't call it anyway to make the code cleaner.

PR #279, opened 2025-06-25
Improve CoreData Implementations

naftaly:
This PR improves our CoreData usage by making sure we only save changes when necessary, using the hasChanges check. Additionally, the CoreData logic is now streamlined—all operations are handled through a single function, making the code cleaner and easier to maintain.

### DIFF
--- a/Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
+++ b/Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
@@ -2,19 +2,19 @@
 //  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
 //
 
-import Foundation
 import CoreData
+import Foundation
+import UIKit
+
 #if !EMBRACE_COCOAPOD_BUILDING_SDK
 import EmbraceCommonInternal
 #endif
 
 public class CoreDataWrapper {
-
     public let options: CoreDataWrapper.Options
 
     var container: NSPersistentContainer!
     public private(set) var context: NSManagedObjectContext!
-
     let logger: InternalLogger
 
     private let isTesting: Bool
@@ -22,7 +22,7 @@ public class CoreDataWrapper {
     public init(options: CoreDataWrapper.Options, logger: InternalLogger) throws {
         self.options = options
         self.logger = logger
-        self.isTesting = ProcessInfo.processInfo.isTesting
+        isTesting = ProcessInfo.processInfo.isTesting
 
         // create model
         let model = NSManagedObjectModel()
@@ -30,20 +30,20 @@ public class CoreDataWrapper {
 
         // create container
         let name = options.storageMechanism.name
-        self.container = NSPersistentContainer(name: name, managedObjectModel: model)
+        container = NSPersistentContainer(name: name, managedObjectModel: model)
 
         // force db on memory during tests
         if isTesting {
             let description = NSPersistentStoreDescription()
             description.type = NSInMemoryStoreType
-            self.container.persistentStoreDescriptions = [description]
+            container.persistentStoreDescriptions = [description]
 
         } else {
             switch options.storageMechanism {
             case .inMemory:
                 let description = NSPersistentStoreDescription()
                 description.type = NSInMemoryStoreType
-                self.container.persistentStoreDescriptions = [description]
+                container.persistentStoreDescriptions = [description]
 
             case let .onDisk(_, baseURL):
                 try FileManager.default.createDirectory(at: baseURL, withIntermediateDirectories: true)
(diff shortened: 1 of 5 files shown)


Sources/EmbraceCommonInternal/BackgroundTasks/BackgroundTaskWrapper.swift
```diff
@@ -0,0 +1,251 @@
...
+        
+        return lock.locked {
+            if let curTask = currentTaskID {
+                currentTaskRefCount += 1
+                return curTask
+            }
+            
+            var newTask: UIBackgroundTaskIdentifier? = nil
+            newTask = app.beginBackgroundTask(withName: taskName) { [self] in
+                lock.locked {
```
ArielDemarco: Question; would this open the possibility of a deadlock by locking two times the same lock if that happens on the same thread?
Edit: I don't know if the ".invalid" block happens always on a specific thread, or something.

naftaly: I don't think so. The locks are all as small as possible and don't call out to external code. Can you give me an example of what might cause the deadlock? 

Small hint that the expiring block is escaping.

Tests/EmbraceCoreTests/Payload/LogPayloadBuilderTests.swift
```diff
@@ -110,7 +110,7 @@ class LogPayloadBuilderTests: XCTestCase {
         // then the payload is correct
         XCTAssertEqual(payload.resource.appVersion, "1.0.0")
         XCTAssertEqual(payload.metadata.username, "test")
-        XCTAssertEqual(payload.metadata.personas, ["tag1", "tag2"])
+        XCTAssertEqual(payload.metadata.personas.sorted(), ["tag1", "tag2"].sorted())
```
jsflax: Clearly unrelated to this PR, but why is this necessary? Arrays should be ordered.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -64,203 +68,316 @@ public class CoreDataWrapper {
...
+}
+
+// MARK: - Fetch
 
+extension CoreDataWrapper {
+    
     /// Synchronously fetches the records that satisfy the given request
     public func fetch<T>(withRequest request: NSFetchRequest<T>) -> [T] where T: NSManagedObject {
-
+
```
jsflax: Unnecessary new line.

jsflax (dismissed):
This PR is difficult to review in its current state.

There are a number of risky operations happening and without testing both what exists today and what the changes are here, it's difficult to consider this.

Before we go fixing this problem, we should write a test showing the problem, discuss, and then show that with this PR, the problem is fixed.

naftaly:
@jsflax 

Hey, thanks for taking a look — totally fair that it’s still a bit rough, it’s a draft after all. That said, if there are specific parts that feel hard to review, I’d really appreciate a bit more detail so I can clean things up or add context.

You mentioned some risky operations — totally open to digging into those, but it’d help a lot if you could point out which ones you're referring to so we’re on the same page.

And agreed on the testing front: I’m all for writing a test that clearly shows the issue before we land on a fix. Let’s just make sure we’re aligned on what the problem actually is — happy to sketch that out first.

naftaly:
In order to be much more concise around what this PR actually accomplishes, I've removed a lot of improvements that aren't specifically related to making our CoreData implementation more robust.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -61,206 +61,184 @@ public class CoreDataWrapper {
                 logger.critical("Error initializing CoreData \"\(name)\": \(error.localizedDescription)")
             }
         }
-
-        self.context = NSManagedObjectContext(concurrencyType: .privateQueueConcurrencyType)
-        self.context.persistentStoreCoordinator = self.container.persistentStoreCoordinator
+        
+        context = container.newBackgroundContext()
```
NachoEmbrace, line 66: Is there a difference on doing it this way or is it just cleaner?

naftaly: to my knowledge, Apple recommends doing it this way. I can't find proof in any docs though so we could easily go either way. Otherwise I'd say they're equivalent except for it being a one liner vs. a two liner. I dropped the `self` however which I should put back.

NachoEmbrace (approved):

NachoEmbrace (approved):

PR #281, opened 2025-06-26
[EMBR-8014][UI_TESTS_APP] Added basic checks for OTel semantics.

ferdraghi:
Added a starting point to check for OTel semantics on Attribute names. So far it checks for usages of capital letters (discouraged) and fails the test when otel namespace is detected (this is not allowed).


Examples/EmbraceIOTestApp/EmbraceIOTestApp/Tests/OTel Semantics/OTelSemanticsValidation.swift
```diff
@@ -0,0 +1,37 @@
...
+        var testItems = [TestReportItem]()
+        let names = attributes.keys.sorted()
+
+        for name in names {
+            /// Not a FAIL but OTel recommends names to be lowercase only.
+            /// https://opentelemetry.io/docs/specs/semconv/general/naming/#general-naming-considerations
+            if name.containsUppercase() {
+                testItems.append(.init(target: "Attribute: \(name)", expected: "only lowercase", recorded: "contains uppercase", result: .warning))
+            }
+
```
ArielDemarco, line 21: while performing tests, did you find any attribute name whose name has uppercases?
edit: maybe is good to know so we change it

ferdraghi: yes, "isPrewarmed" fails this test. Although I set it as a "warning" for now since this is more a guideline than a strict rule.

PR #282, opened 2025-06-26
Logs limits

NachoEmbrace:
Log limits were not being applied on the SDK side, this PR addresses that.
I've imitated the same limit logic that is done on the backend. This means we shouldn't send data that is getting dropped anyways now.

These changes also fix metric kit crash logs not being exported (unintended bug that was happening due to how we treat crash logs in a special way).

Note: Couldn't call the class "LogLimits" because it would clash with an OTel type.


naftaly (approved):

Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig/RemoteConfigPayload.swift
```diff
@@ -199,7 +238,11 @@ public struct RemoteConfigPayload: Decodable, Equatable {
         metricKitHangCaptureEnabled = false
 
         swiftUiViewInstrumentationEnabled = true
-        
+
+        logsInfoLimit = 100
+        logsWarningLimit = 200
+        logsErrorLimit = 500
```
ArielDemarco, line 244: Those are the defaults today in the back or other platforms?

Sources/EmbraceConfiguration/LogsLimits.swift
```diff
@@ -0,0 +1,31 @@
+//
+//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+/// LogsLimits manages limits for the logs generated through the SDK
+/// This is broken into the major log severities so each can be managed
+/// These logs will only be emitted if the ``Embrace.logLevel``
```
ArielDemarco: Seems like there's some missing part in this sentence

Sources/EmbraceCore/Internal/Logs/Exporter/StorageEmbraceLogExporter.swift
```diff
@@ -57,4 +101,30 @@ class StorageEmbraceLogExporter: LogRecordExporter {
     func forceFlush(explicitTimeout: TimeInterval?) -> ExportResult {
         .success
     }
+
+    func limitLevel(for severity: LogSeverity?) -> LogLevel {
+        guard let severity = severity else {
+            return .info
+        }
```
ArielDemarco, line 108: I think it'll make code cleaner if we remove this condition, we make the `severity` non-optional, and do the `guard let` in
`func export(logRecords: [ReadableLogRecord], explicitTimeout: TimeInterval?) -> ExportResult {` (inside the `withLock` and, in case we cannot cast the severity to a `LogSeverity` we return `false` to `canExport`.

Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig/RemoteConfigPayload.swift
NachoEmbrace: Back end

Sources/EmbraceCore/Internal/Logs/Exporter/StorageEmbraceLogExporter.swift
NachoEmbrace: Can't really do it since Severity is optional on the log for some reason.

PR #283, opened 2025-06-27
[EMBR-8119] Enable UI Tests Action on CI

ferdraghi:
Summary:

- Adds a workflow file to CI to run the UI Tests App on every PR.
- Tweaks automated UI tests in order to stabilize them: 
-- Waits for elements to appear on screen.
-- Scrolls to elements before tapping them.
-- runs multiple tests on one launch by swiping down the results screen.
-- Corrects logic on Posted Session tests in order to correctly match exported spans to spans included on the payload.
-- Disables UI animations when running tests via environment var on launch.
- Returns project to iOS 18 (Xcode 16 appears to run more stable on CI than xcode 15).
- Removes iOS 18 SwiftUI warnings on previewable code.
- Removes `ui_tests` script file since it's not gonna be used (tests run via workflow file directly)


.github/workflows/run-ui-tests.yml
```diff
@@ -31,6 +34,16 @@ jobs:
...
+      - name: Shutdown Simulators
+        run: |
+            xcrun simctl shutdown all
+      - name: Erase Simulators
+        run: |
+            xcrun simctl erase all
+#      - name: Boot Simulator
+#        run: |
+#          xcrun simctl boot "${{ matrix.device }}"
+
```
ArielDemarco: Shall this be commented?

.github/workflows/run-ui-tests.yml
```diff
@@ -44,36 +57,27 @@ jobs:
...
-          show-code-coverage: false
-          upload-bundles: "never" # Uploading the `xcresult` using the xcresultool action seems to fail
+#      - uses: kishikawakatsumi/xcresulttool@2b37f116695328d5e503ac8dadca0f1545a51da4
+#        if: always()
+#        with:
+#          path: .build/test/output.xcresult
+#          title: "UI Test Results"
+#          show-passed-tests: true
+#          show-code-coverage: false
+#          upload-bundles: "never" # Uploading the `xcresult` using the xcresultool action seems to fail
```
ArielDemarco: I'd totally remove this
```suggestion
```

.github/workflows/run-ui-tests.yml
ferdraghi: This one is still for debate. Turns out you can run the tests "headless" if the simulator hasn't booted and it appears to be a bit more stable than having the simulator actually running. I commented the line for now but I'll be experimenting with it to see if it helps at all. I'd leave it off for now. (I had instances on my local machine where tests would simply never start when running from command line and the simulator was running... but then running the exact same test command without the simulator running would let the tests run... it's a lot of fun and not frustrating at all!)

.github/workflows/run-ui-tests.yml
ferdraghi: yeah probably better to remove this one.

.github/workflows/run-ui-tests.yml
ferdraghi: I uncommented it and seems to be working fine. I'll leave it in and see how it goes.

naftaly:
@ferdraghi you can cd into the repo directory and run `make all` to get all the formatting issues at once.

naftaly (approved):

PR #288, opened 2025-06-28
ci: Pin all 3rd-party GitHub actions

vitaliyf:
.. and some other minor CI housekeeping.


.github/workflows/create-release.yml
```diff
@@ -214,12 +214,11 @@ jobs:
 
     steps:
       - name: Record SDK Version History
-        uses: embrace-io/public-actions/upload-sdk-version@v1
+        uses: embrace-io/public-actions/upload-sdk-version@88167cd1a3fce3418e26c8c842026e6dfab99e41
```
ArielDemarco, line 217: Is there any specific reason why we do this change? considering we're the ones maintaining this repository?

.github/workflows/build-example-apps.yml
```diff
@@ -46,14 +46,14 @@ jobs:
   build-demoobjc-app:
     name: Build Objc Demo App
     timeout-minutes: 20
-    runs-on: macos-14
+    runs-on: macos-latest
     strategy:
       fail-fast: false
       matrix:
         xcode_version: ["15.4"]
```
ArielDemarco: If for some reason, `macos-latest` changes to be `macos-15` this will suddenly stop wokring, as [it doesn't have support](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md#xcode) for Xcode version `15.4`.

Edit: I prefer to make those version changes (e.g. switching the Xcode version) consciously rather than being forced into it because my CI pipeline suddenly stops working

.github/workflows/create-release.yml
vitaliyf: Right, as far as "do we need to pin our own actions" - probably don't truly have to, but CodeQL scanning wants you to pin everything so it doesn't hurt.

.github/workflows/build-example-apps.yml
vitaliyf: Ahh. and are we purposely not upgrading to newer xcode? sooner or later, github will deprecate macos-14 and it'll stop working as well.

.github/workflows/create-release.yml
ArielDemarco: Didn't know the CodeQL stuff. If that's a requirement, im ok :)

.github/workflows/build-example-apps.yml
ArielDemarco: We'll be doing that soon

PR #293, opened 2025-07-02
Run zizmor for static analysis of .github/workflows/

vitaliyf:
See https://github.com/zizmorcore/zizmor

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -19,8 +19,7 @@ on:
         default: false
 
 permissions:
-  id-token: write
-  contents: write
+  contents: read
   packages: read
 
 jobs:
@@ -41,6 +40,7 @@ jobs:
         with:
           fetch-depth: 0
           fetch-tags: true
+          persist-credentials: false
 
       - name: Define Production Readiness
         run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/create-release.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           # If the workflow wasn't triggered manually, extract the version number either from branch or tag
           # (depending the event which triggered the workflow).
           if [ -z "$RC_VERSION" ]; then
-            if [[ "${{ github.ref_type }}" == "branch" ]]; then
+            if [[ "${GITHUB_REF_TYPE}" == "branch" ]]; then
```
ArielDemarco, line 58: Does this work properly? Just in case, I found the other in the docs that's why im asking

vitaliyf: Seems to be - it's autofix by zizmor, based on https://github.com/zizmorcore/zizmor/blob/v1.11.0/crates/zizmor/src/utils.rs#L154-L159

It's listed here: https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables

paolostancato (approved):

ArielDemarco (approved):
LGTM. Test are not failing in reality (the `run-tests` job worked fine)

PR #297, opened 2025-07-11
CoreData Date issues

NachoEmbrace:
This PR attempts to address the crash we've seen regarding non-optional Dates in CoreData.

1. It sets a default value in the model for all non-optional Dates.
2. We noticed there were some places where we were fetching an object in an operation (performAndWait), then reading / modifying that object in another performAndWait block. This PR combines those operations into 1 in all places that makes sense.

### DIFF
--- a/Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
+++ b/Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
@@ -138,7 +138,9 @@ public class CoreDataWrapper {
             }
 
             do {
-                try context.save()
+                if context.hasChanges {
+                    try context.save()
+                }
             } catch {
                 let name = context.name ?? "???"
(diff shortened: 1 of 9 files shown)


Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -138,7 +138,9 @@ public class CoreDataWrapper {
             }
 
             do {
-                try context.save()
+                if context.hasChanges {
```
naftaly, line 141: Should help fix high disk writes and cpu reported by Apple as well, but not completely.

jsflax (approved):

naftaly:
I think the real fix here is putting everything behind the performOperation and this way we don't cause any race conditions so I'd prefer if we didn't add the default values, or the hasChanges so that we can really know what fixes the issue.

ArielDemarco (approved):
LGTM

PR #299, opened 2025-07-11
Add macOS support; fix CADisplayLink bug

jsflax:
Add macOS support. A number of a calls/imports were disallowed on macOS, so I've modified those. I have also enabled macOS support for the BrandGame app.

Fix CADisplayLink bug: CADisplayLink was not being correctly invalidated (on all platforms), making `onFrameUpdate` be called on every frame update, which is extremely often.

This PR also allows us to test macro expansion, which can only be run on macOS.

### DIFF
--- a/Examples/BrandGame/BrandGame.xcodeproj/project.pbxproj
+++ b/Examples/BrandGame/BrandGame.xcodeproj/project.pbxproj
@@ -43,6 +43,13 @@
 		7BCB66282AC4742800ABD33A /* MenuList.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7BCB66272AC4742800ABD33A /* MenuList.swift */; };
 		7BCB662A2AC4767B00ABD33A /* MainMenu.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7BCB66292AC4767B00ABD33A /* MainMenu.swift */; };
 		7BCB662E2AC476BC00ABD33A /* NetworkStressTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7BCB662D2AC476BC00ABD33A /* NetworkStressTest.swift */; };
+		9618109E2E1592B4009CC62C /* EmbraceCore in Frameworks */ = {isa = PBXBuildFile; productRef = 9618109D2E1592B4009CC62C /* EmbraceCore */; };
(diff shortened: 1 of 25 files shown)


Examples/BrandGame/BrandGame.xcodeproj/project.pbxproj
```diff
@@ -828,7 +850,7 @@
 				CURRENT_PROJECT_VERSION = 5;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
 				DEVELOPMENT_ASSET_PATHS = "\"BrandGame/Preview Content\"";
-				DEVELOPMENT_TEAM = L5RVT7J8CV;
+				DEVELOPMENT_TEAM = 2Z8M9MTEVV;
```
NachoEmbrace, line 853: Not really sure if this matters, but probably an unintended change.

Examples/BrandGame/BrandGame.xcodeproj/project.pbxproj
```diff
@@ -860,7 +885,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 5;
 				DEVELOPMENT_ASSET_PATHS = "\"BrandGame/Preview Content\"";
-				DEVELOPMENT_TEAM = L5RVT7J8CV;
+				DEVELOPMENT_TEAM = 2Z8M9MTEVV;
```
NachoEmbrace, line 888: Same here

Sources/EmbraceCoreDataInternal/BackgroundTaskWrapper.swift
```diff
@@ -130,17 +130,41 @@ fileprivate class BackgroundTaskProvider {
     }
 }
 
-#else
-
+#elseif os(watchOS)
+#if !EMBRACE_COCOAPOD_BUILDING_SDK
+import EmbraceCommonInternal
+#endif
```
NachoEmbrace, line 136: If we ever implement a version for WatchOS and MacOS (would it even make sense in MacOS?) we might wanna just do 3 separate files.

Sources/EmbraceCore/Internal/MetricKit/MetricKitHandler.swift
```diff
@@ -26,13 +26,13 @@ import EmbraceCommonInternal
     let sessionLinkGracePeriod: TimeInterval = 5
 
     func install() {
-#if !os(tvOS)
+#if !os(tvOS) && !os(macOS)
```
naftaly, line 29: Not that it matters too much today, but MetricKit is supported on macOS 26.

PR #300, opened 2025-07-14
Adding reset upload cache API

NachoEmbrace:
This PR also adds code that removes any data from previous versions of SDK (v5) on start.

### DIFF
--- a/Sources/EmbraceCore/Internal/Embrace+Setup.swift
+++ b/Sources/EmbraceCore/Internal/Embrace+Setup.swift
@@ -14,6 +14,7 @@ import EmbraceUploadInternal
 import OpenTelemetryApi
 
 extension Embrace {
+
     static func createStorage(options: Embrace.Options) throws -> EmbraceStorage {
 
         let partitionId = options.appId ?? EmbraceFileSystem.defaultPartitionId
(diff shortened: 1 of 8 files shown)


Sources/EmbraceCore/Internal/Embrace+Setup.swift
```diff
@@ -121,5 +130,20 @@ extension Embrace {
             .startSpan()
             .end()
     }
+}
 
+extension Embrace {
+    func cleanUpOldVersionsData() {
+        let urls = EmbraceFileSystem.oldVersionsDirectories()
```
ArielDemarco, line 147: Might be worth using UserDefaults or something like that to do this only once per device?

ArielDemarco (approved):

Sources/EmbraceCore/Internal/Embrace+Setup.swift
NachoEmbrace: The way this was done it also contemplates for future changes in the data.
So if tomorrow we launch a new version of this SDK and decide to use a different version folder, the old one would be removed.

I guess this could be optimized but I don't think its worth it tbh.

PR #303, opened 2025-07-15
Fix crash when finishing network request spans

NachoEmbrace:
We are referencing the URLSessionTask on an asynchronous operation, and inside we access some members of it, like the request, response and data.
This could cause crashes if the task is released from memory.

We've had this exact crash on the previous SDK and was fixed the same way.


Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
```diff
@@ -132,6 +132,11 @@ final class DefaultURLSessionTaskHandler: NSObject, URLSessionTaskHandler {
     }
 
     func finish(task: URLSessionTask, bodySize: Int, error: (any Error)?) {
+        // save a local copy of the task in case it gets released
+        guard let _task = task.copy() as? URLSessionTask else {
```
naftaly:
```suggestion
        guard let taskCopy = task.copy() as? URLSessionTask else {
```

PR #304, opened 2025-07-15
Fixing potential race condition when a session ends

NachoEmbrace:
We've seen a lot of core data crashes related to an internal auto release pool.
Even though we don't understand exactly why that happens, we noticed these crashes happen when a session ends.

At this time, there's a few things happening at the same time:
1. We end the span, which triggers our processor and exporter to re-export the span into the database. This happens asynchronously.
2. We update our session record with the end time and other flags.
3. We fetch the session data and span to build the payload.

After analyzing the crashes, we determined that points 1 and 3 are usually happening at the same time, which was unintended.

This PR changes the process in the `SessionController` and `SingleSpanProcessor` so:
1. We end the OTel span, but don't update it in the database through the exporter anymore (this was done by adding a check so we never export a session span on end).
2. We manually set the end time on the span record in our database synchronously
3. Continue with the process

Even though we don't fully understand the crashes, I think these changes have the potential to address the real issue.

Note: I've also improved the code that builds the session payload by reducing the amount of fetches needed to get the resources, custom properties and persona tags.

### DIFF
--- a/Sources/EmbraceOTelInternal/Trace/Tracer/Span/Processor/SingleSpanProcessor.swift
+++ b/Sources/EmbraceOTelInternal/Trace/Tracer/Span/Processor/SingleSpanProcessor.swift
@@ -74,6 +74,14 @@ public class SingleSpanProcessor: SpanProcessor {
             return
         }
 
+        // Prevent exporting our session spans on end.
+        // This process is handled by the `SessionController` to prevent
+        // race conditions when a session ends and its payload gets built.
+        if let spanType = span.getAttributes()[SpanSemantics.keyEmbraceType]?.description,
+           spanType == SpanType.session.rawValue {
+            return
+        }
+
         var data = span.toSpanData()
         if data.hasEnded && data.status == .unset {
             if let errorCode = data.errorCode {
--- a/Sources/EmbraceStorageInternal/Records/EmbraceStorage+Metadata.swift
+++ b/Sources/EmbraceStorageInternal/Records/EmbraceStorage+Metadata.swift
@@ -329,19 +329,15 @@ extension EmbraceStorage {
         return result
     }
 
-    /// Returns immutable copies of all records with types `.requiredResource` or `.resource` that are tied to a given session id
-    public func fetchResourcesForSessionId(_ sessionId: SessionIdentifier) -> [EmbraceMetadata] {
-
-        guard let session = fetchSession(id: sessionId) else {
-            return []
-        }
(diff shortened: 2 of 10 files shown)


Sources/EmbraceOTelInternal/Trace/Tracer/Span/Processor/SingleSpanProcessor.swift
```diff
@@ -74,6 +74,14 @@ public class SingleSpanProcessor: SpanProcessor {
...
         }
 
+        // Prevent exporting our session spans on end.
+        // This process is handled by the `SessionController` to prevent
+        // race conditions when a session ends and its payload gets built.
+        if let spanType = span.getAttributes()[SpanSemantics.keyEmbraceType]?.description,
+           spanType == SpanType.session.rawValue {
+            return
+        }
+
```
ArielDemarco: I think this should be handled at the exporter level. If we do it at the processor level, all the customer exporters that depend on the session data (in particular, the `endTime`) wont work

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Metadata.swift
```diff
@@ -366,7 +371,7 @@ extension EmbraceStorage {
             type: .and,
             subpredicates: [
                 resourcePredicate(),
-                lifespanPredicate(processId: processId)
+                lifespanPredicate(processId: processId.hex)
```
ArielDemarco, line 374: This was working without the `.hex`? or this change is unnecessary?

Sources/EmbraceOTelInternal/Trace/Tracer/Span/Processor/SingleSpanProcessor.swift
NachoEmbrace: You're right

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Metadata.swift
NachoEmbrace: This internal method now receives the String directly.
If you follow back the chain of method calls, I did it this way so I didn't have to deal with nullable session and process identifiers when building the payloads.

ArielDemarco: That's great. Thanks for the explanation

ArielDemarco (approved):

PR #305, opened 2025-07-15
Potential CoreData fix when saving upload data

NachoEmbrace:
This PR combines two CoreData operations that were happening in the upload module cache in hopes that it fixes some crashes we've seen in these methods.

This is the same paradigm we applied to the main database not long ago.

### DIFF
--- a/Sources/EmbraceUploadInternal/Cache/EmbraceUploadCache.swift
+++ b/Sources/EmbraceUploadInternal/Cache/EmbraceUploadCache.swift
@@ -109,32 +109,26 @@ class EmbraceUploadCache {
 
         var result = false
 
-        // update if it already exists
-        let request = fetchUploadDataRequest(id: id, type: type)
-        coreData.fetchFirstAndPerform(withRequest: request) { [weak self] record in
-            guard let uploadData = record else {
-                return
-            }
+        coreData.performOperation { context in
 
-            uploadData.data = data
-
+            // update if it already exists
+            let request = fetchUploadDataRequest(id: id, type: type)
             do {
-                try self?.coreData.context.save()
-                result = true
+                if let uploadData = try context.fetch(request).first {
+                    uploadData.data = data
+                    try context.save()
+
+                    result = true
+                    return
+                }
             } catch {
-                self?.logger.warning("Error upading upload data:\n\(error.localizedDescription)")
+                logger.warning("Error upading upload data:\n\(error.localizedDescription)")
             }
-        }
 
-        if result {
-            return true
-        }
+            // check limit and delete if necessary
+            checkCountLimit(context)
 
-        // check limit and delete if necessary
-        checkCountLimit()
-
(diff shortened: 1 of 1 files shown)


Sources/EmbraceUploadInternal/Cache/EmbraceUploadCache.swift
```diff
@@ -107,22 +107,33 @@ class EmbraceUploadCache {
     /// - Returns: Boolean indicating if the operation was successful
     @discardableResult func saveUploadData(id: String, type: EmbraceUploadType, data: Data) -> Bool {
 
+        var result = false
+
         // update if it already exists
-        if let record = fetchUploadData(id: id, type: type) {
+        let request = fetchUploadDataRequest(id: id, type: type)
+        coreData.fetchFirstAndPerform(withRequest: request) { [weak self] record in
```
naftaly: You shouldn't need `weak self` here since the closure is always going to be called no matter what.

Sources/EmbraceUploadInternal/Cache/EmbraceUploadCache.swift
```diff
@@ -135,12 +146,10 @@ class EmbraceUploadCache {
 
                 do {
                     try context.save()
+                    result = true
                 } catch {
                     context.delete(record)
```
naftaly, line 145: good to know we could use a transaction API for things like this in the future.

Sources/EmbraceUploadInternal/Cache/EmbraceUploadCache.swift
```diff
@@ -107,22 +107,33 @@ class EmbraceUploadCache {
     /// - Returns: Boolean indicating if the operation was successful
     @discardableResult func saveUploadData(id: String, type: EmbraceUploadType, data: Data) -> Bool {
```
naftaly, line 108: What do you think of putting almost the whole method in a performOperation? This way it is 100% atomic including the `checkCountLimit` and everything else. It's ok to keep the ones you already have, just wrap it up in a bow so nothing else can touch the db while this is processing.

naftaly (commented):
Just a small comment above about taking this one step and making the method truly atomic as far as the db is concerned.

ArielDemarco (approved):
LGTM

Sources/EmbraceUploadInternal/Cache/EmbraceUploadCache.swift
NachoEmbrace: I suppose that would be fine since performAndWait supports re-entrancy, right?

NachoEmbrace: I ended up re-writing the method into a single operation.
All I had to do is pass the context to the `checkCountLimit` method, figured it was gonna look cleaner this way.

PR #307, opened 2025-07-18
Implementing block list for the ViewCaptureService

NachoEmbrace:
Supports local and remote configurations.


Sources/EmbraceCore/Capture/UX/View/ViewCaptureService.swift
```diff
@@ -29,6 +31,8 @@ public final class ViewCaptureService: CaptureService, UIViewControllerHandlerDa
         return options.instrumentFirstRender && Embrace.client?.config.isUiLoadInstrumentationEnabled == true
     }
 
+    var blockList: ViewControllerBlockList
```
ArielDemarco: Considering this will be accessed (probably by the main thread) and modified whenever the config service updates, wouldn't be better to make it protected?

Sources/EmbraceCore/Capture/UX/View/ViewCaptureService.swift
```diff
@@ -51,6 +55,34 @@ public final class ViewCaptureService: CaptureService, UIViewControllerHandlerDa
...
+
+    @objc private func onConfigUpdated(_ notification: Notification) {
+        let config = notification.object as? EmbraceConfigurable
+        updateBlockList(config: config)
+    }
+
+    private func updateBlockList(config: EmbraceConfigurable?) {
+        if let list = config?.uiInstrumentationBlockList {
+            let blockHostingControllers = config?.uiInstrumentationCaptureHostingControllers ?? true
+            blockList = ViewControllerBlockList(names: list, blockHostingControllers: blockHostingControllers)
```
ArielDemarco: Shouldn't you re-include the original types (`self.blockList.types`) in the new `ViewControllerBlockList`? In that way you can have the "user-specified" controllers + the ones that we disable from the remote config

Tests/EmbraceCoreTests/Capture/UX/View/ViewControllerBlockListTests.swift
```diff
@@ -0,0 +1,96 @@
...
+class TestViewController: UIViewController {
+
+}
+
+class TestFakeHostingController: UIViewController {
+
+}
+
+struct TestView: View {
+    var body: some View { }
```
ArielDemarco:
```suggestion
    var body: some View { EmptyView() }
```

Sources/EmbraceCore/Capture/UX/View/ViewCaptureService.swift
NachoEmbrace: We could do that but I figured the remote config should completely override the local one.
Otherwise you could end up in situations where changing the remote config wouldn't be enough and the client might need to make a new release of their app, which defeats the purpose of having a remote config.

ArielDemarco: Fair enough. I think this is something to take into account when we document the feature then.

Sources/EmbraceConfigInternal/EmbraceConfig.swift
```diff
@@ -105,6 +105,18 @@ extension EmbraceConfig /* EmbraceConfigurable delegation */ {
         configurable.isUiLoadInstrumentationEnabled
     }
 
+    public var uiInstrumentationBlockList: [String] {
```
naftaly: It might be better to give this a clearer name, something like `viewControllerClassNameBlocklist`.

Sources/EmbraceCore/Capture/UX/View/ViewControllerBlockList.swift
```diff
@@ -0,0 +1,65 @@
...
+    private func isHostingController(_ vc: UIViewController?) -> Bool{
+        guard let vc else {
+            return false
+        }
+
+        if vc is EmbraceIdentifiableHostingController {
+            return true
+        }
+
+        return isHostingController(vc.parent)
```
naftaly: nice!

naftaly (approved):
Very nice. I think other than the comment around the protection level of the list, this looks really solid. (I'd like the name to change but that's not that relevant) :) .

README.md
```diff
@@ -172,7 +172,7 @@ To ensure consistent formatting across the codebase, we use both [swift-format](
 The easiest way to run both formatters is via:
 
 ```sh
-make format
+make all
```
naftaly, line 175: thank you :)

naftaly (approved):

PR #312, opened 2025-07-23
Move CoreData work off the main queue

naftaly:
This PR tries to move as much CoreData work off the main queue as possible. This is "ok" since all operations are serialized to one queue so they will always run one after the other no matter what. This means that as long as our "live" cache is good, we can push the save off async and still be always up to date.

To the user, this should have a few benefits, mostly on the performance side since there's no UI contention for the CoreData queue.

The PR also removes the sync queue from the Embrace class and replaces it with a read/write lock. This makes the stacks easier to read as well as sets us up for a future refactor of the class in order to clean up any data races between things that are read constantly but only ever written to once on startup.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -90,8 +90,7 @@ let package = Package(
                 "EmbraceStorageInternal",
                 "EmbraceUploadInternal",
                 "EmbraceObjCUtilsInternal",
-                "EmbraceSemantics",
-                "EmbraceCrash"
+                "EmbraceSemantics"
             ],
             resources: [
                 .copy("PrivacyInfo.xcprivacy")
--- a/Package.swift
+++ b/Package.swift
@@ -90,7 +90,8 @@ let package = Package(
                 "EmbraceStorageInternal",
                 "EmbraceUploadInternal",
                 "EmbraceObjCUtilsInternal",
-                "EmbraceSemantics"
+                "EmbraceSemantics",
+                "EmbraceCrash"
             ],
             resources: [
                 .copy("PrivacyInfo.xcprivacy")
--- a/Package.swift
+++ b/Package.swift
@@ -41,9 +41,8 @@ let package = Package(
     ],
     dependencies: [
         .package(
-            //url: "https://github.com/kstenerud/KSCrash",
-            //exact: "2.2.0"
-            path: "../KSCrash"
+            url: "https://github.com/kstenerud/KSCrash",
+            exact: "2.2.0"
         ),
         .package(
             url: "https://github.com/open-telemetry/opentelemetry-swift",
--- a/Package.swift
+++ b/Package.swift
@@ -41,8 +41,9 @@ let package = Package(
     ],
     dependencies: [
         .package(
-            url: "https://github.com/kstenerud/KSCrash",
-            exact: "2.2.0"
+            //url: "https://github.com/kstenerud/KSCrash",
+            //exact: "2.2.0"
+            path: "../KSCrash"
         ),
         .package(
             url: "https://github.com/open-telemetry/opentelemetry-swift",
--- a/Sources/EmbraceCore/Capture/OneTimeServices/DeviceInfoCaptureService.swift
+++ b/Sources/EmbraceCore/Capture/OneTimeServices/DeviceInfoCaptureService.swift
@@ -14,6 +14,16 @@ class DeviceInfoCaptureService: ResourceCaptureService {
 
     override func onStart() {
 
+        let criticalResources: [String: String] = [
+            // os type
+            // Should always be "darwin" as can be seen in semantic convention docs:
+            // https://opentelemetry.io/docs/specs/semconv/resource/os/
+            ResourceAttributes.osType.rawValue: "darwin",
+
+            // os variant
+            DeviceResourceKey.osVariant.rawValue: EMBDevice.operatingSystemType
+        ]
+
         let resourcesMap: [String: String] = [
             // jailbroken
             DeviceResourceKey.isJailbroken.rawValue: EMBDevice.isJailbroken ? "true" : "false",
@@ -33,14 +43,6 @@ class DeviceInfoCaptureService: ResourceCaptureService {
             // os build
             DeviceResourceKey.osBuild.rawValue: EMBDevice.operatingSystemBuild,
 
-            // os variant
-            DeviceResourceKey.osVariant.rawValue: EMBDevice.operatingSystemType,
-
-            // os type
-            // Should always be "darwin" as can be seen in semantic convention docs:
-            // https://opentelemetry.io/docs/specs/semconv/resource/os/
-            ResourceAttributes.osType.rawValue: "darwin",
-
             // model
             ResourceAttributes.deviceModelIdentifier.rawValue: EMBDevice.model,
 
@@ -48,6 +50,7 @@ class DeviceInfoCaptureService: ResourceCaptureService {
             DeviceResourceKey.architecture.rawValue: EMBDevice.architecture
         ]
 
+        addCriticalResources(criticalResources)
         addRequiredResources(resourcesMap)
     }
 }
--- a/Sources/EmbraceCore/Embrace.swift
+++ b/Sources/EmbraceCore/Embrace.swift
@@ -101,7 +101,7 @@ import Foundation
         label: "com.embrace.synchronization",
         qos: .utility,
         autoreleaseFrequency: .workItem,
-        target: .global(qos: .utility)
+        target: .global(qos: .userInitiated)
     )
 
     static let notificationCenter: NotificationCenter = NotificationCenter()
@@ -125,14 +125,16 @@ import Foundation
         if ProcessInfo.processInfo.isSwiftUIPreview {
             throw EmbraceSetupError.initializationNotAllowed("Embrace cannot be initialized on SwiftUI Previews")
         }
-
+        
+        let setupTime = Date()
+        
         return try Embrace.synchronizationQueue.sync {
             if let client = client {
                 Embrace.logger.warning("Embrace was already initialized!")
                 return client
             }
 
-            EMBStartupTracker.shared().sdkSetupStartTime = Date()
+            EMBStartupTracker.shared().sdkSetupStartTime = setupTime
 
             try options.validate()
 
@@ -364,7 +366,7 @@ import Foundation
             throw EmbraceSetupError.invalidThread("Embrace must be stopped on the main thread")
         }
 
-        Embrace.synchronizationQueue.sync {
+        Embrace.synchronizationQueue.async { [self] in
             guard state != .stopped else {
                 Embrace.logger.warning("Embrace was already stopped!")
                 return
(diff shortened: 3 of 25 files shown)


Sources/EmbraceCore/Embrace.swift
```diff
@@ -295,36 +301,42 @@ import Foundation
...
+                storage.fetchLatestSession { [self] session in
+                    metricKit.lastSession = session
+                    metricKit.install()
+                }
 
-                self.processingQueue.async { [weak self] in
+                self.captureServices.start()
 
-                    self?.captureServices.start()
+                self.processingQueue.async { [weak self] in
```
ArielDemarco: Initially, we moved the `captureServices.start()` to work in the `async` block so we don't impact the startup performance (the slowest part of the SDK is basically some work from the swizzlers).
Is there any reason why we should keep it in the main queue?

naftaly: It's better on the main queue since it mostly deals with UI, but it's definitely something we should look at to see if it can help.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -84,10 +93,30 @@ public class CoreDataWrapper {
         return result
     }
 
+    /// Asynchronously performs the given block on the current context
+    /// behind a background task assertion.
+    /// And automatically save if requested.
+    public func performAsyncOperation(
+        _ name: String = #function, save _: Bool = false, _ block: @escaping (NSManagedObjectContext) -> Void
```
ArielDemarco: Seems the `save` parameter is unused. Shall you wrap the `saveIfNeeded`? or it's to do something else?

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Session.swift
```diff
@@ -175,6 +129,33 @@ extension EmbraceStorage {
...
+            request.fetchLimit = 1
+            request.sortDescriptors = [NSSortDescriptor(key: "startTime", ascending: false)]
+
+            if let sessionId = sessionId {
+                request.predicate = NSPredicate(format: "idRaw != %@", sessionId.toString)
+            }
+
+            if let session = coreData.fetch(withRequest: request).first {
+                let result = session.toImmutable()
+                DispatchQueue.global(qos: .default).async {
```
ArielDemarco, line 158: Why this dispatch is needed? (same for the one below)

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Session.swift
```diff
@@ -226,42 +207,79 @@ extension EmbraceStorage {
...
-                session.crashReportId = crashReportId
+                fetchedSession.crashReportId = crashReportId
             }
 
             coreData.save()
-            result = session.toImmutable()
         }
 
-        return result
+        return session.updated(
```
ArielDemarco, line 265: Smart move

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Session.swift
```diff
@@ -37,27 +38,10 @@ extension EmbraceStorage {
...
-                startTime: startTime,
-                endTime: endTime,
-                lastHeartbeatTime: lastHeartbeatTime,
-                crashReportId: crashReportId,
-                coldStart: coldStart,
-                cleanExit: cleanExit,
-                appTerminated: appTerminated
-            ) {
-                return session
-            }
```
ArielDemarco, line 58: It wasn't necessary to update the session (seems the name of the method doesn't imply that). Meaning... wans't this being used by a caller that could expect an update on an existing session?

Sources/EmbraceStorageInternal/Records/SessionRecord.swift
```diff
@@ -45,32 +45,26 @@ public class SessionRecord: NSManagedObject {
...
-            record.startTime = startTime
-            record.endTime = endTime
-            record.lastHeartbeatTime = lastHeartbeatTime ?? startTime
-            record.crashReportId = crashReportId
-            record.coldStart = coldStart
-            record.cleanExit = cleanExit
-            record.appTerminated = appTerminated
-
-            result = record.toImmutable()
+    ) -> SessionRecord? {
```
ArielDemarco: Considering this, i'd add some docs so nobody uses this result outside the `NSManagedObjectContext` (seems the only one using this method is using it properly)

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
naftaly: good catch, I'll fix that.

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Session.swift
naftaly: Just so we don't hijack the CoreData queue and keep it for core data things as much as possible.

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Session.swift
naftaly: I verified with @NachoEmbrace and he mentioned that this should have been removed a while ago.

Sources/EmbraceStorageInternal/Records/SessionRecord.swift
naftaly: I can also simply add the performOperation back, it might be safer since this can't be used without it anyway.

Sources/EmbraceCore/Embrace.swift
NachoEmbrace: Just wanted to note that the actual swizzling happens when `captureServices.install()` gets called, which is inside a sync dispatch called from the main queue.

`captureServices.start()` essentially just sets the state of the capture services to `active` so they can start capturing data.

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Session.swift
NachoEmbrace: We used to use the same method back in the GRDB days, and the method served the purpose of addition and updates on a session.
With the core data structure this doesn't really make much sense anymore, since there's already a dedicated method to update the session (heartbeat, clean exit, crash report id, etc).

Sources/EmbraceStorageInternal/Records/SessionRecord.swift
ArielDemarco: Im ok with the change; I'd not add the performOperation back

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Session.swift
```diff
@@ -37,28 +38,11 @@ extension EmbraceStorage {
...
+        let hbTime = lastHeartbeatTime ?? Date()
 
-            // create new
-            if let session = SessionRecord.create(
+        coreData.performAsyncOperation { [self] _ in
+            if SessionRecord.create(
```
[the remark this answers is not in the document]
naftaly: Adding a critical log here for now should be enough. There was also discussion about stopping the SDK at this point since a failure here is very problematic, but I think this might be better to do in a future PR. What do you think @NachoEmbrace @ArielDemarco

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Session.swift
```diff
@@ -226,42 +207,79 @@ extension EmbraceStorage {
...
+        }
 
-        coreData.fetchFirstAndPerform(withRequest: request) { session in
-            guard let session else {
+        coreData.performAsyncOperation { [self] _ in
+
+            let request = fetchSessionRequest(id: sessionId)
+            let fetchedSession = coreData.fetch(withRequest: request).first
+            guard let fetchedSession else {
                 return
```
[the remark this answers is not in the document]
naftaly: same as https://github.com/embrace-io/embrace-apple-sdk/pull/312#discussion_r2229608543

naftaly:
I've updated everything based on the conversations. Thanks for all the great feedback.

Sources/EmbraceStorageInternal/Records/EmbraceStorage+Metadata.swift
```diff
@@ -65,7 +65,7 @@ extension EmbraceStorage {
     /// Adds or updates all the given required resources
     public func addRequiredResources(_ map: [String: String], processId: ProcessIdentifier = .current) {
 
-        coreData.performOperation { context in
+        coreData.performAsyncOperation { [self] context in
```
NachoEmbrace: This will probably cause sessions with no resources when they end up (most likely by a crash) too quickly.
If we're missing this data when constructing the session payload, it basically makes it unusable by the back end, so we should make sure this change doesn't cause a spike in cases like that.

naftaly: That's a rough one. We can't synchronously write as it might cause a deadlock which would be worse than an early crash. Any suggestions on what we could do?

Package.swift
```diff
@@ -90,7 +90,8 @@ let package = Package(
                 "EmbraceStorageInternal",
                 "EmbraceUploadInternal",
                 "EmbraceObjCUtilsInternal",
-                "EmbraceSemantics"
+                "EmbraceSemantics",
+                "EmbraceCrash"
```
NachoEmbrace: `EmbraceCore` shouldn't depend on `EmbraceCrash`.
Any functionality you need should be defined on the protocols in `EmbraceCommonInternal`.

Sources/EmbraceCore/Capture/OneTimeServices/DeviceInfoCaptureService.swift
```diff
@@ -14,6 +14,16 @@ class DeviceInfoCaptureService: ResourceCaptureService {
 
     override func onStart() {
 
+        let criticalResources: [String: String] = [
+            // os type
+            // Should always be "darwin" as can be seen in semantic convention docs:
+            // https://opentelemetry.io/docs/specs/semconv/resource/os/
+            ResourceAttributes.osType.rawValue: "darwin",
```
ArielDemarco, line 21: Didn't remember this value is always `"darwin"` when we talked about the required resources... maybe we can just hardcode this value in the payload instead of saving it to the db? I dont think it'll make a huge impact tho

naftaly: I didn't change the value, just made it critical.

naftaly: we can definitely hard code on send if we want since its hard coded here.

ArielDemarco: Nevermind! We can leave it like this and do this and other changes when we fully move the resources to the other discussed solutions.

PR #314, opened 2025-07-24
[EASY] Otel 2.0.0 upgrade

naftaly:


ArielDemarco (commented):
Should we include a `~> 2.0.0` in the .`podspec`? What do you think? I was thinking about it mostly because of the breaking changes in OTel's public API

naftaly:
> Should we include a `~> 2.0.0` in the .`podspec`? What do you think? I was thinking about it mostly because of the breaking changes in OTel's public API

We don’t currently have a version specified in the podspec (that i could fine at least, maybe i missed it), so i thought best to leave it as is. We sort can almost let the user choose their version since the change isn’t “really” breaking and if they depend on the older version, why not let them.

fwiw, it was a hassle to update the BrandGame demo because it specified an otel version in the SPM app.

naftaly:
@ArielDemarco I change it to from 2.0.0 in SPM.

naftaly:
also updated the pod spec

naftaly:
Changed it all back to support of exactly 2.0.0

PR #318, opened 2025-07-29
Truncating span names instead of dropping the span entirely

NachoEmbrace:
The validation system we had in place was a bit too much and didn't even allow to modify the span data properly so I got rid of it...


Sources/EmbraceCore/Internal/Tracing/StorageSpanExporter.swift
```diff
@@ -33,39 +32,42 @@ class StorageSpanExporter: SpanExporter {
...
+                let data = try spanData.toJSON()
+
+                // spanData endTime is non-optional and will be set during `toSpanData()`
+                let endTime = spanData.hasEnded ? spanData.endTime : nil
+
+                // Prevent exporting our session spans on end.
+                // This process is handled by the `SessionController` to prevent
+                // race conditions when a session ends and its payload gets built.
+                if endTime != nil && spanData.embType == SpanType.session
+                {
```
ArielDemarco:
```suggestion
                if endTime != nil && spanData.embType == SpanType.session {
```

Sources/EmbraceCore/Internal/Tracing/StorageSpanExporter.swift
```diff
@@ -81,4 +83,27 @@ class StorageSpanExporter: SpanExporter {
...
+              type != .view,
+              type != .viewLoad else {
+            return name
+        }
+
+        var result = name
+
+        // trim white spaces
+        let trimSet: CharacterSet = .whitespacesAndNewlines.union(.controlCharacters)
+        result = name.trimmingCharacters(in: trimSet)
```
ArielDemarco, line 98: Looking at the previous code, seems that we're running this for all spans (`WhitespaceSpanNameValidator` is in the `validators` array before `LengthOfNameValidator`). So this logic should be moved _before_ the `guard`.

However... was that intended or a bug? may be your code is ok? shall we trim whitespaces from `networkRequest`, `.view` and `.viewLoad` spans?

ArielDemarco: _Note:_ if you decide on moving the trimming whitespaces _before_ the `guard`, i'd suggest to have a test so we remember that we made this decision

NachoEmbrace: I made this slight change intentionally.
The actual idea was to only do this validation for user created spans, but in our current system that's not easy to achieve.
The compromise was to only validate spans that are not the ones in that guard (same types as before).

ArielDemarco: ok; at the same time, I don't think there'll be whitesapces in those captured spans tbh.

PR #319, opened 2025-07-29
Ignoring AVFoundation related URLSessionTasks

NachoEmbrace:
This is to prevent crashes like `Application threw exception NSGenericException: AVAssetDownloadTask does not support originalRequest property`.

I pretty much copied the fix from the OTel sdk.

Note: Noticed some code duplication around the "finish" method in the handler, so I cleaned that up as well.

### DIFF
--- a/Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
+++ b/Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
@@ -24,6 +24,8 @@ protocol URLSessionTaskHandlerDataSource: AnyObject {
     var injectTracingHeader: Bool { get }
     var requestsDataSource: URLSessionRequestsDataSource? { get }
     var ignoredURLs: [String] { get }
+
+    var ignoredTaskTypes: [AnyClass] { get }
 }
 
 final class DefaultURLSessionTaskHandler: NSObject, URLSessionTaskHandler {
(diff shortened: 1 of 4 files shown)


Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
```diff
@@ -46,9 +48,22 @@ final class DefaultURLSessionTaskHandler: NSObject, URLSessionTaskHandler {
             payloadCaptureHandler ?? DefaultNetworkPayloadCaptureHandler(otel: dataSource?.otel)
     }
 
+    func shouldIgnoreTask(_ task: URLSessionTask) -> Bool {
+        if let dataSource {
+            return dataSource.ignoredTaskTypes.contains(where: { task.isKind(of: $0) })
+        }
```
ArielDemarco, line 54: Considering the `avTaskTypes` is a `static let` (meaning, is not a configurable thing). And this only matters to the handler (as it's the only one accessing `originalRequest` and `currentRequest`)... wouldn't be better to move it directly to this class?

ArielDemarco (approved):
Made a comment; is not a blocker.
Approved 👍

Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
NachoEmbrace: Did it this way to take advantage of the current injection pattern for tests.

ArielDemarco: 👍 ! Marked as solved

naftaly (approved):

PR #320, opened 2025-07-29
Re-architect Crash Support

naftaly:
This PR moves third party crash party reporters out of the hard dependency line (outside of EmbraceIO dev on the 3rd party KSCrashReporter).

`EmbraceCrashReporter` is now simply a private intermediary that takes in a `CrashReporter` either supplied by a customer chosen from the two we have `KSCRashReporter` or `CrashlyticsReporter`. Due the the removal of the dependency, we can now move `EmbraceCrashReporter` into the Core and it can become a full citizen as the rest.


NachoEmbrace (approved):

Sources/EmbraceCore/Crash/EmbraceCrashReporter.swift
```diff
@@ -0,0 +1,150 @@
...
+        super.init()
+    }
+
+    /// Used to determine if the last session ended cleanly or in a crash.
+    public func getLastRunState() -> LastRunState {
+        reporter.getLastRunState()
+    }
+
+    public func install(context: CrashReporterContext) {
+        sdkVersion = context.sdkVersion
```
ArielDemarco, line 92: Considering the crash reporter is not installed (and the setter of this with call `appendCrashInfo`), shouldn't be this called after `reporter.install` succeeded? We're expecting the setter of extra info to work in the reporters without being installed/initialized

Sources/EmbraceCore/Crash/EmbraceCrashReporter.swift
```diff
@@ -0,0 +1,150 @@
...
+
+    private let reporter: CrashReporter
+    private let logger: InternalLogger?
+    internal let queue: DispatchQueue = DispatchQueue(
+        label: "com.embrace.crashreporter", qos: .utility, autoreleaseFrequency: .workItem)
+    private let signalsBlockList: [CrashSignal]
+    
+    struct MutableData {
+        let internalKeys: [String] = [CrashReporterInfoKey.sdkVersion, CrashReporterInfoKey.sessionId]
+        var allowsInternalDataChange: Bool = false
```
ArielDemarco, line 23: Question: can you clarify why this property is needed? What is its purpose or what issue does it prevent?

Sources/EmbraceCore/Crash/EmbraceCrashReporter.swift
naftaly: I would expect so. We should be able to set key values on them without much ado about it being running or not. Otherwise, they can simply take in the value and set it when they have everything they need.

Sources/EmbraceCore/Crash/EmbraceCrashReporter.swift
naftaly: This is basically there so that anything that is set as a key in internal keys is set specifically by the EmbraceCrashReporter and not something external like a simple append. We don't want people overwriting our data for Sid and sdk. There's a test for this specifically and it helped catch this.

For example, we only allow setting the sessionId on the embrace crash reporter by calling `embraceCrashReporter.sessionId = ""`. We won't allow it being set by calling `embraceCrashReporter.appendCrashInfo("emb-sid", "")` unless that flag is set.

ArielDemarco: Awesome! Didn't even think about that usecase. Thanks for the explanation!

naftaly: Someone thought about it, kudos to them!!

ArielDemarco (approved):

NachoEmbrace (approved):

Sources/ThirdParty/EmbraceCrashlyticsSupport/CrashlyticsWrapper.swift
```diff
@@ -91,8 +110,9 @@ class CrashlyticsWrapper {
             let value = crashlyticsClass.perform(selector)
             instance = value?.takeUnretainedValue()
 
-            updateSessionId()
-            updateSDKVersion()
+            customValues.forEach { key, value in
```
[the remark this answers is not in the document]
naftaly: it's a computed property that is well protected.

PR #325, opened 2025-08-01
Fix handling config updates

ArielDemarco:
# Overview
Some configurations were not being updated because the notification wasn't posting the correct object.
Also, fixed the handling of the config to turn on/off capturing `UIHostingController`s.


NachoEmbrace:
If only the hosting controller flag is sent, then we should keep using the view controller class name list from the local options.

NachoEmbrace (approved):

PR #331, opened 2025-08-05
Platform tests cleanup

naftaly:
- cleanup tests for all platforms supported so they all run and pass.
- disabled some tests for features that are not support on certain platforms.
- skip tests on platforms with features that are currently unsupported but we might want to support in the future.
- format.

---

The only large difference from platform to platform is the lack of crash support on watchOS, everything else is very minor.


Sources/EmbraceCommonInternal/Swizzling/Swizzlable.swift
```diff
@@ -58,7 +58,8 @@ extension Swizzlable {
             to: ImplementationType.self)
         let newImplementationBlock: BlockImplementationType = block(originalTypifiedImplementation)
         let newImplementation = imp_implementationWithBlock(newImplementationBlock)
-        method_setImplementation(method, newImplementation)
+        let original = method_setImplementation(method, newImplementation)
+        assert(originalImplementation == original)
```
NachoEmbrace: Was this intended?

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService.swift
```diff
@@ -323,27 +323,27 @@ struct DataTaskWithURLRequestAndCompletionSwizzler: URLSessionSwizzler {
...
         try swizzleInstanceMethod { originalImplementation -> BlockImplementationType in
-            return { [weak handler = self.handler] urlSession, urlRequest, completion -> URLSessionDataTask in
+            return { [handler = self.handler] urlSession, urlRequest, completion -> URLSessionDataTask in
 
                 let request = urlRequest.addEmbraceHeaders()
 
                 guard let completion = completion else {
                     let task = originalImplementation(urlSession, Self.selector, request, completion)
-                    handler?.create(task: task)
+                    handler.create(task: task)
```
NachoEmbrace: Were these changes intended?
If so shouldn't we do it for al swizzlers?

Sources/EmbraceCommonInternal/Swizzling/Swizzlable.swift
naftaly: yes. It basically checks to see if the swizzle actually worked, only when debugging though. I can remove it since it's really not relevant to this PR.

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService.swift
naftaly: Intended, yes and no :) We try and use the handler a few times within the closure but it could go nil between those uses which might lead to weird results. We don't want to guard against it because the code needs to run regardless, so it makes it difficult to use consistently. There's also the probability of a cycle without the weak.

I'll return it to it's original state for now and revisit later.

PR #339, opened 2025-08-18
Feature request: Console log for breadcrumbs

NachoEmbrace:


Sources/EmbraceCore/Public/Embrace+OTel.swift
```diff
@@ -104,6 +104,13 @@ extension Embrace: EmbraceOpenTelemetry {
             return
         }
 
+        // console logs for breadcrumbs
+        for event in eventsToAdd where event.isBreadcrumb {
```
naftaly: I think this is fine, but there's a lot going on here even if nothing actually gets logged due to the level. Is there an opportunity to check the level before the loop and string work?

PR #341, opened 2025-08-18
Adding session id attribute to all exported spans

NachoEmbrace:


Sources/EmbraceSemantics/Span/SpanSemantics.swift
```diff
@@ -4,6 +4,7 @@
 
 public struct SpanSemantics {
     public static let keyEmbraceType = "emb.type"
+    public static let keySessionId = "session.id"
```
naftaly, line 7: Does this exist in Otel or are we inventing it?

Sources/EmbraceCore/Internal/Tracing/StorageSpanExporter.swift
```diff
@@ -32,28 +32,35 @@ class StorageSpanExporter: SpanExporter {
...
 
-                // sanitize name
-                let spanName = sanitizedName(spanData.name, type: spanData.embType)
-                guard !spanName.isEmpty else {
-                    logger?.warning("Can't export span with empty name!")
-                    result = .failure
-                    continue
+            do {
+                // add session id attribute
+                if let sessionId = sessionController?.currentSession?.idRaw {
```
naftaly, line 57: For my curiosity, how could we end up with span outside of a session?

Sources/EmbraceSemantics/Span/SpanSemantics.swift
NachoEmbrace: This is an Embrace specific attribute.

Sources/EmbraceCore/Internal/Tracing/StorageSpanExporter.swift
NachoEmbrace: While in the background with background sessions disabled.

Sources/EmbraceSemantics/Span/SpanSemantics.swift
naftaly: why don't we have a prefix for it?

PR #344, opened 2025-08-20
Exposed all CaptureServices in builder

ArielDemarco:


NachoEmbrace:
I think we should make a hotfix release for this (6.13.1).

ArielDemarco:
> I think we should make a hotfix release for this (6.13.1).

Im ok with it. I'll cherry pick this change and create a subsequent PR to a 6.13.1 release branch

PR #347, opened 2025-08-21
Fix backtrace bug

naftaly:
This PR ensures logs that use backtraces are taken on the same thread as the call, then moves all the processing to a background thread.

### DIFF
--- a/Sources/EmbraceCore/Internal/Logs/LogController.swift
+++ b/Sources/EmbraceCore/Internal/Logs/LogController.swift
@@ -25,7 +25,8 @@ protocol LogControllable: LogBatcherDelegate {
         attachmentId: String?,
         attachmentUrl: URL?,
         attributes: [String: String],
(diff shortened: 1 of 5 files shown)


ArielDemarco (approved):
LGTM; shall we include some sort of info around the `log` API usage in the `Embrace.swift` file? (e.g. recommending to avoid sending multiple `.warning`/`.error` in like, for example, a loop; or just mentioning that getting the stacktrace is a heavy duty and would cause some delays)

naftaly:
> LGTM; shall we include some sort of info around the `log` API usage in the `Embrace.swift` file? (e.g. recommending to avoid sending multiple `.warning`/`.error` in like, for example, a loop; or just mentioning that getting the stacktrace is a heavy duty and would cause some delays)

It's not that heavy anymore. Taking the trace should be very quick ( < 1ms depending on stack length ). Processing it was the heavy part, and that's been moved off main.

PR #354, opened 2025-08-26
[CI] Updates to build, test and formatter CI

naftaly:
- Build and tests for a lot more platforms and splits the runs out to actually know what fails and when.
- Clean up formatters, move them to run on macOS and ensure they work.


.github/workflows/run-tests.yml
```diff
@@ -50,39 +99,22 @@ jobs:
...
           path: .build/test/output.xcresult
-
-      - uses: kishikawakatsumi/xcresulttool@2b37f116695328d5e503ac8dadca0f1545a51da4 # v1.7.1
-        if: always()
-        with:
-          path: .build/test/output
-          title: "Test Results"
-          show-passed-tests: true
-          show-code-coverage: true
-          upload-bundles: "never" # Uploading the `xcresult` using the xcresultool action seems to fail;
```
ArielDemarco, line 85: 👏🏻

.github/workflows/run-tests.yml
```diff
@@ -5,32 +5,81 @@ on:
...
     timeout-minutes: 30
-    runs-on: macos-14
+    runs-on: macos-${{ matrix.macos-version }}
     strategy:
       fail-fast: false
       matrix:
-        platform_args: ["iOS tvOS macOS watchOS"]
+        platform: ["iOS"]
         xcode_version: ["15.4"]
+        macos-version: ["14"]
```
ArielDemarco, line 31: This means Sonoma? does it support Xcode 16.4 and 26?

.github/workflows/run-tests.yml
```diff
@@ -5,32 +5,81 @@ on:
...
     branches:
       - main
 
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
```
ArielDemarco, line 20: I like this; didn't know about it 👍

.github/workflows/swift-format.yml
```diff
@@ -16,18 +16,15 @@ permissions:
...
       timeout-minutes: 4
       with:
         persist-credentials: false
-
-    - name: Install make
-      run: |
-        apt-get update && apt-get install -y make
+    
+    - name: Install swift-format
+      run: brew install swift-format
```
ArielDemarco, line 27: Do you see any particular benefit on doing this? I try to do every single thing possible on linux as those runners are 10x cheaper than macOS runners ([source](https://docs.github.com/es/billing/concepts/product-billing/github-actions#free-use-of--data-variablesproductprodname_actions-))

ArielDemarco (commented):
Nice PR! Im curious: were you able to test how much time does it take the whole workflow now that you build more platforms in more versions but in a concurrent manner?

.github/workflows/run-tests.yml
naftaly: No, up to 16.2. But that's ok, we're in a matrix of runs so we also have macOS 15 + Xcode 16.4 and 26 in there.

naftaly: These numbers are simply the defaults.

.github/workflows/swift-format.yml
naftaly: @vitaliyf  mentioned that macOS runners are free in public repo's. We all work on macOS, so better off making sure we use the tools on the same platforms.

naftaly:
In the latest commit, I changed the version matrix to what I think should be enough. 

```
# iOS
          - platform: "iOS"
            xcode_version: "16.2"
            macos-version: "14"

          - platform: "iOS"
            xcode_version: "16.4"
            macos-version: "15"

          - platform: "iOS"
            xcode_version: "26.0"
            macos-version: "15"

          # tvOS
          - platform: "tvOS"
            xcode_version: "16.4"
            macos-version: "15"

          # macOS
          - platform: "macOS"
            xcode_version: "16.4"
            macos-version: "15"

          # watchOS
          - platform: "watchOS"
            xcode_version: "16.4"
            macos-version: "15"
            ```

.github/workflows/swift-format.yml
ArielDemarco: If that's the case, fair enough

ArielDemarco:
> > Nice PR! Im curious: were you able to test how much time does it take the whole workflow now that you build more platforms in more versions but in a concurrent manner?
> 
> It takes longer regardless. I think we could figure out what matrix of versions we want to run on and that'll help limit things a bit.
> 
> Like, do we still want to run on Xcode 15.4? I think we should move to 16.2 or 16.4 and remove that whole set of runs, what do you think.

I like that; saw that you changed it and removed `Xcode 15.4`. The job is flexible enough, so we can add it again if we need it too (IMO I don't think we will)

naftaly:
In a future PR, maybe we can add actual SDK's so we' really test against iOS 13 since we support it :)

PR #360, opened 2025-08-29
Fix crashes related to not being able to load the persistent store

ArielDemarco:
# Overview
This PR prevents crashes that could be caused whenever we're unable to load the persistent store (e.g. `container.loadPersistentStores(completionHandler:)` ). 

## Context
Currently, if an error occurs while loading the persistent store, we simply create a critical log and move on. However, later on, when an operation such as `save()` is executed, a crash will occur.

To address this, this PR introduces a fail-fast approach during SDK initialization. If the persistent store cannot be loaded, we now `throw` an error in `Embrace.setup`. The possible errors are:
- `EmbraceSetupError.failedStorageCreation`
- `EmbraceSetupError.failedUploadModuleCreation`

### DIFF
--- a/Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
+++ b/Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
@@ -18,10 +18,14 @@ public class CoreDataWrapper {
 
     private let isTesting: Bool
 
-    public init(options: CoreDataWrapper.Options, logger: InternalLogger) throws {
+    public init(
+        options: CoreDataWrapper.Options,
+        logger: InternalLogger,
+        isTesting: Bool = ProcessInfo.processInfo.isTesting
+    ) throws {
         self.options = options
         self.logger = logger
-        isTesting = ProcessInfo.processInfo.isTesting
+        self.isTesting = isTesting
 
         // create model
         let model = NSManagedObjectModel()
@@ -32,7 +36,7 @@ public class CoreDataWrapper {
         container = NSPersistentContainer(name: name, managedObjectModel: model)
 
         // force db on memory during tests
-        if isTesting {
+        if self.isTesting {
             let description = NSPersistentStoreDescription()
             description.type = NSInMemoryStoreType
             container.persistentStoreDescriptions = [description]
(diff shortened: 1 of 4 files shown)


Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -54,17 +58,24 @@ public class CoreDataWrapper {
                 description.type = NSSQLiteStoreType
                 description.url = options.storageMechanism.fileURL
                 description.setValue(journalMode.rawValue as NSString, forPragmaNamed: "journal_mode")
-
+                // This is the default value; however, we enforce it here so that the `CoreDataWrapper`
+                // is created synchronously in `Embrace.init`, allowing us to throw as needed and fail early.
+                description.shouldAddStoreAsynchronously = false
```
naftaly, line 63: Do the same for testing so we have a similar setup.

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -54,17 +58,24 @@ public class CoreDataWrapper {
...
 
+        var loadPersistentStoreError: Error?
         container.loadPersistentStores { _, error in
-            if let error {
-                logger.critical("Error initializing CoreData \"\(name)\": \(error.localizedDescription)")
-            }
+            loadPersistentStoreError = error
         }
 
+        if let loadPersistentStoreError {
```
naftaly, line 81: A comment here explaining why we're doing this will be useful for future us when we have no clue why anymore. :)

Sources/EmbraceCoreDataInternal/CoreDataWrapper.swift
```diff
@@ -54,17 +58,32 @@ public class CoreDataWrapper {
...
                 description.setValue(journalMode.rawValue as NSString, forPragmaNamed: "journal_mode")
-
+                // This is the default value; however, we enforce it here so that the `CoreDataWrapper`
+                // is created synchronously in `Embrace.init`, allowing us to throw as needed and fail early.
+                description.shouldAddStoreAsynchronously = false
                 container.persistentStoreDescriptions = [description]
             }
         }
 
+        // Even though this happens inside a block, by default it runs synchronously on the same thread
```
naftaly, line 68: ❤️

naftaly (approved):

PR #361, opened 2025-08-29
Update the way we run tests in CI

naftaly:
- use mxcl/xcodebuild, which uploads artifacts on failure.
- update codecov/codecov-action

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -62,7 +62,6 @@ jobs:
     permissions:
       checks: write
     steps:
-        # See https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md
       - name: Select Xcode
         run: |
           sudo xcode-select -s /Applications/Xcode_${{ matrix.xcode_version }}.app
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -74,46 +73,17 @@ jobs:
...
-      - name: List result files
-        run: ls -R .build/test | grep xcresult
-
-      - name: Convert results for Codecov
-        run: |
-          brew install a7ex/homebrew-formulae/xcresultparser
-          xcresultparser --output-format cobertura .build/test/output.xcresult > .build/test/output.xcresult.xml
-
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24 # v5.4.3
```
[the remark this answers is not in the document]
naftaly: Nope, they are meant to work together.

.github/workflows/run-tests.yml
```diff
@@ -74,46 +73,17 @@ jobs:
...
-        with:
-          name: output_xcresult_${{ matrix.platform }}_xcode_${{ matrix.xcode_version }}_macos_${{ matrix.macos-version }}
-          path: .build/test/output.xcresult
+          platform: ${{ matrix.platform }}
+          action: test
+          code-coverage: true
+          configuration: Debug
+          scheme: EmbraceIO-Package
+          upload-logs: on-failure
+          trust-plugins: false
```
ArielDemarco, line 84: Is there any specific flag to do the retry mechanism we have in place in the `bin/test`? Hopefully is not needed as we got rid of a lot of flakyness in tests, but sometimes is helpful.

ArielDemarco (approved):

.github/workflows/run-tests.yml
naftaly: There is no flag for that, we could suggest it however. tbh, I prefer to get failures than a test that sometimes works.

PR #362, opened 2025-08-29
reduce unit test duration

naftaly:


Tests/EmbraceCoreTests/Session/SessionControllerTests.swift
```diff
@@ -215,6 +217,16 @@ final class SessionControllerTests: XCTestCase {
...
+    func wait(_ until: () -> Bool) {
+        // Here, we end up having to wait for the DefaultSession uploader which
+        // isn't set up to give us a completion, so we'll fake it 'till we make it.
+        let maxTimes = 200  // 2 seconds basically
+        for _ in 0..<maxTimes {
+            wait(delay: 0.01)
+            if until() { return }
+        }
+    }
+
```
ArielDemarco: There's a similar function in `TestSupport`:
```swift
public func wait(timeout: TimeInterval = .defaultTimeout, interval: TimeInterval = 0.1, until block: @escaping () throws -> Bool)
```

Tests/ThirdParty/EmbraceCrashlyticsSupportTests/CrashlyticsReporterTests.swift
```diff
@@ -66,6 +66,6 @@ class CrashlyticsReporterTests: XCTestCase {
         NotificationCenter.default.post(name: Notification.Name("networkRequestCaptured"), object: task)
 
         // then a report is not generated
-        wait(for: [expectation], timeout: .veryLongTimeout)
+        wait(for: [expectation], timeout: 0.05)
```
ArielDemarco:
```suggestion
        wait(for: [expectation], timeout: .shortTimeout)
```

Tests/EmbraceCoreTests/Session/SessionControllerTests.swift
naftaly: the largest difference between the two is the closure which is escaping. I've changed it to use the one we already have.

Tests/ThirdParty/EmbraceCrashlyticsSupportTests/CrashlyticsReporterTests.swift
naftaly: I added `veryShortTimeout = 0.05`

PR #363, opened 2025-08-29
[EASY] Concurrency and lock improvements

naftaly:
I added comments inline for each change.

### DIFF
--- a/Sources/EmbraceConfigInternal/EmbraceConfig.swift
+++ b/Sources/EmbraceConfigInternal/EmbraceConfig.swift
@@ -77,9 +77,7 @@ public class EmbraceConfig {
                 self?.lastUpdateTime = Date().timeIntervalSince1970
 
                 if didChange {
-                    DispatchQueue.global(qos: .default).async {
-                        self?.notificationCenter.post(name: .embraceConfigUpdated, object: self?.configurable)
-                    }
+                    self?.notificationCenter.post(name: .embraceConfigUpdated, object: self?.configurable)
                 }
             }
         }


Sources/EmbraceCommonInternal/Protocols/DispatchableQueue.swift
```diff
@@ -35,6 +35,8 @@ extension DispatchableQueue where Self == DefaultDispatchableQueue {
         qos: DispatchQoS = .unspecified,
         attributes: DispatchQueue.Attributes = []
     ) -> DispatchableQueue {
-        DefaultDispatchableQueue(queue: .init(label: label, qos: qos, attributes: attributes))
+        DefaultDispatchableQueue(
+            queue: DispatchQueue(label: label, qos: qos, attributes: attributes, autoreleaseFrequency: .workItem)
```
naftaly, line 39: Default to autorelease pools around each block. It's the recommended way to go about it.

Sources/EmbraceConfigInternal/EmbraceConfig.swift
```diff
@@ -30,7 +30,7 @@ public class EmbraceConfig {
         options: Options,
         notificationCenter: NotificationCenter,
         logger: InternalLogger,
-        queue: DispatchableQueue = .with(label: "com.embrace.config", attributes: .concurrent)
+        queue: DispatchableQueue = .with(label: "com.embrace.config")
```
naftaly, line 33: Made the queue serial. This queue is simply used to update changes to remote configs and we don't want multiples at the same time.

Sources/EmbraceConfigInternal/EmbraceConfig.swift
```diff
@@ -77,7 +77,9 @@ public class EmbraceConfig {
                 self?.lastUpdateTime = Date().timeIntervalSince1970
 
                 if didChange {
-                    self?.notificationCenter.post(name: .embraceConfigUpdated, object: self?.configurable)
+                    DispatchQueue.global(qos: .default).async {
```
naftaly: Don't give external users access to our internal queue.

Sources/EmbraceCore/Internal/Embrace+Setup.swift
```diff
@@ -91,7 +91,7 @@ extension Embrace {
 
         do {
             let options = EmbraceUpload.Options(endpoints: uploadEndpoints, cache: cache, metadata: metadata)
-            let queue = DispatchQueue(label: "com.embrace.upload", qos: .background, attributes: .concurrent)
+            let queue = DispatchQueue(label: "com.embrace.upload", qos: .utility)
```
naftaly, line 94: Made the queue .utility. The `.background` queue might not run in certain circumstances like the app being in the background (yeah, I know it's weird) or other low power/memory situations.

Sources/EmbraceUploadInternal/Operations/AsyncOperation.swift
```diff
@@ -4,8 +4,12 @@
 
 import Foundation
 
+#if !EMBRACE_COCOAPOD_BUILDING_SDK
+    import EmbraceCommonInternal
+#endif
+
 class AsyncOperation: Operation, @unchecked Sendable {
-    private let lockQueue = DispatchQueue(label: "com.embrace.asyncoperation", attributes: .concurrent)
+    private let lock = ReadWriteLock()
```
naftaly, line 12: Moved away from a queue and instead use a pthread r/w lock.

Sources/EmbraceConfigInternal/EmbraceConfig.swift
ArielDemarco: In which way they have access? The `notificationCenter` used here is the private one generated inside Embrace (`Embrace.notificationCenter`); only internal objects that use that reference are going to receive notifications

naftaly: ha, interesting, I've almost never seen that done before. cool trick. What if someday it changes and some other Notification Center is passed in?

naftaly: I changed it back to the way it was.

ArielDemarco (approved):

PR #367, opened 2025-09-02
Pull Request Performance Regression Testing CI

naftaly:
This PR creates a very simple Github action that takes in a set of tests, and compares them between this PR and main and spits out a comment like [this one](https://github.com/embrace-io/embrace-apple-sdk/pull/367#issuecomment-3249806196).

To add tests, simply go to `run-perf.yml` and add new tests to the `TESTS` env var, one per line. The tests are the `xcodebuild` format `<module>/<class>/<test>` and you can leave out any of those pieces to get all the tests in a group.

```
env:
  TESTS: |
    EmbraceIOTests
    EmbraceIOTests/PerformanceTests
    EmbraceIOTests/PerformanceTests/test_embraceStartBasic
```

Maybe in the future we'll drop this list in a file in the repo as to not update the action everytime.

### DIFF
--- a/.github/workflows/run-perf.yml
+++ b/.github/workflows/run-perf.yml
@@ -145,9 +145,6 @@ jobs:
                 owner: context.repo.owner,
                 repo: context.repo.repo,
                 issue_number: context.issue.number,
-               body,
+                body,
               });
             }
-
-                
-
--- a/.github/workflows/run-perf.yml
+++ b/.github/workflows/run-perf.yml
@@ -12,10 +12,6 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
-defaults:
-  run:
-    shell: bash -euo pipefail
-
 env:
   TESTS: |
     EmbraceIOTests/PerformanceTests
--- a/.github/workflows/run-perf.yml
+++ b/.github/workflows/run-perf.yml
@@ -1,0 +1,149 @@
+name: Perf Check
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  pull-requests: write
+  contents: read
+
+env:
+  TESTS: |
+    EmbraceIOTests/PerformanceTests
+
+jobs:
+  perf:
+    runs-on: macos-latest
+    
+    steps:
+
+      # checkout this PR branch
+      - name: Checkout PR
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      
+      - name: Copy perf test binary
+        run: 
+          cp bin/perf /tmp/perf || true
+          chmod +x tmp/perf || true
+
+      - name: Run PR perf tests
+        id: perfpr
+        run: |
+          /tmp/perf >perf.txt
+          echo "out<<EOF" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


.github/workflows/run-perf.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      # checkout this PR branch
+      - name: Checkout PR
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      
+      - name: Copy perf test binary
+        run: 
+          cp bin/perf /tmp/perf || true
```
fnewberg: 1. we should have this explicitly fail if the copy fails rather than hide it with `|| true`
2. why do we have to copy this to /tmp?

.github/workflows/run-perf.yml
```diff
@@ -0,0 +1,149 @@
...
+          chmod +x tmp/perf || true
```
fnewberg:
```suggestion
          chmod +x /tmp/perf || true
```

.github/workflows/run-perf.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: Run PR perf tests
+        id: perfpr
+        run: |
+          /tmp/perf >perf.txt
```
fnewberg, line 39: does it matter if we write to the same file for the PR and main, or is that irrelevant since we're appending the data to $GITHUB_OUTPUT?

.github/workflows/run-perf.yml
ArielDemarco:
```suggestion
          chmod +x /tmp/perf || true
```

.github/workflows/run-perf.yml
```diff
@@ -0,0 +1,150 @@
...
+    steps:
+
+      # checkout this PR branch
+      - name: Checkout PR
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      
+      - name: Copy perf test binary
+        run: |
```
fnewberg, line 32: to avoid having to set `set -eou pipefail` in every bash section in the workflow, i think it makes sense to add

```
defaults:
  run:
    shell: bash -euo pipefail
    ```

.github/workflows/run-perf.yml
naftaly: 1. Can do, didn't think mattered too much as the rest will fail quickly afterwards.
2. Two reasons. First is so both runs use the same binary. Second, in case the other branch doesn't have it yet as is the case now.

.github/workflows/run-perf.yml
naftaly: `$GITHUB_OUTPUT` is setup so that anything that pushed to it goes to a slot based on the `id`, so it doesn't matter.

.github/workflows/run-perf.yml
naftaly: I just tried that and everything started failing on me. Typical ChatGTP thinking things will work until everything is broken :)

fnewberg: oh interesting. and i was all excited thinking that you could do that since it makes things boilerplate-y 😂

naftaly: There's probably a way if we get AI to focus on it.

PR #368, opened 2025-09-04
[EMBR-7724][UI_TESTS_APP] Add tests to measure performance impact of exporting Spans and Logs on an app.

ferdraghi:
Adds a screen that stress tests the sdk running hashing operations while exporting spans or logs and measures how long it takes vs running the same operations without exporting anything.


Examples/EmbraceIOTestApp/EmbraceIOTestApp/Tests/Performance/Tests/PerformanceLogTest.swift
```diff
@@ -0,0 +1,93 @@
...
+    var runImmediatelyIfSpansFound: Bool { true }
+    var numberOfLoops: Int = 0
+    var calculationsPerLoop: Int = 0
+    var maxNumberOfLogs: Int = 0
+
+    func runTestPreparations() {
+        Embrace.client?.buildSpan(name: "LoggingTestStart").startSpan().end()
+    }
+
+    func test(spans: [SpanData]) -> TestReport {
```
naftaly, line 23: Have you given any thought to making this async and using Task groups instead. It would make the code easier to follow + async is fun :)

naftaly (approved):

PR #371, opened 2025-09-05
[EASY] Fixed autotermination attribute not being added to Span

ferdraghi:
Fixed a silly mistake that would prevent Auto Termination code from being added to the span attributes.

Also added a Unit Test for it although the whole section where it belongs is currently commented out due to it being unstable on CI. Tested it locally and it passes though.

### DIFF
--- a/Sources/EmbraceCore/Public/Embrace+OTel.swift
+++ b/Sources/EmbraceCore/Public/Embrace+OTel.swift
@@ -33,8 +33,8 @@ extension Embrace: EmbraceOpenTelemetry {
     /// - Parameters:
     ///    - name: The name of the span.
     ///    - type: The type of the span. Will be set as the `emb.type` attribute.
-    ///    - autoTerminationCode: `SpanErrorCode` to be used to automatically close this span if the current session ends while the span is open.
     ///    - attributes: A dictionary of attributes to set on the span.
+    ///    - autoTerminationCode: `SpanErrorCode` to be used to automatically close this span if the current session ends while the span is open.
     /// - Returns: An OpenTelemetry `SpanBuilder`.
     public func buildSpan(
         name: String,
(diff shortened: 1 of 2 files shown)


Sources/EmbraceCore/Public/Embrace+OTel.swift
```diff
@@ -42,12 +42,14 @@ extension Embrace: EmbraceOpenTelemetry {
         attributes: [String: String] = [:],
         autoTerminationCode: SpanErrorCode? = nil
     ) -> SpanBuilder {
+        var extraAttributes = [String: String]()
```
naftaly: You could do something possibly a bit more performant by taking advantage of COW in case there are no changes to the attributes:

```
// no copy here since we haven't mutated yet
var finalAttributes = attributes
if let autoTerminationCode = autoTerminationCode {
    // copy happens here (COW) since we change the atts
    finalAttributes[SpanSemantics.keyAutoTerminationCode] = autoTerminationCode.rawValue
}
return otel.buildSpan(name: name, type: type, attributes: finalAttributes)
```

naftaly (approved):
Good, but check my comments with possible improvements.

Sources/EmbraceCore/Public/Embrace+OTel.swift
ferdraghi: I'll double down and only do the mutable copy if there is a termination code, otherwhise just return as is.

Sources/EmbraceCore/Public/Embrace+OTel.swift
```diff
@@ -33,20 +33,22 @@ extension Embrace: EmbraceOpenTelemetry {
...
         autoTerminationCode: SpanErrorCode? = nil
     ) -> SpanBuilder {
-        if let autoTerminationCode = autoTerminationCode {
-            var attributes = attributes
-            attributes[SpanSemantics.keyAutoTerminationCode] = autoTerminationCode.rawValue
+        guard let autoTerminationCode = autoTerminationCode else {
+            return otel.buildSpan(name: name, type: type, attributes: attributes)
         }
 
+        var attributes = attributes
```
naftaly, line 49: does this actually work? Can you use the same name like that?

ferdraghi: yup

PR #372, opened 2025-09-06
Stat Sig Perf Regression CI

naftaly:


.github/workflows/run-perf.yml
```diff
@@ -13,6 +13,9 @@ concurrency:
   cancel-in-progress: true
 
 env:
+  ITERATIONS: 20
+  SHOULD_CLEAN: 1
+  THRESHOLD: 5.0
```
fnewberg: nitpick: i'd call this `THRESHOLD_PCT`

.github/workflows/run-perf.yml
```diff
@@ -28,10 +31,19 @@ jobs:
         with:
           persist-credentials: false
       
-      - name: Copy perf test binary
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
```
fnewberg: any reason to not use 3.13? 3.11 is a bit on the old side

ArielDemarco (approved):
LGTM

.github/workflows/run-perf.yml
naftaly: no reason I know of. Let me see if everything still works.

naftaly: they've got 3.13.7 in the macOS 15 image, so i'll try that.

PR #374, opened 2025-09-08
[EMBR-9027] Improved Memory Instrumentation

naftaly:
Move to #461 in order to use the new branch naming system.

### DIFF
--- a/Sources/EmbraceIO/Services/AppMemoryCaptureService.swift
+++ b/Sources/EmbraceIO/Services/AppMemoryCaptureService.swift
@@ -22,10 +22,6 @@ import OpenTelemetryApi
 
 public class AppMemoryCaptureService: CaptureService {
 
-    // This is shared for now until we have access to the AppMemoryTracker
-    // and can add an observer. See `init` for why.
-    public static let shared = AppMemoryCaptureService()
-
     struct ServiceMutableData {
         var enabled: Bool = true
         var observer: Any? = nil
--- a/Sources/EmbraceIO/Capture/CaptureService+Helpers.swift
+++ b/Sources/EmbraceIO/Capture/CaptureService+Helpers.swift
@@ -67,6 +67,6 @@ import Foundation
     }
 
     static func appMemory() -> AppMemoryCaptureService {
-        return .shared
+        return AppMemoryCaptureService()
     }
 }


Sources/EmbraceIO/Services/AppMemoryCaptureService.swift
```diff
@@ -0,0 +1,155 @@
...
+                stackTraceBehavior: .notIncluded
+            )
+        }
+
+        // update the footprint at the process level
+        try? metadata?.setProcessProperty(key: "emb.memory_level", value: memory.level.asString())
+        try? metadata?.setProcessProperty(key: "emb.memory_pressure", value: memory.pressure.asString())
+        try? metadata?.setProcessProperty(key: "emb.memory_footprint", value: "\(memory.footprint)")
+        try? metadata?.setProcessProperty(key: "emb.memory_limit", value: "\(memory.limit)")
+        try? metadata?.setProcessProperty(key: "emb.memory_remaining", value: "\(memory.remaining)")
```
NachoEmbrace: Probably move these hardcoded keys to LogSemantics so we don't duplicate them and and also keep it consistent with the rest of the SDK.

naftaly: done.

naftaly:
Remove the Otel log and left it simply at the RecordedSpanEvent level to make it just like the low memory capture service.

Sources/EmbraceIO/Services/AppMemoryCaptureService.swift
```diff
@@ -0,0 +1,138 @@
...
+
+    private override init() {
+        // In KSCrash 2.3, the shared AppMemoryTracker is not accessible to the outside,
+        // so for now i'm simply swizzling it. In 3.0, I've fixed it so it's accessible
+        // and we'll just use that.
+        AppMemoryTracker.embraceSwizzle()
+    }
+
+    fileprivate func memoryChanged(memory: AppMemory, with changes: AppMemoryTrackerChangeType) {
+        guard state == .active else {
```
ArielDemarco: I'd include the possibility toggle this feature remotely

Sources/EmbraceIO/Services/AppMemoryCaptureService.swift
```diff
@@ -0,0 +1,105 @@
...
+    import KSCrashRecording
+#elseif canImport(KSCrash)
+    import KSCrash
+#endif
+
+public class AppMemoryCaptureService: CaptureService {
+
+    // This is shared for now until we have access to the AppMemoryTracker
+    // and can add an observer. See `init` for why.
+    public static let shared = AppMemoryCaptureService()
```
NachoEmbrace: I don't understand why this has to be a static shared instance, can you explain?

naftaly: Ah yes, no need anymore since we've updated to KSCrash 2.4. Pre-2.4 it involved swizzling to get access to the instance used in KSCrash, now it's exposed. I'll fix it.

PR #379, opened 2025-09-10
Improved Backtrace/Stacktrace support

naftaly:
- New Backtracking and Symbolication protocols. Users can pass in implementations in order to use their own versions. By default, in EmbraceIO, we use a KSCrash based version.
- New Backtrace classes (EmbraceBacktrace) allow to take a performant backtrace of any thread at any point in time.
- Added a `StackTraceBehavior.main` which forces the stack trace to be taken off the main queue regardless of where the code executes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ install-hooks:
 	done
 	@echo "Done!"
 
+# remove hooks
 uninstall-hooks:
 	@echo "Uninstalling Git hooks..."
 	@for hook in $(HOOKS_DIR)/*; do \


Sources/ThirdParty/EmbraceKSCrashBacktraceSupport/EmbraceKSCrashBacktraceSupport.swift
```diff
@@ -0,0 +1,110 @@
...
+    @_exported import KSCrashDemangleFilter
+#endif
+
+public class KSCrashBacktracing: Backtracer, Symbolicator {
+
+    public init() {}
+
+    public func backtrace(of thread: pthread_t) -> [UInt] {
+
+        // In KSCrash there, a bug that causes a backtrace on the pthread_self
```
ferdraghi:
```suggestion
        // In KSCrash there's a bug that causes a backtrace on the pthread_self
```

Small typo

Sources/EmbraceCommonInternal/CrashReporter/CrashReporter.swift
```diff
@@ -4,30 +4,192 @@
 
 import Foundation
 
+// MARK: - Run State
```
naftaly, line 7: Simply added a load of docs.

Sources/EmbraceCommonInternal/Models/StackTraceBehavior.swift
```diff
@@ -10,8 +10,13 @@ public enum StackTraceBehavior {
     case notIncluded
 
     /// The default behavior for automatically capturing stack traces.
+    /// which takes a stack of the current thread.
     case `default`
 
+    /// Stack traces are taken for the main thread.
+    /// Only available if `EmbraceBacktrace.isAvailable == true`.
+    case main
```
naftaly, line 21: Special case where .log can take a stack trace of the main thread instead of the thread it's called on.

Sources/EmbraceCore/Internal/Logs/LogController.swift
```diff
@@ -105,19 +106,33 @@ class LogController: LogControllable {
...
         let addStacktraceBlock: ((_ builder: EmbraceLogAttributesBuilder) -> Void)?
         switch stackTraceBehavior {
         case .default where severity == .warn || severity == .error:
-            let stackTrace = Thread.callStackSymbols
-            addStacktraceBlock = { $0.addStackTrace(stackTrace) }
+            if EmbraceBacktrace.isAvailable {
+                let backtrace = EmbraceBacktrace.backtrace(of: pthread_self(), suspendingThreads: false)
+                addStacktraceBlock = { $0.addBacktrace(backtrace) }
+            } else {
+                let stacktrace = Thread.callStackSymbols
```
naftaly, line 125: Old style backtrace fallback.

Sources/EmbraceCore/Options/Embrace+Options.swift
```diff
@@ -43,6 +45,10 @@ extension Embrace {
         ///   - logLevel: The `LogLevel` to use for console logs.
         ///   - export: `OpenTelemetryExport` object to export telemetry using OpenTelemetry protocols
         ///   - processors: `OpenTelemetryProcessor` objects to do extra processing
+        ///   - backtracer: Optional `Backtracer` to capture stack traces. Defaults to the
```
naftaly, line 48: New comments around optional values.

Sources/ThirdParty/EmbraceKSCrashBacktraceSupport/EmbraceKSCrashBacktraceSupport.swift
```diff
@@ -0,0 +1,110 @@
...
+#endif
+
+#if canImport(KSCrash)
+    import KSCrash
+#else
+    import KSCrashRecording
+    @_exported import KSCrashDemangleFilter
+#endif
+
+public class KSCrashBacktracing: Backtracer, Symbolicator {
```
naftaly, line 14: Backtracing and Symbolication plugin using KSCrash

Sources/EmbraceCommonInternal/Models/StackTraceBehavior.swift
ferdraghi: The new "main" type should also be declared on the equatable protocol below. I can foresee this getting out of hand very quickly though... We should consider simply adding a type to the enum, maybe, or something like that.

naftaly: Ah, that's what `@unknown default` is for :) otherwise with a `default:` case it's hard to do anything about it.

Examples/BrandGame/BrandGame/View/Menu/StressTests/Logging/LoggingView.swift
```diff
@@ -134,6 +136,7 @@ extension LoggingView {
     enum Behavior: Int {
         case notIncluded
         case `default`
+        case main
         case custom
```
ArielDemarco, line 140: Just in case sombody is converting from `Int` to `Behavior`, would you be able to move the new one to be the last one?
```suggestion
        case custom
        case main
```

ArielDemarco: Nevermind! Saw the file and this is from Brandgame, so no problem at all :D

naftaly: I'll reorder them, but an enum without a raw type can't be used as an int.

PR #383, opened 2025-09-13
[EMBR-9024] Homegrown 0x8badf00d with stacks (watchdog events)

naftaly:
This PR implements watchdog (0x8badf00d) crash detection for main thread hangs by writing synthetic crash reports when hangs occur.

### What is it exactly

This is a type of termination/crash we do not currently support. It happens a lot, but is not currently counted in our crash rate metrics. It appears to users as a crash on launch, or a frozen app that just disappears on screen. This gives us the same information as the 0x8badf00d crash report from MetricKit, but **un**sampled and fully in our control. Since it simply uses the exact same mechanism as our current crash system, it's basically plug and play.

### How it works

When the hang watchdog detects a main thread hang:
1. Hang starts → writes a synthetic crash report to disk with reason "0x8badf00d, main thread blocked for X seconds"
2. Hang ends → deletes the synthetic crash report

If iOS terminates the app during a hang (watchdog timeout), the synthetic report remains on disk and is automatically picked up on next launch as a real crash report.

### Changes

- New notification system: Added WatchdogEvent model and .hangEventStarted/.hangEventUpdated/.hangEventEnded notifications
- Moved NanosecondClock to EmbraceCommonInternal with public subtraction operator
- KSCrashReporter observes hang notifications and writes/deletes reports accordingly
- Marks main thread (thread 0) as crashed thread for watchdog reports
- Uses KSCrash's didWriteReportCallback to track report IDs for deletion
- HangCaptureService: Posts notifications at each hang lifecycle event
- Added takeValue() helper to EmbraceMutex for nilable types
  
 ### What's needed to land this?
 
 - [ ] A few adjustments in the dashboard (crash rate, split them from crashes maybe??)
 - [x] A gate on the client side. (depends on EMBR-9292)
---
  
[Example Session](https://dash.embrace.io/app/wnrhv/grouped_sessions/last_3_hour/025DF0F1191D44A1BEB2D3E088A283C4/AE232244903D413BB7A42D004339BD2D@/AE232244903D413BB7A42D004339BD2D?first=6FE78221CF144CC4898DAF2587C4AD2C&last=69177E13663748448593C956D2FE0389&source=session)

Screenshot 2025-10-01 at 3 31 36 PM

### DIFF
--- a/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
+++ b/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
@@ -91,7 +91,8 @@ extension RemoteConfig: EmbraceConfigurable {
     public var hangLimits: HangLimits {
         HangLimits(
             hangPerSession: payload.hangLimitsHangPerSession,
-            samplesPerHang: payload.hangLimitsSamplesPerHang
+            samplesPerHang: payload.hangLimitsSamplesPerHang,
+            reportsWatchdogEvents: payload.hangLimitsReportsWatchdogEvents
         )
     }
 
--- a/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig/RemoteConfigPayload.swift
+++ b/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig/RemoteConfigPayload.swift
@@ -41,6 +41,7 @@ public struct RemoteConfigPayload: Decodable, Equatable {
 
     var hangLimitsHangPerSession: UInt
     var hangLimitsSamplesPerHang: UInt
+    var hangLimitsReportsWatchdogEvents: Bool
 
     var networkPayloadCaptureRules: [NetworkPayloadCaptureRule]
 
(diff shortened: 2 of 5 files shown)


Sources/EmbraceCommonInternal/CrashReporter/HangEvent.swift
```diff
@@ -0,0 +1,68 @@
...
+    /// The notification’s `object` is a `WatchdogEvent` containing the
+    /// current timestamp and the updated duration.
+    public static let hangEventUpdated = Notification.Name("io.embrace.hang.updated")
+
+    /// Posted when a hang ends, indicating that the main thread recovered.
+    ///
+    /// The notification’s `object` is a `WatchdogEvent` representing the
+    /// final timestamp and total hang duration.
+    public static let hangEventEnded = Notification.Name("io.embrace.hang.ended")
+}
```
Tischel, line 68: Do you think there's value in exposing these events?
Otherwise we could just use the internal notification center form them that we already have in place.

NachoEmbrace, line 68: Do you think there's much value in exposing these events?
I think we could just use the internal notification center we have in place instead.

naftaly: I believe it's pretty useful. I've used things like this in the past to augment my own events unrelated to hangs.

NachoEmbrace (approved):

PR #390, opened 2025-09-18
[DRAFT] 7.0

NachoEmbrace:


Sources/EmbraceCaptureService/CaptureService+OTel.swift
```diff
@@ -0,0 +1,111 @@
...
+    ///   - startTime: Start time of the span. Defaults to the current time.
+    ///   - endTime: End time of the span, if any.
+    ///   - events: Events for the span.
+    ///   - links: Links for the span.
+    ///   - attributes: Attributes of the span.
+    ///   - autoTerminationCode: If a code is passed, the span will be automatically ended when the current Embrace session ends and will have a special attribute with the given code.
+    /// - Returns: The newly created `EmbraceSpan`.
+    /// - Throws: `EmbraceOTelError.spanLimitReached` if the span limit has been reached for the current Embrace session.
+    @discardableResult
+    public func createSpan(
```
naftaly, line 31: I've always found it confusing that we can create spans from many different places. I've wonder if we might pull it in a bit and make sure there's only one place that it can be done from.

Sources/EmbraceCaptureService/CaptureService.swift
```diff
@@ -22,11 +21,10 @@ import OpenTelemetryApi
 ///
 /// This base class provides the necessary functionality and structure that should be used
 /// by all capture services.
-@objc(EMBCaptureService)
 open class CaptureService: NSObject {
```
naftaly, line 24: remove NSObject

Project.swift
```diff
@@ -36,7 +36,6 @@ let project = Project(
                 .target(name: "EmbraceCaptureService"),
                 .target(name: "EmbraceCommonInternal"),
                 .target(name: "EmbraceConfigInternal"),
-                .target(name: "EmbraceOTelInternal"),
```
ArielDemarco, line 39: I think this changes are unnecessary. We're not really building `xcframeworks` any more so maintaing the `Project.swift` might be unnecessary.

Sources/EmbraceCaptureService/CaptureService.swift
```diff
@@ -35,7 +33,7 @@ open class CaptureService: NSObject {
     @ThreadSafe
     private(set) public var state: CaptureServiceState = .uninstalled
 
-    public func install(otel: EmbraceOpenTelemetry?, logger: InternalLogger? = nil) {
+    package func install(otel: EmbraceOTelSignalsHandler?, logger: InternalLogger? = nil) {
```
ArielDemarco, line 36: I like all this methods are now `package` instead of `public` 👍

Sources/EmbraceCommonInternal/OTelBridge/EmbraceOTelSignalBridge.swift
```diff
@@ -0,0 +1,43 @@
...
+    /// Called when a span is created and started
+    func startSpan(
+        name: String,
+        parentSpan: EmbraceSpan?,
+        status: EmbraceSpanStatus,
+        startTime: Date,
+        endTime: Date?,
+        events: [EmbraceSpanEvent],
+        links: [EmbraceSpanLink],
+        attributes: [String: String]
```
ArielDemarco, line 23: As discussed, we may want to improve this so it respects the original type. IMO i'd abstract to an internal implementation so it's not a `[String: Any]` (I think it'll be a bit more difficult to then transform this to `[String: AttributeValue]` if the dictionary uses `Any`)

Sources/EmbraceCommonInternal/OTelBridge/EmbraceOTelSignalsHandler.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    func _addSessionEvent(
+        name: String,
+        type: EmbraceType?,
+        timestamp: Date,
+        attributes: [String: String],
+        isInternal: Bool
+    ) throws
+
+    func _log(
```
ArielDemarco, line 37: Why the `_`?  As far as I can see, the names in the `extension` (which seems they're created to be used internally based on the usage of the `package` keyword) are already different than the ones in the protocol.

Sources/EmbraceCore/Capture/CaptureServices.swift
```diff
@@ -23,14 +24,17 @@ final class CaptureServices {
     var crashReporter: EmbraceCrashReporter?
 
     weak var config: EmbraceConfigurable?
+    weak var otel: InternalOTelSignalsHandler?
```
ArielDemarco, line 27: Should those be `weak`? are we expecting the injected objects to be deallocated at some point? By making them `strong` It'd make our life easier as we can define them as non-optional and that'll make our logic more resilient.

Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
```diff
@@ -141,21 +139,24 @@ extension HangCaptureService: HangObserver {
...
-                    "sample_code": .int(0),
-                    "stacktrace": .string(stackString)
-                ],
-                timestamp: at.date
-            )
+        queue.async { [self] in
+            do {
+                let stackString = frames.joined()
+                try span.addEvent(
+                    name: SpanEventSemantics.ThreadBlockage.name,
```
ArielDemarco, line 146: Nitpick: I'd name this `ThreadBlockageSample`

Sources/EmbraceCore/Capture/Network/URLSessionTaskHandler.swift
```diff
@@ -109,6 +107,11 @@ final class DefaultURLSessionTaskHandler: NSObject, URLSessionTaskHandler {
                 attributes[SpanSemantics.NetworkRequest.keyMethod] = httpMethod
             }
 
+            // This should be modified if we start doing this for streaming tasks.
+            if let bodySize = request.httpBody {
+                attributes[SpanSemantics.NetworkRequest.keyBodySize] = String(bodySize.count)
+            }
```
ArielDemarco, line 113: Nice catch moving this up here.

Sources/EmbraceCore/Capture/PushNotifications/PushNotificationCaptureService.swift
```diff
@@ -34,6 +33,10 @@ import UserNotifications
         self.options = options
         self.lock = lock
         self.proxy = UNUserNotificationCenterDelegateProxy(captureData: options.captureData)
+
+        super.init()
+
+        proxy.otel = otel
```
ArielDemarco: Can you modify the `UNUserNotificationCenterDelegateProxy` constructor to accept the OTel directly? seems cleaner

Sources/EmbraceCore/Capture/System/LowMemoryWarningCaptureService.swift
```diff
@@ -3,21 +3,17 @@
...
     import EmbraceCommonInternal
-    import EmbraceOTelInternal
     import EmbraceSemantics
 #endif
 
 /// Service that generates OpenTelemetry span events when the application receives a low memory warning.
 @objc(EMBLowMemoryWarningCaptureService)
 public class LowMemoryWarningCaptureService: CaptureService {
 
-    public var onWarningCaptured: (() -> Void)?
```
ArielDemarco, line 19: Finally 😂

Sources/EmbraceCore/Capture/UX/View/Protocols/InstrumentableViewController.swift
```diff
@@ -23,46 +22,21 @@
...
-        ///    - name: The name of the span.
-        ///    - type: The type of the span. Will be set as the `emb.type` attribute.
-        ///    - startTime: The start time of the span.
-        ///    - endTime: The end time of the span.
-        ///    - attributes: A dictionary of attributes to set on the span.
-        /// - Throws: `ViewCaptureService.noServiceFound` if no `ViewCaptureService` is active.
-        /// - Throws: `ViewCaptureService.firstRenderInstrumentationDisabled` if this functionallity was not enabled when setting up the `ViewCaptureService`, or the remote configuration for this feature was not enabled.
-        /// - Throws: `ViewCaptureService.parentSpanNotFound` if no parent span was found for this `UIViewController`.
-        ///           This could mean the `UIViewController` was already rendered / deemed interactive, or the `UIViewController` has already disappeared.
-        public func recordCompletedChildSpan(
```
ArielDemarco, line 58: One of the reasons to use this rather than the other was that we're not returning an a `Span` as it's now immutable. If we follow the other approach (of letting you create an `EmbraceSpan`, and optionally add the `.end` time), is the ended `EmbraceSpan` immutable?

Sources/EmbraceCore/Embrace.swift
```diff
@@ -154,44 +155,58 @@ import Foundation
 
     init(
         options: Embrace.Options,
-        logControllable: LogControllable? = nil,
```
ArielDemarco, line 157: IDK if we should do this. It'd allow to do proper testing of `Embrace`. That said, I think we're not doing it right now

Sources/EmbraceCore/Internal/Embrace+Setup.swift
```diff
@@ -121,26 +120,38 @@ extension Embrace {
 /// Extension to handle observability of SDK startup
 extension Embrace {
 
-    func createProcessStartSpan() -> Span {
-        let builder = buildSpan(name: "emb-process-launch", type: .performance)
-            .markAsPrivate()
+    func createProcessStartSpans() -> [EmbraceSpan] {
```
ArielDemarco, line 123: Is this thing still worth it considering we have the automatic startup instrumentation?

Sources/EmbraceCore/Embrace.swift
```diff
@@ -441,7 +417,6 @@ import Foundation
     /// Called every time the remote config changes
     @objc private func onConfigUpdated() {
         Embrace.logger.limits = config.internalLogLimits
-        Embrace.client?.logController.limits = config.logsLimits
```
ArielDemarco, line 444: This is because `logSeverityLimits` is now attached to the limiter inside `DefaultOTelSignalsHandler`? does that cover all the usecases to add limits?

Sources/EmbraceCore/Internal/Identifiers/DeviceIdentifier+Persistence.swift
```diff
@@ -5,24 +5,24 @@
 import Foundation
 
 #if !EMBRACE_COCOAPOD_BUILDING_SDK
-    import EmbraceCommonInternal
+    import EmbraceSemantics
 #endif
 
-extension DeviceIdentifier {
-    static func retrieve(fileURL: URL?) -> DeviceIdentifier {
+class DeviceIdentifierHelper {
```
ArielDemarco: Nitpick: I don't like the `Helper` suffix; I'd rather name it something more specific... like:
```suggestion
class DeviceIdentifierProvider {
```

Sources/EmbraceCore/Internal/Logs/Batcher/DefaultLogBatcher.swift
```diff
@@ -3,65 +3,43 @@
...
-    var limits: LogsLimits { get }
+
+    var logBatchLimits: LogBatchLimits { get }
+    var delegate: LogBatcherDelegate? { get set }
+    var batch: LogsBatch? { get }
 }
 
 class DefaultLogBatcher: LogBatcher {
-    private let repository: LogRepository
+    private(set) var logBatchLimits: LogBatchLimits
```
ArielDemarco: I dont think anybody is changing the `logBatchLimits`. I understand now the `LogController` is using this value to perform it's logic to divide in batches. So i'd rather make this `internal` and `let`
```suggestion
    let logBatchLimits: LogBatchLimits
```

Sources/EmbraceCore/Internal/Logs/EmbraceLogAttributesBuilder.swift
```diff
@@ -115,41 +111,29 @@ class EmbraceLogAttributesBuilder {
...
-            attributes[LogSemantics.keyState] == nil
-        else {
+    func addApplicationState(_ state: SessionState?) -> Self {
+        guard let state = state else {
             return self
         }
-        attributes[LogSemantics.keyState] = state
+        attributes[LogSemantics.keyState] = state.rawValue
         return self
     }
```
ArielDemarco, line 124: Seems weird that each `addApplicationState` has some specific logic (one of them overriding the previous state, and the other not). Why don't you unify both methods? something like this
```swift
@discardableResult
func addApplicationState(state: SessionState? = nil) -> Self {
     // Dont change the already added value
     guard attributes[LogSemantics.keyState] == nil else {
          return self
     }
     
     // if value was provided, use that one
     if let state = state {
          attributes[LogSemantics.keyState] = state.rawValue
          return self
     } 
     
     // if we reached this point, try to get the state from the current session (if present)
     if let currentSessionState = currentSession?.state {
          attributes[LogSemantics.keyState] = currentSessionState.rawValue
     }

     return self
}
````

File: Sources/EmbraceCore/Internal/Logs/LogController.swift [hunk 19]
```diff
@@ -10,61 +10,57 @@ import Foundation
...
-        var limits: LogsLimits = LogsLimits()
+    var currentSessionId: EmbraceIdentifier? {
+        sessionController?.currentSession?.id
     }
-    private let state = EmbraceMutex(MutableState())
 
-    var limits: LogsLimits {
-        get { state.withLock { $0.limits } }
-        set { state.withLock { $0.limits = newValue } }
+    struct Constants {
```
ArielDemarco:
```suggestion
    private struct Constants {
```

Sources/EmbraceCore/Internal/Logs/LogController.swift
```diff
@@ -153,55 +149,77 @@ class LogController: LogControllable {
...
+                attributes: finalAttributes,
+                sessionId: sessionController.currentSession?.id
+            )
+
+            if send {
+                addLog(log)
             }
 
-            otel.log(message, severity: severity, timestamp: timestamp, attributes: finalAttributes)
+            completion?(log)
```
ArielDemarco, line 200: Can you explain me why this changed in the interface? When wouldn't we want to "send" the log?

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
```diff
@@ -0,0 +1,375 @@
...
+
+        guard let span = sessionController?.currentSessionSpan else {
+            throw EmbraceOTelError.invalidSession
+        }
+
+        try span.addSessionEvent(
+            name: name,
+            type: type,
+            timestamp: timestamp,
+            attributes: attributes,
```
ArielDemarco, line 107: Genuine question: don't we need to sanitize the name/attributes for this?

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
```diff
@@ -0,0 +1,375 @@
...
+
+        // check limits
+        var onlyUpdate = false
+        if !limiter.shouldCreateCustomSpan() {
+            onlyUpdate = true
+            Embrace.logger.warning("Limit reached for spans on the current Embrace session!")
+        }
+
+        // update db
+        storage?.upsertSpan(span, onlyUpdate: onlyUpdate)
```
ArielDemarco, line 357: Should this even happen? Call the `onStart` after the span was already created in order to update it?

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
```diff
@@ -0,0 +1,341 @@
...
+        try _addEvent(
+            name: name,
+            type: type,
+            timestamp: timestamp,
+            attributes: attributes,
+            isInternal: false
+        )
+    }
+
+    func _addEvent(
```
ArielDemarco, line 122: Why the `_`? better make it `private`; the extra arguments are enough to make them different

Sources/EmbraceCommonInternal/OTelBridge/EmbraceOTelSignalBridge.swift
naftaly: See this for a way to do it. https://github.com/embrace-io/embrace-apple-sdk/blob/de07ba68533561633309794f2e40b50df6f4e489/Sources/EmbraceCommonInternal/CrashReporter/CrashReporter.swift#L197-L215

Sources/EmbraceCommonInternal/OTelBridge/EmbraceOTelSignalBridge.swift
```diff
@@ -0,0 +1,43 @@
...
+
+/// Protocol used to bridge telemetry signals created through our SDK into 3rd party OTel implementations
+public protocol EmbraceOTelSignalBridge {
+
+    /// Called when a span is created and started
+    func startSpan(
+        name: String,
+        parentSpan: EmbraceSpan?,
+        status: EmbraceSpanStatus,
+        startTime: Date,
```
naftaly, line 19: Let's build our own clock and work with monotonic times. we can expand to Date as required by exporters. See NanosecondClock in the HangCaptureService, we could expand it.

Sources/EmbraceCore/Public/OTel/Events/DefaultOTelSignalsHandler+Breadcrumb.swift
```diff
@@ -0,0 +1,41 @@
...
+    /// - Parameters:
+    ///   - message: Message of the breadcrumb.
+    ///   - timestamp: Timestamp of the breadcrumb.
+    ///   - attributes: Attributes of the breadcrumb.
+    /// - Throws: `EmbraceOTelError.invalidSession` if there is not active Embrace session.
+    /// - Throws: `EmbraceOTelError.spanEventLimitReached` if the limit hass ben reached for the given span even type.
+    public func addBreadcrumb(
+        _ message: String,
+        timestamp: Date = Date(),
+        attributes: [String: String] = [:]
```
ArielDemarco, line 23: Shouldn't this be part of the `Embrace` public API? a "Breadcrumb" is an Embrace concept rather than an OTel one

Sources/EmbraceCore/Public/OTel/Events/DefaultOTelSignalsHandler+PushNotification.swift
```diff
@@ -3,76 +3,90 @@
...
+    /// - Throws: `EmbraceOTelError.invalidSession` if there is not active Embrace session.
+    /// - Throws: `EmbraceOTelError.spanEventLimitReached` if the limit hass ben reached for the given span even type.
     /// - Throws: `PushNotificationError.invalidPayload` if the `aps` object is not present in the `userInfo` of the `UNNotification`.
-    init(
+    public func addPushNotificationEvent(
         userInfo: [AnyHashable: Any],
         timestamp: Date = Date(),
-        attributes: [String: AttributeValue] = [:],
+        attributes: [String: String] = [:],
         captureData: Bool = true
```
ArielDemarco, line 56: Same comment as above: shouldn't this kind of "helper"/"convenience" methods be closer to the public API rather than be wrapped inside the `otel` object?

Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -182,45 +192,34 @@ class SessionController: SessionControllable {
...
-            }
+            currentSessionSpan?.end(endTime: now)
+
+            // TODO: Check if needed
+            // Manually updating the span record synchronously.
+            //            storage?.endSpan(
+            //                id: currentSessionSpan.context.spanId.hexString,
+            //                traceId: currentSessionSpan.context.traceId.hexString,
+            //                endTime: now
+            //            )
```
ArielDemarco, line 216: Why do you think is this unnecessary right now? shouldn't this prevent the data races we had in the past?

Sources/EmbraceCore/SwiftUI/Internal/EmbraceTraceViewLogger.swift
```diff
@@ -3,21 +3,20 @@
...
     import EmbraceSemantics
 #endif
 
 /// The environment key for injecting `EmbraceTraceViewLogger` into SwiftUI's `EnvironmentValues`.
 @available(iOS 13, macOS 10.15, tvOS 13, watchOS 6.0, *)
 private struct EmbraceTraceViewLoggerEnvironmentKey: EnvironmentKey {
+    // TODO: pass otel handler
     static let defaultValue: EmbraceTraceViewLogger = EmbraceTraceViewLogger(
-        otel: Embrace.client,
+        otel: nil,
```
ArielDemarco: You should probably inject a global provider in here... `EnvironmentKey`s require to define the `defaultValue`. So, even though `Embrace.client` right now shouldn't be "useful" maybe a global wrapper over the `otel` property that lives inside the client

Sources/EmbraceCore/SwiftUI/Internal/EmbraceTraceViewLogger.swift
```diff
@@ -112,28 +111,18 @@ extension EmbraceTraceViewLogger {
...
 
         // Build the span with full name: "swiftui.view.<viewName>.<semantics>"
-        let builder = client.buildSpan(
+        return try? otel.createInternalSpan(
             name: "emb-swiftui.view.\(name).\(semantics)",
-            type: SpanType.viewLoad,
-            attributes: attributes ?? [:],
-            autoTerminationCode: nil
+            parentSpan: parent,
+            type: .viewLoad,
```
ArielDemarco, line 122: you're missing adding the `time` (as `startTime`) when this methods receives one

Sources/EmbraceCore/Utils/W3C/W3C+TraceParent.swift
```diff
@@ -35,7 +32,7 @@ extension W3C {
     ///     - traceId: The Span's traceId
     ///     - spanId: The Span's spanId
     ///     - sampled: Whether the trace is sampled
-    public static func traceparent(traceId: String, spanId: String, sampled: Bool = false) -> String {
+    public static func traceparent(traceId: String, spanId: String, sampled: Bool = true) -> String {
```
ArielDemarco, line 35: I've doubts on this; probably it'd be good to test the actual received value today to see. Seems that by default `TraceFlags` has `isSampled` as `0x1` (`true`)

Sources/EmbraceOTelInternal/EmbraceOpenTelemetry.swift
```diff
@@ -1,7 +1,7 @@
 //
 //  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
 //
-
+/*
```
ArielDemarco, line 4: This is commented; is this still an useful file?

Sources/EmbraceSemantics/Log/EmbraceLog.swift
```diff
@@ -0,0 +1,33 @@
...
+    var type: EmbraceType { get }
+
+    /// Date when the log was emitted
+    var timestamp: Date { get }
+
+    /// Contents of the log
+    var body: String { get }
+
+    /// Attributes of the log
+    var attributes: [String: String] { get }
```
ArielDemarco, line 26: Mentioned before; maybe is good to reconsider using an internal interface for the `Value`

Sources/EmbraceSemantics/Log/EmbraceLogSeverity.swift
```diff
@@ -1,10 +1,8 @@
 //
-//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
 //
 
-import Foundation
-
-@objc public enum LogSeverity: Int, Codable {
+@objc public enum EmbraceLogSeverity: Int, Codable {
```
ArielDemarco, line 5:
```suggestion
public enum EmbraceLogSeverity: Int, Codable {
```

Sources/EmbraceSemantics/EmbraceIdentifier.swift
```diff
@@ -0,0 +1,86 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Wrapper around UUID used for all Embrace signals.
+@objc public class EmbraceIdentifier: NSObject {
```
ArielDemarco, line 8: You can get rid of all the `@objc` and `NSObject` related stuff

Sources/EmbraceSemantics/EmbraceType.swift
```diff
@@ -0,0 +1,108 @@
...
+//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+/// An EmbraceType is an Embrace specific concept to allow for better categorization of Telemetry Primitives.
+/// - This struct will be serialized into an `emb.type` attribute.
+/// - This struct is encoded as a String with the format `<primary>.<secondary>`.
+/// - The primary category is required, but the secondary category is optional.
+@objc public class EmbraceType: NSObject, RawRepresentable {
```
ArielDemarco, line 11: You can get rid of all the `@objc` and `NSObject` related stuff

Sources/EmbraceSemantics/EmbraceType.swift
```diff
@@ -0,0 +1,108 @@
...
+    @objc override public var hash: Int {
+        var hasher = Hasher()
+        hasher.combine(primary)
+        hasher.combine(secondary ?? "")
+        return hasher.finalize()
+    }
+}
+
+/// Top level category for the EmbraceType
+@objc public enum PrimaryType: Int, CaseIterable {
```
ArielDemarco, line 83: Idem above

Sources/EmbraceSemantics/Log/EmbraceLog.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Represents an OTel log signal.
+@objc public protocol EmbraceLog {
```
ArielDemarco, line 8:
```suggestion
public protocol EmbraceLog {
```

Sources/EmbraceSemantics/Log/EmbraceLogAttachment.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Class used to add attachments when creating `EmbraceLogs`.
+@objc
+public class EmbraceLogAttachment: NSObject {
```
ArielDemarco, line 9: You can get rid of all the `@objc` and `NSObject` related stuff

Sources/EmbraceSemantics/Log/StackTrace/EmbraceStackTrace.swift
```diff
@@ -4,7 +4,9 @@
 
 import Foundation
 
-public struct EmbraceStackTrace: Equatable {
+/// Class used to add custom stack traces to `EmbraceLogs`.
+@objc
+public class EmbraceStackTrace: NSObject {
```
ArielDemarco, line 9:
```suggestion
public class EmbraceStackTrace {
```

Sources/EmbraceSemantics/Log/StackTrace/EmbraceStackTrace.swift
```diff
@@ -32,8 +34,8 @@ public struct EmbraceStackTrace: Equatable {
     /// - Throws: An `EmbraceStackTraceError.invalidFormat` if any of the frames are not in the expected format
     /// - Throws: An `EmbraceStackTraceError.frameIsTooLong` if any of the frames has more than the `maximumFrameLength` (10.000 characters).
     ///
-    /// - Important: a stacktrace can't have more than `maximumAmountOfFrames` (200); if that happens, we'll trim the exceeding frames.
-    public init(frames: [String]) throws {
+    /// - Important: A stacktrace can't have more than `maximumAmountOfFrames` (200); if that happens, we'll trim the exceeding frames.
+    @objc public init(frames: [String]) throws {
```
ArielDemarco, line 38:
```suggestion
    public init(frames: [String]) throws {
```

Sources/EmbraceSemantics/Span/EmbraceSpan.swift
```diff
@@ -0,0 +1,88 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Represents an OTel span signal.
+@objc public protocol EmbraceSpan {
```
ArielDemarco, line 8:
```suggestion
public protocol EmbraceSpan {
```

Sources/EmbraceSemantics/Span/EmbraceSpanContext.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Represents the context for a `EmbraceSpan`.
+@objc public class EmbraceSpanContext: NSObject {
```
ArielDemarco, line 8: You can get rid of all the `@objc` and `NSObject` related stuff

Sources/EmbraceSemantics/Span/EmbraceSpanErrorCode.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
+//
+
+/// Embrace specific error status for spans
+@objc public enum EmbraceSpanErrorCode: Int {
```
ArielDemarco, line 6: You can get rid of all the `@objc` and `NSObject` related stuff and at the same time change this to be a `String` enum. In that way the `name` is just the `rawValue`

Sources/EmbraceSemantics/Span/EmbraceSpanEvent.swift
```diff
@@ -0,0 +1,46 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Represents an OTel span event
+@objc open class EmbraceSpanEvent: NSObject {
```
ArielDemarco, line 8: You can get rid of all the `@objc` and `NSObject` related stuff

Sources/EmbraceSemantics/Span/EmbraceSpan.swift
```diff
@@ -0,0 +1,88 @@
...
+    var endTime: Date? { get }
+
+    /// Array of events in the span
+    var events: [EmbraceSpanEvent] { get }
+
+    /// Array of links in the span
+    var links: [EmbraceSpanLink] { get }
+
+    /// Attributes of the span
+    var attributes: [String: String] { get }
```
ArielDemarco, line 38: Mentioned before; maybe is good to reconsider using an internal interface for the `Value`

Sources/EmbraceSemantics/Span/EmbraceSpanEvent.swift
```diff
@@ -0,0 +1,46 @@
...
+    @objc public let name: String
+
+    /// Embrace specific type of the event, if any
+    @objc public let type: EmbraceType?
+
+    /// Date when the event occured
+    @objc public let timestamp: Date
+
+    /// Attributes of the event
+    @objc public let attributes: [String: String]
```
ArielDemarco, line 20: Mentioned before; maybe is good to reconsider using an internal interface for the `Value`

Sources/EmbraceSemantics/Span/EmbraceSpanLink.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Represents an OTel span link
+@objc public class EmbraceSpanLink: NSObject {
```
ArielDemarco, line 8: You can get rid of all the `@objc` and `NSObject` related stuff

Sources/EmbraceSemantics/Span/EmbraceSpanStatus.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
+//
+
+/// Defines all possible statuses for spans.
+@objc public enum EmbraceSpanStatus: Int {
```
ArielDemarco, line 6: To keep it simple, you can just transform this to be `EmbraceSpanStatus: String`

Sources/EmbraceCore/Public/OTel/Events/DefaultOTelSignalsHandler+Breadcrumb.swift
naftaly: out of curiosity, what's the diff between a breadcrumb and a log. Basically the same API, and same display in our dash.

Sources/EmbraceCommonInternal/OTelBridge/EmbraceOTelSignalsHandler.swift
NachoEmbrace: Was made to make it obvious that this was the internal method and not the one exposed publicly in `DefaultOTelSignalsHandler`. I was a bit worried it was going to be confusing having the signatures being too similar and not being sure which one was being called at a glance while looking at the code.
That's why I went for something really obvious with the "_" prefix.

In reality these methods are not really called from other parts of the SDK since capture services are using the `addInternalSessionEvent`, `addInternalSessionEvent` and `internalLog` methods (which use these _ prefixed methods internally).

Sources/EmbraceCore/Capture/CaptureServices.swift
NachoEmbrace: I don't really have a strong opinion either way.
This object will always be held by the Embrace client and injected into the capture services.
Since I know the object is being held somewhere else I opted for a weak reference, that's about it.

Sources/EmbraceCaptureService/CaptureService+OTel.swift
NachoEmbrace: In reality there is only 1 place to do it from (technically 2 if you count using the OTel SDK directly).
These methods in CaptureService are just for convenience so users don't have to do Embrace.client?.createSpan.. in their custom capture services.

Sources/EmbraceCore/Capture/PushNotifications/PushNotificationCaptureService.swift
NachoEmbrace: This is actually wrong entirely.
The otel reference is injected on install, so the set has to be moved there.

Sources/EmbraceCore/Capture/UX/View/Protocols/InstrumentableViewController.swift
NachoEmbrace: I haven't added mutability checks related to the end time yet.
If we want to do that it would be easy to achieve by modifying DefaultEmbraceSpan.
We could even add a new error type to throw when ended spans are attempted to be mutated.

Sources/EmbraceCore/Embrace.swift
NachoEmbrace: The embrace client tests have been disabled for a long time.
If we ever figure out how to do those properly we can add this back.

Sources/EmbraceCore/Internal/Embrace+Setup.swift
NachoEmbrace: I kept them to not break anything.
Not really sure if we or our customers are using these spans.
I agree that they seem redundant and should be removed.

Sources/EmbraceCore/Embrace.swift
NachoEmbrace: Yes and yes.
LogController only gets called from the signals handler which applies the limits beforehand.

Sources/EmbraceCore/Internal/Logs/LogController.swift
NachoEmbrace: This whole "send" parameter is to maintain the logic we had where we'd export some logs but not include them in our normal log batches.
This was used for special cases like crashes.

This logic was part of the processor / exporters before and has now been moved to the signals handler and log controller.

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
NachoEmbrace: They do get sanitized.
`span.addSessionEvent` internally goes through the limiters and sanitizers.

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
NachoEmbrace: This is the method that would be called from the EmbraceOTel module (name TBD).
Meaning this span was created directly with the OTel SDK and we need to handle it.
We don't really know the contents of the span so we have to expect anything.

If the limit of custom spans was reached already, we will always update the span if for some reason we have it.
I agree that in a perfect world it wouldn't make much sense for the span to already exist in our db, but I decided to make the code as flexible as possible and contemplate even weird cases like this.

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
NachoEmbrace: Same reason as before, just to make it more obvious when reading the code.
I'm ok going with a different naming convention or removing the prefix entirely.

Sources/EmbraceCore/Public/OTel/Events/DefaultOTelSignalsHandler+Breadcrumb.swift
NachoEmbrace: > Shouldn't this be part of the `Embrace` public API? a "Breadcrumb" is an Embrace concept rather than an OTel one

I think regardless it will end up being Embrace.client.addBreadcrumb (since we pretty much decided to remove the accessors entirely, aka no more Embrace.client.otel, Embrace.client.metadata, etc)

Sources/EmbraceCore/Session/SessionController.swift
NachoEmbrace: This race condition was happening because a part of the logic was happening on our span processor.
Obviously this is not the case anymore so I have to double check this is working fine now.

Sources/EmbraceCore/SwiftUI/Internal/EmbraceTraceViewLogger.swift
NachoEmbrace: I just forgot to clean this up, I was meant to simply pass `Embrace.client?.otel`

Sources/EmbraceCore/SwiftUI/Internal/EmbraceTraceViewLogger.swift
NachoEmbrace: Good catch

Sources/EmbraceOTelInternal/EmbraceOpenTelemetry.swift
NachoEmbrace: The EmbraceOTelInternal module is not really being used, but I left the code since a big part of it will still be re-used when we make the new bridge module.

Sources/EmbraceCore/Utils/W3C/W3C+TraceParent.swift
NachoEmbrace: I changed the default to true so the result would be the same as before.

Sources/EmbraceCaptureService/CaptureService.swift
NachoEmbrace: I'll handle the drop of objc support in a different pr.

PR #393, opened 2025-09-23
Added fix to allow pushing other variants of the OS to the backend

ArielDemarco:
# Overview 
We were incorrectly sending the OS variant (aka. the `systemName`) in the `os_name` field when sending traces/logs to the backend. This caused payloads from other platforms (such as tvOS) to be dropped at ingestion time.

This PR fixes that.

### DIFF
--- a/Sources/EmbraceCore/Payload/ResourcePayload.swift
+++ b/Sources/EmbraceCore/Payload/ResourcePayload.swift
@@ -30,6 +30,7 @@ struct ResourcePayload: Codable {
     var environmentDetail: String?
     var appFramework: Int?
     var launchCount: Int?
+    var sdkPlatform: String?
     var sdkVersion: String?
     var appVersion: String?
     var appBundleId: String?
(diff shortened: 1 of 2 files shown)


Sources/EmbraceCore/Payload/ResourcePayload.swift
```diff
@@ -110,7 +113,10 @@ struct ResourcePayload: Codable {
 
         // bundle_id is constant and won't change over app install lifetime
         self.appBundleId = Bundle.main.bundleIdentifier
-        self.osName = EMBDevice.operatingSystemType
+        // The backend uses this values to identify the platform through out all the system
+        // The specific OS is going to be reported in the `osAlternateType` field
+        self.osName = "ios"
```
naftaly, line 118: it's odd to hard code this here? Shouldn't we instead extend the backend to support the other systems (tvOS, watchOD, macOS, ...)?

ArielDemarco: It was like that before (we expected the backend to behave in that way). However the backend uses that specific value to basically split "features" that are (let's say) Apple related through out different places. Ideally, eventually they could use something more generic like the value we send in `os_type` (`darwin`). But that's why I commented as I understand is misleading.

naftaly: the question stands, why not let the backend do the work, no real need to change it here.

naftaly: what I mean is instead of hacking it, I think the more global change isn't complicated enough to warrant a hack that we'll probably never get back to fixing.

PR #394, opened 2025-09-26
Built in atomics

naftaly:
This branch introduces a comprehensive atomic operations system for the Embrace Apple SDK. The implementation provides thread-safe, lock-free atomic operations with proper memory ordering semantics.

Key Changes:
  - New atomic wrapper system: Added EmbraceAtomic<T> generic class that provides type-safe atomic operations for various data types
  - C interop layer: Created EmbraceAtomicsShim with C headers and implementation to bridge Swift with C11/C++11 atomic primitives
  - Memory ordering support: Implemented full memory ordering semantics (relaxed, acquire, release, sequential consistency, etc.)
  - Comprehensive test suite: Added 278 lines of tests covering atomic operations, memory ordering, and thread safety

Files Added:
  - EmbraceAtomic.swift - Main Swift atomic wrapper API
  - EmbraceAtomic+Imp.swift - Implementation details and type protocols
  - EmbraceAtomicsShim.{h,m} - C interop layer for atomic operations
  - EmbraceAtomicTests.swift - Test suite for atomic functionality
  - Updated Package.swift to include new dependencies

The implementation provides a safe, performant foundation for concurrent programming within the SDK, replacing potential race conditions with proper atomic operations.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -115,6 +115,7 @@ let package = Package(
         .target(
             name: "EmbraceCommonInternal",
             dependencies: [
+                "EmbraceAtomicsShim",
                 .product(name: "OpenTelemetrySdk", package: "opentelemetry-swift-core")
             ]
         ),
@@ -124,6 +125,12 @@ let package = Package(
                 "EmbraceCommonInternal",
                 "TestSupport"
             ]
+        ),
+
+        // atomics ------------------------------------------------------------------
+        .target(
+            name: "EmbraceAtomicsShim",
+            publicHeadersPath: "include"
         ),
 
         // semantics -----------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


NachoEmbrace:
There doesn't seem to be a need for adding a new module. Can't we just drop this in `EmbraceCommonInternal`?
Also I don't see this being used anywhere, is that in a different PR?

naftaly:
> Edit: Just realized it's in a separate module to use ObjC so disregard the first question (though we should probably drop it in `EmbraceObjCUtilsInternal` instead).

I split it up to make it very clear what it was, I don't see it as a problem for it to be modular like it is. Does it cause issues like longer build times or something?

> Also I don't see this being used anywhere, is that in a different PR?

It's not used yet. There are tons of places I plan on using it instead of locks to make things simpler and more performant.

PR #401, opened 2025-10-01
Improve Startup Fist Frame Tracking

naftaly:


Examples/EmbraceIOTestApp/EmbraceIOTestApp/Common/Protocols/PayloadTestProtocol.swift
```diff
@@ -105,4 +108,10 @@ extension PayloadTest {
 
     // TODO: Remove default initializer when 'testRelevantPayloadNames' is removed.
     var expectedNotificationsForTestReady: [String] { [] }
+
+    func resetStartupInstrumentation() {
```
ferdraghi: I think you should rather implement `runTestPreparations` on the tests that need to run this code and run it on there instead. That's what the method is for. The tests already try to run that method automatically and will trigger the actual test once the target span is found on the exporter.

naftaly: There were about 10 payloads that needed to implement the same code so I just wrapped it in `resetStartupInstrumentation` and they all call that from `runTestPreparations`. So unless I'm misunderstanding, this is what I'm already doing.

ie: this is implemented on about 10 payloads.

```
func runTestPreparations() {
   resetStartupInstrumentation()
}
```

ferdraghi: Wrapping it into a method that would be called on `runTestPreparations` is fine, but I'm not sure about adding the declaration into the test protocol itself... Maybe we could have like a `TestUtils` sort of class for methods that several tests need to run?... Speaking of which, I just remembered that I implemented `testMetadataInclussion` into a class method "temporarily" because several tests needed to run it and also had the idea to eventually move it into a "test tools" class or even an extension, but never got around to it... probably a good idea to start it?

naftaly: Does that make sense @ferdraghi. If I didn't do it the correct way, could you leave a small snippet to show the right way?

Sources/EmbraceCore/Embrace.swift
```diff
@@ -288,6 +286,14 @@ import Foundation
 
         EMBStartupTracker.shared().sdkStartStartTime = Date()
 
+        if EMBStartupTracker.shared().appDidFinishLaunchingEndTime != nil {
+            Embrace.logger.error("Embrace SDK can't start after didFinishLaunching if you want startup instrumentation to work.")
```
NachoEmbrace:
```suggestion
            Embrace.logger.warning("The Embrace SDK startup instrumentation won't work correctly if the SDK is started after didFinishLaunching.")
```

Sources/EmbraceCore/Embrace.swift
```diff
@@ -288,6 +286,14 @@ import Foundation
...
+        }
+
+        if EMBStartupTracker.shared().appFirstDidBecomeActiveTime != nil {
+            Embrace.logger.error("Embrace SDK can't start after the first didBecomeActive if you want startup instrumentation to work.")
```
NachoEmbrace:
```suggestion
            Embrace.logger.warning("The Embrace SDK startup instrumentation won't work correctly if the SDK is started after didBecomeActive.")
```

Sources/EmbraceCore/Embrace.swift
naftaly: I feel it's more of an error to start embrace after startup. I'm not really bothered, but it feels the severity is important here. Any reason you think it should just be a warning?

Examples/EmbraceIOTestApp/EmbraceIOTestApp/Common/Protocols/PayloadTestProtocol.swift
ferdraghi: What I meant to say is that I feel like method declarations on the protocol should be for methods that make sense in the context of _all_ tests. Sure, not every test needs to run preparations before testing, but it does make sense that the option is there for every test... This specific method I feel like it only makes sense in the context of certain tests. That's why I suggested instead of adding the declaration on the protocol, to simply start some sort of "TestHelpers" or "TestTools" or whatever object that contain all the helper methods like this one for tests to use as necessary.

Sources/EmbraceCore/Embrace.swift
NachoEmbrace: I don't think this should count as an error since it's only a small thing the SDK won't be able to do.
Regardless of that which is kinda subjective, the main issue I have is the fact that we limit the amount of internal logs we send (which by default means we only send the first 3 internal error logs we produce). This would eat 2 of them which is not ideal.

Also this will probably always trigger in Unity since the SDK initialization is usually delayed (I haven't checked in a while but I think that's still the case). This is not a big deal but figured it was worth mentioning.

naftaly: Good point in the internal limits, I'll change it to warning. I'll also make it only one check instead of two.

Examples/EmbraceIOTestApp/EmbraceIOTestApp/Common/Protocols/PayloadTestProtocol.swift
naftaly: Gotcha, I can fix that part at least to make the PR cleaner.

PR #403, opened 2025-10-03
[Draft] 7.0 - Dropping objc support

NachoEmbrace:
This is pointed to the 7.0 branch.

* Removed objc support from the public interfaces
* Removed objc example app
* Removed stuff that was already flagged as deprecated in 6.x
* Fixed some stuff that was incorrectly flagged as public in 6.x


naftaly:
I look forward to this being in main. All looks good.

PR #535, opened 2026-03-13
[EMBR-11117] Fix: Fix attempt for GZip crash

NachoEmbrace:
This PR attempts to fix a crash that occurs when we compress payload data with GZip on memory pressure situations.

There are 4 layers:
1. **Adaptive batches for logs:** Log batch size changes depending on available memory.
2. **Sequential processing:** Process log batches sequentially so no more than 1 compressed payload lives in memory at the same time.
3. **GZip deflate error check:** this just adds an early error check inside the GZip code so the buffer doesn't keep growing unnecessarily.
4. **Input size guard in GZip:** Added a hard cap of 50mb in the GZip code. Any data larger than this gets dropped.

### DIFF
--- a/Sources/EmbraceCore/Internal/Logs/LogController.swift
+++ b/Sources/EmbraceCore/Internal/Logs/LogController.swift
@@ -320,18 +320,22 @@ extension LogController {
     }
 
     static func adaptiveMaxLogsPerBatch() -> Int {
-        let availableMemory = os_proc_available_memory()
-
-        switch availableMemory {
-        case 0..<(15 * 1024 * 1024):
-            return 1
-        case ..<(30 * 1024 * 1024):
-            return 5
-        case ..<(50 * 1024 * 1024):
-            return 10
-        default:
+        #if os(macOS)
             return maxLogsPerBatch
-        }
+        #else
+            let availableMemory = os_proc_available_memory()
+
+            switch availableMemory {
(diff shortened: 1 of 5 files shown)


Sources/EmbraceCore/Internal/Logs/LogController.swift
```diff
@@ -322,16 +319,31 @@ extension LogController {
         }
     }
 
-    fileprivate func divideInBatches(_ logs: [EmbraceLog]) -> [LogsBatch] {
+    static func adaptiveMaxLogsPerBatch() -> Int {
+        let availableMemory = os_proc_available_memory()
+
+        switch availableMemory {
+        case 0..<(15 * 1024 * 1024):
```
ferdraghi: Is there any reasoning behind these numbers?

Sources/EmbraceCore/Internal/Logs/LogController.swift
```diff
@@ -322,16 +319,31 @@ extension LogController {
         }
     }
 
-    fileprivate func divideInBatches(_ logs: [EmbraceLog]) -> [LogsBatch] {
+    static func adaptiveMaxLogsPerBatch() -> Int {
+        let availableMemory = os_proc_available_memory()
```
ferdraghi: `os_proc_available_memory` is unavailable on macOS.

ferdraghi (approved):

PR #538, opened 2026-03-16
Implement frame-pointer stack walker and profiling test infrastructure.

kstenerud:
Replaced the stub emb_stack_walk with a real implementation that reads thread state (arm64 + x86_64), walks the frame pointer chain with ptrauth stripping, and uses stack bounds to guard against unsafe memory access.

The frame type is now uintptr_t rather than uint64_t for better C spec adherence (this is mostly cosmetic since all pointers on Apple hardware is 64 bits now).

The tests can be run with and without frame pointers (EmbraceProfilingTestSupportNoFP target), but these won't pass until KSCrash is updated.

Added sampler benchmark tests which will also measure no-frame-pointer fallback mode once available.

## Checklist

### DIFF
--- a/Sources/EmbraceProfiling/ProfilingSample.swift
+++ b/Sources/EmbraceProfiling/ProfilingSample.swift
@@ -8,7 +8,7 @@ public enum StackUnwindMethod {
     case framePointer
     /// KSCrash's DWARF-based unwinder (slower, more robust fallback)
     case kscrash
-    /// Partial frame-pointer walk (below minFPFrames threshold, returned as last resort)
+    /// Partial frame-pointer walk (below minFPFrames threshold, returned as last resort if kscrash returned nothing)
     case framePointerPartial
     /// Stack capture failed entirely — no frames were obtained.
     case failed
--- a/Sources/EmbraceProfiling/ProfilingSample.swift
+++ b/Sources/EmbraceProfiling/ProfilingSample.swift
@@ -2,6 +2,16 @@
 //  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
 //
 
+/// How a stack trace was captured.
+public enum StackUnwindMethod {
+    /// Frame-pointer based stack walking (fast, works when FP is available)
+    case framePointer
+    /// KSCrash's DWARF-based unwinder (slower, more robust fallback)
+    case kscrash
+    /// Partial frame-pointer walk (below minFPFrames threshold, returned as last resort)
+    case framePointerPartial
+}
+
 /// A single profiling sample containing a stack trace captured at a point in time.
 public struct ProfilingSample {
     /// `CLOCK_MONOTONIC_RAW` timestamp in nanoseconds when this sample was captured.
@@ -10,8 +20,12 @@ public struct ProfilingSample {
     /// Raw return addresses from the captured stack trace.
     public let frames: [UInt]
 
-    public init(timestamp: UInt64, frames: [UInt]) {
+    /// The method used to capture this stack trace.
+    public let unwindMethod: StackUnwindMethod
+
+    public init(timestamp: UInt64, frames: [UInt], unwindMethod: StackUnwindMethod) {
         self.timestamp = timestamp
         self.frames = frames
+        self.unwindMethod = unwindMethod
     }
 }
--- a/Sources/EmbraceProfiling/ProfilingSample.swift
+++ b/Sources/EmbraceProfiling/ProfilingSample.swift
@@ -10,6 +10,8 @@ public enum StackUnwindMethod {
     case kscrash
     /// Partial frame-pointer walk (below minFPFrames threshold, returned as last resort)
     case framePointerPartial
+    /// Stack capture failed entirely — no frames were obtained.
+    case failed
 }
 
 /// A single profiling sample containing a stack trace captured at a point in time.
--- a/Sources/EmbraceProfiling/ProfilingSample.swift
+++ b/Sources/EmbraceProfiling/ProfilingSample.swift
@@ -8,9 +8,9 @@ public struct ProfilingSample {
     public let timestamp: UInt64
 
     /// Raw return addresses from the captured stack trace.
-    public let frames: [UInt64]
+    public let frames: [UInt]
 
-    public init(timestamp: UInt64, frames: [UInt64]) {
+    public init(timestamp: UInt64, frames: [UInt]) {
         self.timestamp = timestamp
         self.frames = frames
     }
--- a/Sources/EmbraceProfiling/ProfilingSample.swift
+++ b/Sources/EmbraceProfiling/ProfilingSample.swift
@@ -3,7 +3,7 @@
 //
 
 /// How a stack trace was captured.
-public enum StackUnwindMethod {
+public enum StackUnwindMethod: Equatable, Hashable, CaseIterable, Sendable {
     /// Frame-pointer based stack walking (fast, works when FP is available)
     case framePointer
     /// KSCrash's DWARF-based unwinder (slower, more robust fallback)
--- a/Sources/EmbraceProfiling/ProfilingSample.swift
+++ b/Sources/EmbraceProfiling/ProfilingSample.swift
@@ -1,5 +1,5 @@
 //
-//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
+//  Copyright © 2026 Embrace Mobile, Inc. All rights reserved.
 //
 
 /// How a stack trace was captured.
@@ -15,7 +15,7 @@ public enum StackUnwindMethod {
 }
 
 /// A single profiling sample containing a stack trace captured at a point in time.
-public struct ProfilingSample {
+public struct ProfilingSample: Equatable, Hashable, Sendable {
     /// `CLOCK_MONOTONIC_RAW` timestamp in nanoseconds when this sample was captured.
     public let timestamp: UInt64
 
--- a/Sources/EmbraceProfiling/StackCapture.swift
+++ b/Sources/EmbraceProfiling/StackCapture.swift
@@ -1,5 +1,5 @@
 //
-//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
+//  Copyright © 2026 Embrace Mobile, Inc. All rights reserved.
 //
 
 #if !os(watchOS)
@@ -35,15 +35,29 @@ func captureStack(thread: thread_t, maxFrames: Int = 512) -> (frames: [UInt], me
     var ksFrames = [UInt](repeating: 0, count: maxFrames)
     let pthread = pthread_from_mach_thread_np(thread)
 
+    // Resolve stack bounds before suspending — pthread APIs are not async-safe.
+    var stackBottom: UnsafeRawPointer? = nil
+    var stackTop: UnsafeRawPointer? = nil
+    if let pth = pthread {
+        let top = pthread_get_stackaddr_np(pth)
+        let size = pthread_get_stacksize_np(pth)
+        stackTop = UnsafeRawPointer(top)
+        stackBottom = UnsafeRawPointer(top - size)
+    }
+
     guard thread_suspend(thread) == KERN_SUCCESS else {
         return ([], .failed)
     }
 
-    // Only asyn-safe things allowed between suspend and resume.
+    // Only async-safe things allowed between suspend and resume.
 
     var fpCount = 0
-    // Common case: Frame pointer is available, so we can fast-walk the stack.
-    let fpSuccess = emb_stack_walk(thread, &fpFrames, maxFrames, &fpCount)
+    let fpSuccess: Bool
+    if let bottom = stackBottom, let top = stackTop {
+        fpSuccess = emb_stack_walk(thread, bottom, top, &fpFrames, maxFrames, &fpCount)
+    } else {
+        fpSuccess = false
+    }
 
     // Fall back to KSCrash's unwinder if FP walking didn't produce enough
     // frames.
--- a/Sources/EmbraceProfiling/StackCapture.swift
+++ b/Sources/EmbraceProfiling/StackCapture.swift
@@ -17,45 +17,60 @@ import EmbraceProfilingSampler
 /// successful. Below this threshold we fall back to KSCrash's unwinder.
 private let minFPFrames = 3
 
-/// Captures a stack trace from a suspended thread.
+/// Captures a stack trace from a running thread.
 ///
-/// Uses fast frame-pointer walking first. If that yields fewer than
-/// ``minFPFrames`` frames, falls back to KSCrash's `captureBacktrace`
-/// unwinder.
+/// Suspends the target thread, walks its stack, then resumes the thread.
+///
+/// Tries fast frame-pointer walking first. If that yields fewer than ``minFPFrames``
+/// frames, falls back to KSCrash's `captureBacktrace` unwinder.
 ///
 /// Note: KSCrash will support walking a stack without a FP in the next release.
 ///
-/// The thread **must** already be suspended before calling this function.
-///
 /// - Parameters:
-///   - thread: The Mach thread port of the suspended thread.
+///   - thread: The Mach thread port of the thread to capture.
 ///   - maxFrames: Maximum number of frames to capture.
 /// - Returns: A tuple containing the array of return addresses and the method used to capture them.
 func captureStack(thread: thread_t, maxFrames: Int = 512) -> (frames: [UInt], method: StackUnwindMethod) {
-    // Try fast frame-pointer walking first.
-    var frames = [UInt](repeating: 0, count: maxFrames)
-    var count = 0
-    let success = emb_stack_walk(thread, &frames, maxFrames, &count)
+    var fpFrames = [UInt](repeating: 0, count: maxFrames)
+    var ksFrames = [UInt](repeating: 0, count: maxFrames)
+    let pthread = pthread_from_mach_thread_np(thread)
 
-    if success && count >= minFPFrames {
-        return (Array(frames[0..<count]), .framePointer)
+    guard thread_suspend(thread) == KERN_SUCCESS else {
+        return ([], .framePointerPartial)
     }
 
-    // Fall back to KSCrash's unwinder.
-    if let pthread = pthread_from_mach_thread_np(thread) {
-        var ksFrames = [UInt](repeating: 0, count: maxFrames)
-        let ksCount = captureBacktrace(thread: pthread, addresses: &ksFrames, count: Int32(maxFrames))
+    // Only asyn-safe things allowed between suspend and resume.
 
-        if ksCount > 0 {
-            return (Array(ksFrames[0..<Int(ksCount)]), .kscrash)
-        }
+    var fpCount = 0
+    // Common case: Frame pointer is available, so we can fast-walk the stack.
+    let fpSuccess = emb_stack_walk(thread, &fpFrames, maxFrames, &fpCount)
+
+    // Fall back to KSCrash's unwinder if FP walking didn't produce enough
+    // frames.
+    //
+    // `captureBacktrace` currently does its own suspend/resume of the
+    // target thread, but since Mach uses a suspend count, the nested
+    // suspend/resume just bumps the count to 2 and back to 1,
+    // and then our outer resume below takes it to 0. A future KSCrash
+    // release will expose a variant that skips the extra dance.
+    // (see kstenerud/KSCrash#816).
+    var ksCount: Int32 = 0
+    if (!fpSuccess || fpCount < minFPFrames), let pthread = pthread {
+        ksCount = captureBacktrace(thread: pthread, addresses: &ksFrames, count: Int32(maxFrames))
     }
 
-    // Just return whatever FP walking got us.
-    if success && count > 0 {
-        return (Array(frames[0..<count]), .framePointerPartial)
-    }
+    thread_resume(thread)
 
+    // Safe to allocate again.
+    if fpSuccess && fpCount >= minFPFrames {
+        return (Array(fpFrames[0..<fpCount]), .framePointer)
+    }
+    if ksCount > 0 {
+        return (Array(ksFrames[0..<Int(ksCount)]), .kscrash)
+    }
+    if fpSuccess && fpCount > 0 {
+        return (Array(fpFrames[0..<fpCount]), .framePointerPartial)
+    }
     return ([], .framePointerPartial)
 }
 
--- a/Sources/EmbraceProfiling/StackCapture.swift
+++ b/Sources/EmbraceProfiling/StackCapture.swift
@@ -36,7 +36,7 @@ func captureStack(thread: thread_t, maxFrames: Int = 512) -> (frames: [UInt], me
     let pthread = pthread_from_mach_thread_np(thread)
 
     guard thread_suspend(thread) == KERN_SUCCESS else {
-        return ([], .framePointerPartial)
+        return ([], .failed)
     }
 
     // Only asyn-safe things allowed between suspend and resume.
@@ -71,7 +71,7 @@ func captureStack(thread: thread_t, maxFrames: Int = 512) -> (frames: [UInt], me
     if fpSuccess && fpCount > 0 {
         return (Array(fpFrames[0..<fpCount]), .framePointerPartial)
     }
-    return ([], .framePointerPartial)
+    return ([], .failed)
 }
 
 #endif
--- a/Sources/EmbraceProfiling/StackCapture.swift
+++ b/Sources/EmbraceProfiling/StackCapture.swift
@@ -1,0 +1,58 @@
+//
+//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
+//
+
+#if !os(watchOS)
+
+import Darwin
+import EmbraceProfilingSampler
+
+#if canImport(KSCrash)
+    import KSCrash
+#else
+    import KSCrashRecording
+#endif
+
+/// Minimum number of frames from frame-pointer walking to consider the walk
+/// successful. Below this threshold we fall back to KSCrash's unwinder.
+private let minFPFrames = 3
+
+/// Captures a stack trace from a suspended thread.
+///
+/// Uses fast frame-pointer walking first. If that yields fewer than
+/// ``minFPFrames`` frames, falls back to KSCrash's `captureBacktrace`
+/// unwinder.
+///
+/// Note: KSCrash will support walking a stack without a FP in the next release.
+///
+/// The thread **must** already be suspended before calling this function.
+///
+/// - Parameters:
+///   - thread: The Mach thread port of the suspended thread.
+///   - maxFrames: Maximum number of frames to capture.
+/// - Returns: An array of return addresses, or an empty array on failure.
+func captureStack(thread: thread_t, maxFrames: Int = 512) -> [UInt] {
+    // Try fast frame-pointer walking first.
+    var frames = [UInt](repeating: 0, count: maxFrames)
+    var count = 0
+    let success = emb_stack_walk(thread, &frames, maxFrames, &count)
+
+    if success && count >= minFPFrames {
+        return Array(frames[0..<count])
+    }
+
+    // Fall back to KSCrash's unwinder.
+    if let pthread = pthread_from_mach_thread_np(thread) {
+        var ksFrames = [UInt](repeating: 0, count: maxFrames)
+        let ksCount = captureBacktrace(thread: pthread, addresses: &ksFrames, count: Int32(maxFrames))
+
+        if ksCount > 0 {
+            return Array(ksFrames[0..<Int(ksCount)])
+        }
+    }
+    
+    // Just return whatever FP walking got us.
+    return success && count > 0 ? Array(frames[0..<count]) : []
+}
+
+#endif
--- a/Sources/EmbraceProfiling/StackCapture.swift
+++ b/Sources/EmbraceProfiling/StackCapture.swift
@@ -30,15 +30,15 @@ private let minFPFrames = 3
 /// - Parameters:
 ///   - thread: The Mach thread port of the suspended thread.
 ///   - maxFrames: Maximum number of frames to capture.
-/// - Returns: An array of return addresses, or an empty array on failure.
-func captureStack(thread: thread_t, maxFrames: Int = 512) -> [UInt] {
+/// - Returns: A tuple containing the array of return addresses and the method used to capture them.
+func captureStack(thread: thread_t, maxFrames: Int = 512) -> (frames: [UInt], method: StackUnwindMethod) {
     // Try fast frame-pointer walking first.
     var frames = [UInt](repeating: 0, count: maxFrames)
     var count = 0
     let success = emb_stack_walk(thread, &frames, maxFrames, &count)
 
     if success && count >= minFPFrames {
-        return Array(frames[0..<count])
+        return (Array(frames[0..<count]), .framePointer)
     }
 
     // Fall back to KSCrash's unwinder.
@@ -47,12 +47,16 @@ func captureStack(thread: thread_t, maxFrames: Int = 512) -> [UInt] {
         let ksCount = captureBacktrace(thread: pthread, addresses: &ksFrames, count: Int32(maxFrames))
 
         if ksCount > 0 {
-            return Array(ksFrames[0..<Int(ksCount)])
+            return (Array(ksFrames[0..<Int(ksCount)]), .kscrash)
         }
     }
-    
+
     // Just return whatever FP walking got us.
-    return success && count > 0 ? Array(frames[0..<count]) : []
+    if success && count > 0 {
+        return (Array(frames[0..<count]), .framePointerPartial)
+    }
+
+    return ([], .framePointerPartial)
 }
 
 #endif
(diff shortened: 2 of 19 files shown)


Sources/EmbraceProfilingSampler/source/emb_stack_walker.c
```diff
@@ -6,16 +6,83 @@
...
+#include <ptrauth.h>
+
+// Architecture-specific thread state access.
+#if defined(__arm64__) || defined(__aarch64__)
+  #define EMB_THREAD_STATE_FLAVOR  ARM_THREAD_STATE64
+  #define EMB_THREAD_STATE_COUNT   ARM_THREAD_STATE64_COUNT
+  typedef arm_thread_state64_t     emb_thread_state_t;
+  #define EMB_GET_PC(s) ((void *)arm_thread_state64_get_pc(s))
+  #define EMB_GET_FP(s) ((void *)arm_thread_state64_get_fp(s))
+#elif defined(__x86_64__)
```
fractalwrench, line 19: This is for macOS presumably? @ArielDemarco do we support x86 macs currently?

Sources/EmbraceProfiling/StackCapture.swift
```diff
@@ -0,0 +1,58 @@
...
+/// successful. Below this threshold we fall back to KSCrash's unwinder.
+private let minFPFrames = 3
+
+/// Captures a stack trace from a suspended thread.
+///
+/// Uses fast frame-pointer walking first. If that yields fewer than
+/// ``minFPFrames`` frames, falls back to KSCrash's `captureBacktrace`
+/// unwinder.
+///
+/// Note: KSCrash will support walking a stack without a FP in the next release.
```
fractalwrench: Can this define a specific version rather than 'next release', so that it's accurate forever?

Sources/EmbraceProfiling/StackCapture.swift
```diff
@@ -0,0 +1,58 @@
...
+///
+/// The thread **must** already be suspended before calling this function.
+///
+/// - Parameters:
+///   - thread: The Mach thread port of the suspended thread.
+///   - maxFrames: Maximum number of frames to capture.
+/// - Returns: An array of return addresses, or an empty array on failure.
+func captureStack(thread: thread_t, maxFrames: Int = 512) -> [UInt] {
```
fractalwrench: It feels like it would be beneficial to retain information on what method was used to unwind the stack and send that as part of the telemetry, given the quality of stacktraces might differ

fractalwrench (commented):
I focused my review on `emb_stack_walker.c` and left a few queries inline. I'll ask someone else to review the Swift side of things.

Sources/EmbraceProfilingSampler/source/emb_stack_walker.c
kstenerud: This is for macos. I've added support here because we can't quite rule out x86 builds yet.

Sources/EmbraceProfiling/StackCapture.swift
kstenerud: We're still talking about what version bump to use for the next release of KSCrash. I'll put it in once we've decided.

Sources/EmbraceProfiling/StackCapture.swift
kstenerud: Sounds good. I'll add it to the returned data.

Sources/EmbraceProfiling/StackCapture.swift
```diff
@@ -0,0 +1,62 @@
...
+///
+/// The thread **must** already be suspended before calling this function.
+///
+/// - Parameters:
+///   - thread: The Mach thread port of the suspended thread.
+///   - maxFrames: Maximum number of frames to capture.
+/// - Returns: A tuple containing the array of return addresses and the method used to capture them.
+func captureStack(thread: thread_t, maxFrames: Int = 512) -> (frames: [UInt], method: StackUnwindMethod) {
+    // Try fast frame-pointer walking first.
+    var frames = [UInt](repeating: 0, count: maxFrames)
```
ArielDemarco: Question on async-safety here: as far as I can see from your comment, this function runs after the caller has already suspended the target thread (in our case, the main thread). My understanding is that any heap allocation on this code path can deadlock against libmalloc's internal locks if the suspended thread happened to be inside malloc when we suspended it. The same concern applies to the slice copies in the return paths (you basically create an `Array` directly)

Am I reading this right, or is there something about the (Swift) array allocation here that I'm missing? If it's a real concern, would it make sense to have the caller pass in a preallocated buffer so we never malloc with a thread suspended?

Sources/EmbraceProfiling/StackCapture.swift
```diff
@@ -0,0 +1,62 @@
...
+
+#if canImport(KSCrash)
+    import KSCrash
+#else
+    import KSCrashRecording
+#endif
+
+/// Minimum number of frames from frame-pointer walking to consider the walk
+/// successful. Below this threshold we fall back to KSCrash's unwinder.
+private let minFPFrames = 3
```
ArielDemarco, line 18: Question, mostly to understand: what's the reasoning behind `3` as the threshold? Seems reasonable that we want to validate that the amount of frames are right, but I'm trying to build a mental model of when we expect the `KSCrash` fallback to kick in vs when the FP walker is enough; is 3 covering a specific case (e.g. thread caught mid-prologue before FP is set up?) or is it just a "looks too short to be a real stack" heuristic? Just want to know what to expect when we see samples coming back as `.kscrash` instead of `.framePointer`

Sources/EmbraceProfiling/StackCapture.swift
```diff
@@ -0,0 +1,62 @@
...
+            return (Array(ksFrames[0..<Int(ksCount)]), .kscrash)
+        }
+    }
+
+    // Just return whatever FP walking got us.
+    if success && count > 0 {
+        return (Array(frames[0..<count]), .framePointerPartial)
+    }
+
+    return ([], .framePointerPartial)
```
ArielDemarco: Nit: it may be worth adding an `.error`/`.failure` case to `StackUnwindMethod`, or making `method` optional to better represent failure scenarios

Sources/EmbraceProfiling/StackCapture.swift
kstenerud: This is actually partway through, waiting on the next KSCrash. Sorry I didn't flag it!

Originally it would allocate, suspend, backtrace, resume. But with the fallback there's a wrinkle: `captureBacktraceFromOtherThread` (called by `ksbt_captureBacktraceFromMachThread`) does its own suspend/resume dance, which mucks things up. I've got `ksbt_captureBacktraceFromSuspendedMachThread` going into KSCrash, after which the process will be:

* allocate space for stack trace
* suspend thread
* try fast path backtrace
* fallback to KSCrash backtrace (slower but handles all edge cases)
* resume thread

Actually, multiple suspends doesn't hurt anything besides being slower than not doing it... I could just put the existing `ksbt_captureBacktraceFromMachThread` in there for now, and then use the lighter one later, rather than having a future dependency at the start.

Sources/EmbraceProfiling/StackCapture.swift
kstenerud: It's a heuristic to detect failed stack walks. A trace under 3 frames is not going to capture anything interesting (the best you could get is `_pthread_start`, some syscall that waits, `mach_msg_trap`). A failed capture could get the program counter but nothing more, or even in some cases the PC + 1 more (bogus) frame and then fail. The chances of PC+2 bogus frames being within range is already vanishingly small, and +3 frames effectively impossible, so I picked 3 because it's the biggest trace size that is also guaranteed useless on any random thread (for our purposes).

kstenerud:
Updated to always assume a running thread prior to getting a stack trace, which will match with our use case anyway.

The next KSCrash will have a new backtrace method exposed that [assumes a stopped thread](https://github.com/kstenerud/KSCrash/pull/816), which we can swap in once the next release goes out. No need to gate on this.

fnewberg:
@kstenerud you wrote in the description

> but these won't pass until KSCrash is updated.

and in the most recent comment you wrote

> The next KSCrash will have a new backtrace method exposed that https://github.com/kstenerud/KSCrash/pull/816, which we can swap in once the next release goes out. No need to gate on this.

can you clarify if we need to update KSCrash or not?

Sources/EmbraceProfiling/ProfilingSample.swift
```diff
@@ -2,16 +2,32 @@
 //  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
 //
 
+/// How a stack trace was captured.
+public enum StackUnwindMethod {
+    /// Frame-pointer based stack walking (fast, works when FP is available)
+    case framePointer
+    /// KSCrash's DWARF-based unwinder (slower, more robust fallback)
```
fnewberg, line 9: @kstenerud do you have any measurement of what fast vs slow is? 2x? order of magnitude?

Sources/EmbraceProfiling/StackCapture.swift
```diff
@@ -0,0 +1,77 @@
+//
+//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
```
fnewberg: nitpick: we should update all these dates to be 2026

Sources/EmbraceProfiling/StackCapture.swift
```diff
@@ -0,0 +1,77 @@
...
+    // suspend/resume just bumps the count to 2 and back to 1,
+    // and then our outer resume below takes it to 0. A future KSCrash
+    // release will expose a variant that skips the extra dance.
+    // (see kstenerud/KSCrash#816).
+    var ksCount: Int32 = 0
+    if (!fpSuccess || fpCount < minFPFrames), let pthread = pthread {
+        ksCount = captureBacktrace(thread: pthread, addresses: &ksFrames, count: Int32(maxFrames))
+    }
+
+    thread_resume(thread)
```
fnewberg, line 76: is there any plausible scenario where this would not return `KERN_SUCCESS`? even if it returned an error, what practical options do we have if the main thread failed to resume?  write something to a file that we look at next launch?

Sources/EmbraceProfilingSampler/source/emb_stack_walker.c
```diff
@@ -6,16 +6,83 @@
...
+    kern_return_t kr = thread_get_state(thread,
+                                        EMB_THREAD_STATE_FLAVOR,
+                                        (thread_state_t)&state,
+                                        &state_count);
+    if (kr != KERN_SUCCESS) {
+        return false;
+    }
+
+    // Determine the target thread's stack bounds for safe direct dereference.
+    pthread_t target_pthread = pthread_from_mach_thread_np(thread);
```
fnewberg: [claude]   Deadlock risk: `pthread_from_mach_thread_np` called while target thread is suspended. The Swift wrapper resolves the pthread before suspend, but the C walker re-resolves it inside the suspended window. If the target held libpthread's list lock at suspension, the sampler deadlocks.

 In `captureStack`, `pthread_from_mach_thread_np(thread)` is done before suspend (`Sources/EmbraceProfiling/StackCapture.swift:36`), which is the safe pattern.

Fix: pass `stack_top`/`stack_size` as parameters from Swift.

Sources/EmbraceProfilingSampler/source/emb_stack_walker.c
```diff
@@ -6,16 +6,83 @@
...
+            break;
+        }
+
+        frames_out[count++] = (uintptr_t)ret_addr;
+
+        if (next_fp == NULL || (uintptr_t)next_fp <= (uintptr_t)fp) {
+            break;
+        }
+        fp = next_fp;
+    }
```
fnewberg, line 80: [claude]  Bounds check is `fp < stack_top`, but the loop reads 16 bytes (`record[0]` and `record[1]`). At `fp == stack_top - 8`,  `record[1]` reads past the mapped stack region.

-  `(char *)fp + 2 * sizeof(void *) <= (char *)stack_top` — guarantees both `record[0]` and `record[1]` are inside `[stack_bottom, stack_top)`. The original `fp < stack_top` allowed `fp == stack_top - 8`, which read 8 bytes past the stack. On 64-bit Apple platforms `sizeof(void *) == 8`, so this resolves to `fp + 16 <= stack_top`.
  - Casts to `(char *)` — pointer arithmetic on `void *` is a GNU extension; casting to `char *` is the portable form and is what Clang prefers under `-Wpedantic`.
  - `fp_align_mask = 2 * sizeof(void *) - 1` — `0xF` on 64-bit. Frame records are 16-byte-aligned per arm64 AAPCS and x86_64 SysV. A non-aligned FP almost always means we walked into garbage; bailing is safer than dereferencing.
  - Alignment + bounds applied to `next_fp` — if `next_fp` looks bad, the original code would still adopt it and only fail on the next iteration's bounds check. Catching it here gives a slightly cleaner exit and avoids one wasted iteration. (Strictly the next-iteration bounds check is sufficient, but this makes the invariant explicit.)

```suggestion
    // Each frame record is: [saved_fp, return_address] — 2 * sizeof(void *) bytes.
    // We require the entire record to lie within the thread's stack, and that fp
    // is 16-byte aligned (arm64 AAPCS / x86_64 SysV). A misaligned or out-of-range
    // fp indicates a corrupt chain (or a thread caught mid-prologue) and we stop
    // rather than risk a fault or a torn read.    
   const uintptr_t stack_lo = (uintptr_t)stack_bottom;
   const uintptr_t stack_hi = (uintptr_t)stack_top; 
   const uintptr_t record_size = 2 * sizeof(void *);
   const uintptr_t fp_align_mask = record_size - 1;  // 0xF on 64-bit
   uintptr_t fp_addr = (uintptr_t)fp;
   
   while (count < max_frames
         && fp_addr >= stack_lo
         && fp_addr <= stack_hi - record_size
         && (fp_addr & fp_align_mask) == 0) {
      void **record = (void **)fp_addr;
      void *next_fp = record[0];
      void *ret_addr = ptrauth_strip(record[1], ptrauth_key_return_address);

      if (ret_addr == NULL) {
          break;
      }

      frames_out[count++] = (uintptr_t)ret_addr;

      // Stop on null (top of chain), non-monotonic fp (cycle / corrupt chain),
      // or a misaligned saved fp.
      uintptr_t next_addr = (uintptr_t)next_fp;
      if (next_fp == NULL
          || next_addr <= fp_addr
          || (next_addr & fp_align_mask) != 0) {
          break;
      }
      fp_addr = next_addr;
  }
```

Sources/EmbraceProfilingSampler/source/emb_stack_walker.c
```diff
@@ -6,16 +6,83 @@
...
+  #define EMB_GET_FP(s) ((void *)arm_thread_state64_get_fp(s))
+#elif defined(__x86_64__)
+  #define EMB_THREAD_STATE_FLAVOR  x86_THREAD_STATE64
+  #define EMB_THREAD_STATE_COUNT   x86_THREAD_STATE64_COUNT
+  typedef x86_thread_state64_t     emb_thread_state_t;
+  #define EMB_GET_PC(s) ((void *)(s).__rip)
+  #define EMB_GET_FP(s) ((void *)(s).__rbp)
+#endif
+
+bool emb_stack_walk(thread_t thread,
```
fnewberg, line 27: rather than return `bool`, which limits what we can tell about the failure case, would it be worthwhile to add something like this?

```
typedef enum {
      EMB_STACK_OK              = 0,
      EMB_STACK_BAD_ARGS        = 1,
      EMB_STACK_THREAD_STATE    = 2,  // attach kr separately
      EMB_STACK_PTHREAD_LOOKUP  = 3,
      EMB_STACK_EMPTY           = 4,  // succeeded but produced no frames
  } emb_stack_status_t;
```

and have `emb_stack_walk` return a `emb_stack_status_t` value? we could then propagate the failure up and ultimate log it.

[claude] alternatively split up the work done in swift/C differently

Keep the bool, but split the function into two parts in Swift, doing each Mach call from Swift so the kern_return_t is visible at the call site:

```
  // In StackCapture.swift, before calling into C:
  var state = arm_thread_state64_t() // or x86 equivalent
  var stateCount = mach_msg_type_number_t(ARM_THREAD_STATE64_COUNT)
  let kr = withUnsafeMutablePointer(to: &state) {
      $0.withMemoryRebound(to: natural_t.self, capacity: Int(stateCount)) {
          thread_get_state(thread, ARM_THREAD_STATE64, $0, &stateCount)
      }
  }
  guard kr == KERN_SUCCESS else {
      Embrace.logger.debug("thread_get_state failed: kr=\(kr)")
      metrics.threadStateFailures.increment(reason: machErrorString(kr))
      thread_resume(thread)
      return ([], .failed)
  }

  // Resolve pthread separately, also outside the C function.
  guard let target = pthread_from_mach_thread_np(thread) else {
      Embrace.logger.debug("pthread_from_mach_thread_np returned NULL")
      metrics.staleThreadPort.increment()
      thread_resume(thread)
      return ([], .failed)
  }

  // Now call into C with everything pre-resolved. C function becomes
  // pure stack walking with no Mach/pthread calls — and nothing to fail with.
  emb_stack_walk_unchecked(state, stack_top, stack_size, &frames, &count)
```

Sources/EmbraceProfilingSampler/source/emb_stack_walker.c
```diff
@@ -6,16 +6,83 @@
...
+    void *fp = EMB_GET_FP(state);
+    size_t count = 0;
+
+    if (pc != NULL) {
+        frames_out[count++] = (uintptr_t)pc;
+    }
+
+    // Walk the frame pointer chain.
+    // Each frame record is: [saved_fp, return_address]
+    while (count < max_frames && fp >= stack_bottom && fp < stack_top) {
```
fnewberg: should we consider adding some sort of truncation marker if we exit due to hitting `max_frames`. in our android sampling implementation we added this.

Sources/EmbraceProfiling/ProfilingSample.swift
```diff
@@ -2,16 +2,32 @@
...
+    case kscrash
+    /// Partial frame-pointer walk (below minFPFrames threshold, returned as last resort)
```
fnewberg:
```suggestion
    /// Partial frame-pointer walk (below minFPFrames threshold, returned as last resort if kscrash returned nothing)
```

Tests/EmbraceProfilingTests/FallbackBenchmarks.swift
```diff
@@ -0,0 +1,169 @@
...
+#if canImport(KSCrash)
+    import KSCrash
+#else
+    import KSCrashRecording
+#endif
+
+/// Benchmarks comparing FP-based stack walking vs KSCrash fallback on stacks
+/// compiled with and without frame pointers.
+///
+/// `-fomit-frame-pointer` takes effect on both arm64 and x86_64 with modern
```
fnewberg: this contradicts what is stated [here](https://github.com/embrace-io/embrace-apple-sdk/pull/538/changes#diff-5ddde3e0dd7362f53c03f5170c29190aeb5c433c92dd76ac21a3e5340310b5ccR24)

Tests/EmbraceProfilingSamplerTests/StackWalkerTests.swift
```diff
@@ -0,0 +1,82 @@
...
+        let resumeResult = thread_resume(machThread)
+        XCTAssertEqual(resumeResult, KERN_SUCCESS, "thread_resume should succeed")
+
+        // Verify the walk succeeded and produced frames.
+        XCTAssertTrue(walkResult, "emb_stack_walk should succeed")
+        XCTAssertGreaterThan(count, 0, "Should capture at least one frame")
+
+        // Every reported frame address should be non-zero.
+        for i in 0..<count {
+            XCTAssertNotEqual(frames[i], 0, "Frame \(i) should be a non-zero address")
```
fnewberg: any value in checking for a realistic memory address?

Tests/EmbraceProfilingSamplerTests/StackWalkerTests.swift
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    func test_stackWalk_failsWithInvalidThread() {
```
fnewberg: any value in adding a test targeting a terminated thread?

Tests/EmbraceProfilingSamplerTests/StackWalkerTests.swift
```diff
@@ -0,0 +1,82 @@
...
+
+#if !os(watchOS)
+
+import Darwin
+import EmbraceProfilingSampler
+import XCTest
+
+final class StackWalkerTests: XCTestCase {
+
+    func test_stackWalk_capturesFramesFromSuspendedThread() throws {
```
fnewberg: consider adding a similar test where we exit based on hitting the frame limit

Sources/EmbraceProfiling/StackCapture.swift
```diff
@@ -0,0 +1,77 @@
...
+func captureStack(thread: thread_t, maxFrames: Int = 512) -> (frames: [UInt], method: StackUnwindMethod) {
+    var fpFrames = [UInt](repeating: 0, count: maxFrames)
+    var ksFrames = [UInt](repeating: 0, count: maxFrames)
+    let pthread = pthread_from_mach_thread_np(thread)
+
+    guard thread_suspend(thread) == KERN_SUCCESS else {
+        return ([], .failed)
+    }
+
+    // Only asyn-safe things allowed between suspend and resume.
```
fnewberg:
```suggestion
    // Only async-safe things allowed between suspend and resume.
```

Tests/EmbraceProfilingTestSupport/source/emb_test_thread.c
```diff
@@ -0,0 +1,70 @@
...
+#include "emb_test_thread.h"
+
+#if !TARGET_OS_WATCH
+
+#include <pthread.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <mach/mach.h>
+
+struct emb_test_thread {
```
fnewberg, line 13: in the spirit of avoiding tests that can be flaky -- we have some work to do there in general -- i prefer to look for ways to avoid using sleeps in tests. claude suggested these changes to allow us to wait for a condition to be met when spinning up the thread rather than relying on sleeps

```
  A pthread_cond_t lets the target thread itself signal "I have reached the bottom of the recursion and entered the spin loop" — which is the actual condition the caller cares
  about. There's no guessing about scheduler timing; the caller blocks until the signal, no matter how long it takes.

  struct emb_test_thread {
      pthread_t pthread;
      thread_t mach_port;
      volatile int running;
      size_t stack_depth;

      pthread_mutex_t ready_mutex;
      pthread_cond_t  ready_cond;
      int             ready;     // 0 until target hits the spin loop
  };

  __attribute__((noinline))
  static void recurse(struct emb_test_thread *t, size_t depth) {
      volatile size_t anchor = depth;
      if (depth > 0) {
          recurse(t, depth - 1);
      } else {
          // We've reached target depth. Signal the spawner before spinning.
          pthread_mutex_lock(&t->ready_mutex);
          t->ready = 1;
          pthread_cond_signal(&t->ready_cond);
          pthread_mutex_unlock(&t->ready_mutex);

          while (t->running) { }
      }
      (void)anchor;
  }

  static void *thread_entry(void *arg) {
      struct emb_test_thread *t = (struct emb_test_thread *)arg;
      recurse(t, t->stack_depth);
      return NULL;
  }

  emb_test_thread_t emb_test_thread_create(size_t stack_depth) {
      struct emb_test_thread *t = calloc(1, sizeof(*t));
      if (t == NULL) return NULL;

      t->running = 1;
      t->stack_depth = stack_depth;
      pthread_mutex_init(&t->ready_mutex, NULL);
      pthread_cond_init(&t->ready_cond, NULL);

      if (pthread_create(&t->pthread, NULL, thread_entry, t) != 0) {
          pthread_cond_destroy(&t->ready_cond);
          pthread_mutex_destroy(&t->ready_mutex);
          free(t);
          return NULL;
      }

      // Block until the target thread reaches its spin loop.
      pthread_mutex_lock(&t->ready_mutex);
      while (!t->ready) {
          pthread_cond_wait(&t->ready_cond, &t->ready_mutex);
      }
      pthread_mutex_unlock(&t->ready_mutex);

      t->mach_port = pthread_mach_thread_np(t->pthread);
      return t;
  }

  void emb_test_thread_destroy(emb_test_thread_t t) {
      t->running = 0;
      pthread_join(t->pthread, NULL);
      pthread_cond_destroy(&t->ready_cond);
      pthread_mutex_destroy(&t->ready_mutex);
      free(t);
  }
```

Sources/EmbraceProfiling/ProfilingSample.swift
```diff
@@ -2,16 +2,32 @@
 //  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
 //
 
+/// How a stack trace was captured.
+public enum StackUnwindMethod {
```
fnewberg: [claude]


All four conformances are compiler-synthesized for the current shape (no associated values), so this costs nothing at the declaration site:

- `Equatable` — lets tests use `XCTAssertEqual(sample.unwindMethod, .framePointer)` instead of pattern-matching switches. Matters for pinning the four-way fallback decision in `StackCapture.swift`.
- `Hashable` — enables `[StackUnwindMethod: Int]` for telemetry aggregation (e.g. counting samples per method).
- `CaseIterable` — makes exhaustive tests and pre-seeded histograms robust to new cases. If we add `.framePointerTruncated` later, existing call sites thatiterate `allCases` pick it up automatically.
- `Sendable` — documents intent and prevents strict-concurrency regressions when `ProfilingSample` crosses actor / `Task` boundaries. Also acts as a tripwire: if someone later adds a non-`Sendable` associated value, the compiler flags it.

```suggestion
public enum StackUnwindMethod: Equatable, Hashable, CaseIterable, Sendable {
```

Sources/EmbraceProfiling/ProfilingSample.swift
```diff
@@ -2,16 +2,32 @@
...
+    /// KSCrash's DWARF-based unwinder (slower, more robust fallback)
+    case kscrash
+    /// Partial frame-pointer walk (below minFPFrames threshold, returned as last resort)
+    case framePointerPartial
+    /// Stack capture failed entirely — no frames were obtained.
+    case failed
+}
+
 /// A single profiling sample containing a stack trace captured at a point in time.
 public struct ProfilingSample {
```
fnewberg:
```suggestion
public struct ProfilingSample: Equatable, Hashable, Sendable {
```

Sources/EmbraceProfiling/StackCapture.swift
```diff
@@ -0,0 +1,77 @@
...
+/// frames, falls back to KSCrash's `captureBacktrace` unwinder.
+///
+/// Note: KSCrash will support walking a stack without a FP in the next release.
+///
+/// - Parameters:
+///   - thread: The Mach thread port of the thread to capture.
+///   - maxFrames: Maximum number of frames to capture.
+/// - Returns: A tuple containing the array of return addresses and the method used to capture them.
+func captureStack(thread: thread_t, maxFrames: Int = 512) -> (frames: [UInt], method: StackUnwindMethod) {
+    var fpFrames = [UInt](repeating: 0, count: maxFrames)
```
fnewberg, line 34: should we consider using a pool to avoid lots of allocations if this is running for extended periods of time?

Sources/EmbraceProfiling/StackCapture.swift
```diff
@@ -0,0 +1,77 @@
...
+    var ksFrames = [UInt](repeating: 0, count: maxFrames)
```
fnewberg, line 35: any way to make the additional allocation lazy if we think the fallback is rare? then again, if we go with a pool approach i think this becomes a moot point

fnewberg (changes requested):

Sources/EmbraceProfiling/StackCapture.swift
kstenerud: This gets replaced with a C implementation in a later PR.

Sources/EmbraceProfiling/ProfilingSample.swift
kstenerud: I don't have any measurements yet, but we've just dropped the beta of KSCrash so we can try plugging in and measuring. If the performance drop is too nasty we can just opt to not hook it in. It's all configurable.

Sources/EmbraceProfiling/StackCapture.swift
kstenerud: It would fail if the runtime gets into a bad state. This class actually gets replaced with a C implementation in a later PR because the C Swift bridge tax was just too high.

kstenerud:
I've backported some of the fixes that were in later PRs back to this one.

Sources/EmbraceProfilingSampler/source/emb_stack_walker.c
kstenerud: This should be fixed with the backported stack walker.

Sources/EmbraceProfiling/StackCapture.swift
kstenerud: I have actually moved to a pooled approach in a later PR. The sampler holds a buffer that it passes in on every sampling run, so the entire capture path never allocates. StackCapture.swift goes away due to the C-Swift friction, so this file will no longer exist.

Sources/EmbraceProfiling/StackCapture.swift
kstenerud: See above. This class is going away, so all is well :)

Sources/EmbraceProfilingSampler/source/emb_stack_walker.c
kstenerud: The only two cases where `emb_stack_walk` would fail are: invalid arguments, or a failure in `thread_get_state`.

If we pass in invalid arguments, we'll get failures on every call, and zero samples. This can be caught by tests.

`thread_get_state` will only fail on:

* invalid arguments (in which case every sample attempt will fail, every time, including in the tests).
* The app is shutting down, and so the main thread is terminating. We don't want to log this or try to do anything about it.
* The runtime or operating system is in a catastrophically bad state, and trying to do anything would likely make things worse. Logging wouldn't help us here because there's nothing we could change in the code to defend against this.

Sources/EmbraceProfilingSampler/source/emb_stack_walker.c
kstenerud: We could do this, but realistically the main thread isn't going to exceed 100 frames, let alone the 500 configured by default as a max (and you could increase that to 1024 max if you like).

fnewberg: based on the data we have, 500 feels safe

fnewberg (approved):
there are only a handful of nitpicky things left in my comments. i'll approve it, but it would be good to commit some of the suggestions before you merge it

kstenerud:
> can you clarify if we need to update KSCrash or not?

It's not required that you update KSCrash. You can leave it as-is and this will still work.

Once you do update it, you can benefit from the DWARF stack unwinding that the newer KSCrash supports. Until then, it will use the frame pointer stack walker, which will work in 99% of cases (same as your current level of support provides for crash time stack capture using the existing version of KSCrash).

Tests/EmbraceProfilingTests/FallbackBenchmarks.swift
kstenerud: I've updated the comment with a list of things that can smash the FP.

Tests/EmbraceProfilingTestSupport/include/module.modulemap
```diff
@@ -0,0 +1,9 @@
+//
+//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
```
fnewberg:
```suggestion
//  Copyright © 2026 Embrace Mobile, Inc. All rights reserved.
```

The rest of the branch is on 2026 — this file was missed by the prior copyright-bump passes.

Tests/EmbraceProfilingTestSupportNoFP/include/module.modulemap
```diff
@@ -0,0 +1,9 @@
+//
+//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
```
fnewberg:
```suggestion
//  Copyright © 2026 Embrace Mobile, Inc. All rights reserved.
```

The rest of the branch is on 2026 — this file was missed by the prior copyright-bump passes.

Sources/EmbraceProfilingSampler/include/emb_stack_walker.h
```diff
@@ -20,11 +46,23 @@ extern "C" {
...
-                   uint64_t *frames_out,
-                   uint32_t max_frames,
-                   uint32_t *count_out);
+/// Async-safe: only calls `thread_get_state` (a Mach trap) and walks memory
+/// within the provided stack bounds. Callers must resolve stack bounds before
+/// the target thread is suspended (pthread APIs are not async-safe).
+///
+/// @param thread       Mach thread port (must be suspended).
+/// @param stack_bottom Lowest valid stack address.
+/// @param stack_top    Highest valid stack address (pthread_get_stackaddr_np).
```
fnewberg: `emb_stack_walker.c:64` uses strict `fp < stack_top`, so `stack_top` is the first address *past* the top of the stack — not the highest valid address. Anyone passing `(uintptr_t)stack_top - 1` thinking they were doing the safe thing would actually be passing the highest valid byte and the loop would stop one frame too early.

```suggestion
/// @param stack_top    First address past the top of the stack (treated exclusively); typically the value returned by `pthread_get_stackaddr_np`.
```

PR #555, opened 2026-03-30
Add lock-free ring buffer implementation for profiling samples

kstenerud:
Single-writer, multiple-reader ring buffer using VM double-mapping for transparent wrapping.

The write path is async-safe (no locks, no heap allocation) to avoid deadlocks when threads are stopped during profiling.

Includes comprehensive test coverage for concurrency, edge cases, lifecycle, and write path scenarios.

## Checklist

### DIFF
--- a/Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
+++ b/Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
@@ -1,5 +1,5 @@
 //
-//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
+//  Copyright © 2026 Embrace Mobile, Inc. All rights reserved.
 //
 
 /// Lock-free ring buffer for holding stack trace samples.
--- a/Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
+++ b/Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
@@ -11,6 +11,7 @@
 
 #include <stdbool.h>
 #include <stdatomic.h>
+#include <stddef.h>
 #include <stdint.h>
 
 #ifdef __cplusplus
--- a/Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
+++ b/Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
@@ -1,0 +1,121 @@
+//
+//  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
+//
+
+#ifndef emb_ring_buffer_h
+#define emb_ring_buffer_h
+
+#include <TargetConditionals.h>
+
+#if !TARGET_OS_WATCH
+
+#include <stdbool.h>
+#include <stdatomic.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// Ring buffer for profiling stack trace samples.
+///
+/// Single-writer, multiple-reader. The write path is async-safe (no heap
+/// allocation, no locks). Readers are not async-safe and may retry.
+///
+/// Uses a VM double-mapping trick: the same physical pages are mapped at two
+/// contiguous virtual addresses. A write that straddles the end of the buffer
+/// wraps to the beginning transparently, with no special-case code.
+///
+/// DESIGN ASSUMPTIONS:
+/// - There will only ever be one writer, and multiple concurrent readers.
+/// - The writer will not write often enough to cause more than one torn read for any given reader.
+///   Officially, the minimum is 1ms between writes, even though it could likely handle tighter cadence.
+typedef struct {
+    uint8_t *data;               // Double-mapped virtual region (2 × capacity).
+    size_t capacity;             // Usable buffer size in bytes (page-aligned).
+    uint64_t next_seq;           // Writer-local seqlock counter (single-writer).
+    _Atomic uint64_t write_pos;  // Monotonically increasing write position.
+    _Atomic uint64_t oldest_pos; // Position of the oldest surviving record.
+} emb_ring_buffer_t;
+
+/// Create a ring buffer.
+///
+/// This function is not async-safe.
+///
+/// @param capacity_bytes Minimum usable capacity; rounded up to a page boundary.
+/// @return A newly allocated buffer, or NULL on failure.
+emb_ring_buffer_t *emb_ring_buffer_create(size_t capacity_bytes);
+
+/// Destroy a ring buffer and release all VM and heap resources.
+///
+/// This function is not async-safe.
+///
+/// Safe to call with NULL.
+void emb_ring_buffer_destroy(emb_ring_buffer_t *buf);
+
+/// Write a record to the ring buffer. Evicts old records if necessary.
+///
+/// Copies the frame data into the buffer and makes it visible to readers.
+///
+/// This function is async-safe.
+///
+/// Invariants:
+/// - Only one concurrent writer
+/// - Minimum 1ms between writes
+/// - Less than 1 million stack trace frames
+///
+/// @param buf The ring buffer.
+/// @param timestamp_ns Monotonic timestamp (nanoseconds).
+/// @param frames Array of frame addresses to copy.
+/// @param frame_count Number of frames in the array. Must not exceed UINT32_MAX.
+/// @return true if the write succeeded, false if buf or frames is NULL.
+bool emb_ring_buffer_write(emb_ring_buffer_t *buf,
+                           uint64_t timestamp_ns,
+                           const uintptr_t *frames,
+                           size_t frame_count);
+
+/// Consumer-facing record (returned by read functions).
+typedef struct {
+    uint64_t timestamp_ns;   // Monotonic timestamp (nanoseconds).
+    size_t frame_count;      // Number of frames in this sample.
+    const uintptr_t *frames; // Frame addresses (points into allocation).
+} emb_ring_record_t;
+
+/// Result from a read operation.
+typedef struct {
+    emb_ring_record_t *records; // Array of records.
+    size_t count;               // Number of valid records.
+} emb_ring_read_result_t;
+
+/// Read records from the ring buffer within a time range.
+///
+/// Returns records with timestamps in the range [start_ns, end_ns].
+/// Use end_ns = UINT64_MAX to read up to the current write position.
+///
+/// Returns a single allocation containing the record array and frame data.
+/// Free with emb_ring_read_result_free().
+///
+/// This function is not async-safe.
+///
+/// @param buf The ring buffer.
+/// @param start_ns Start timestamp (inclusive, nanoseconds).
+/// @param end_ns End timestamp (inclusive, nanoseconds). Use UINT64_MAX for "up to now".
+/// @return Result containing matching records. count=0 if empty, out of range, or buf is NULL.
+emb_ring_read_result_t emb_ring_buffer_read_range(const emb_ring_buffer_t *buf,
+                                                   uint64_t start_ns,
+                                                   uint64_t end_ns);
+
+/// Free a buffer read result.
+///
+/// This function is not async-safe.
+///
+/// Safe to call with a zeroed result (records=NULL).
+void emb_ring_read_result_free(emb_ring_read_result_t *result);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* !TARGET_OS_WATCH */
+
+#endif /* emb_ring_buffer_h */
--- a/Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
+++ b/Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
@@ -2,6 +2,30 @@
 //  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
 //
 
+/// Lock-free ring buffer for holding stack trace samples.
+///
+/// Single-writer, multiple-reader. All functions are async-safe except for
+/// create and destroy.
+///
+/// In order to keep memory churn to a minimum, a caller must pass in a buffer
+/// to receive the data when reading. The caller (in our case, ProfilingEngine)
+/// maintains its own persistent read buffer of the same size as this ring buffer,
+/// and then in turn provides smaller allocations to its callers, with its own
+/// concurrency guarantees.
+///
+/// CAPACITY GUIDANCE:
+/// The buffer is designed for records much smaller than the total capacity.
+/// Individual records (header + frames) should be less than 25% of the
+/// buffer size to guarantee safe reads with a write interval down to 1ms.
+/// Concurrent reads may see torn data if a record size approaches the
+/// buffer capacity and/or frequency is too high, because the writer could
+/// overwrite the record mid-read.
+///
+/// DESIGN ASSUMPTIONS:
+/// - There will only ever be one writer, and multiple concurrent readers.
+/// - The writer will not write often enough to cause more than one torn read for any given reader.
+///   Officially, the minimum is 1ms between writes, even though it could likely handle tighter cadence.
+
 #ifndef emb_ring_buffer_h
 #define emb_ring_buffer_h
 
@@ -18,30 +42,56 @@
 extern "C" {
 #endif
 
-/// Ring buffer for profiling stack trace samples.
-///
-/// Single-writer, multiple-reader. The write path is async-safe (no heap
-/// allocation, no locks). Readers are not async-safe and may retry.
-///
-/// Uses a VM double-mapping trick: the same physical pages are mapped at two
-/// contiguous virtual addresses. A write that straddles the end of the buffer
-/// wraps to the beginning transparently, with no special-case code.
-///
-/// DESIGN ASSUMPTIONS:
-/// - There will only ever be one writer, and multiple concurrent readers.
-/// - The writer will not write often enough to cause more than one torn read for any given reader.
-///   Officially, the minimum is 1ms between writes, even though it could likely handle tighter cadence.
 typedef struct {
     uint8_t *data;               // Double-mapped virtual region (2 × capacity).
     size_t capacity;             // Usable buffer size in bytes (page-aligned).
     uint64_t next_seq;           // Writer-local seqlock counter (single-writer).
     _Atomic uint64_t write_pos;  // Monotonically increasing write position.
     _Atomic uint64_t oldest_pos; // Position of the oldest surviving record.
+    _Atomic uint32_t active_readers; // Number of concurrent read_range calls (for reset safety).
+    _Atomic bool resetting;      // True while reset is in progress.
+                                 // Uses Dekker-style mutual exclusion with active_readers:
+                                 // both sides use seq_cst to prevent ARM64 store-buffer
+                                 // reordering. See emb_ring_buffer_reset and read_range.
 } emb_ring_buffer_t;
+
+/// Public facing view of the record header stored in the ring buffer.
+/// We do this to simplify reading the records. It's expected that the caller would
+/// then convert the retrieved records to a more suitable user-facing format.
+///
+/// Each record consists of this header followed by `frame_count` values of
+/// type `uintptr_t` (the captured frame addresses). Use `emb_ring_record_size`
+/// to compute the total byte size of a record.
+///
+/// The `seq` field is an internal seqlock value used for torn-read detection,
+/// and callers should ignore it.
+typedef struct {
+    uint32_t seq;          // Internal seqlock value (ignore).
+    uint32_t frame_count;  // Number of frames following this header.
+    uint64_t timestamp_ns; // Monotonic timestamp (nanoseconds).
+} emb_ring_record_header_t;
+
+// The record layout assumes 64-bit pointers (uintptr_t == 8 bytes).
+// The Swift layer reads frame data as UInt (also 8 bytes on 64-bit).
+// Catch any hypothetical 32-bit platform at compile time.
+_Static_assert(sizeof(uintptr_t) == 8,
+               "Ring buffer record layout requires 64-bit pointers");
+
+/// Compute the total byte size of a record with the given frame count.
+static inline size_t emb_ring_record_size(uint32_t frame_count) {
+    return sizeof(emb_ring_record_header_t) + (size_t)frame_count * sizeof(uintptr_t);
+}
+
+/// Result from a read operation.
+typedef struct {
+    size_t records_offset; // Offset in bytes to the first matching record.
+    size_t record_count;   // Number of matching records written to the output buffer.
+    size_t total_bytes;    // Total size in bytes of the matching record set.
+} emb_ring_read_result_t;
 
 /// Create a ring buffer.
 ///
-/// This function is not async-safe.
+/// This function is NOT async-safe.
 ///
 /// @param capacity_bytes Minimum usable capacity; rounded up to a page boundary.
 /// @return A newly allocated buffer, or NULL on failure.
@@ -49,10 +99,30 @@ emb_ring_buffer_t *emb_ring_buffer_create(size_t capacity_bytes);
 
 /// Destroy a ring buffer and release all VM and heap resources.
 ///
-/// This function is not async-safe.
+/// This function is NOT async-safe.
 ///
 /// Safe to call with NULL.
 void emb_ring_buffer_destroy(emb_ring_buffer_t *buf);
+
+/// Reset a ring buffer, clearing all data and resetting positions.
+///
+/// The caller must ensure no concurrent writer is active.
+///
+/// Returns false if buf is NULL or if readers are active.
+///
+/// @return true if the buffer was successfully reset, false otherwise.
+bool emb_ring_buffer_reset(emb_ring_buffer_t *buf);
+
+/// Return the usable capacity of the ring buffer in bytes.
+///
+/// This is the page-aligned capacity, which may be larger than the value
+/// passed to emb_ring_buffer_create.
+///
+/// This function is async-safe.
+///
+/// @param buf The ring buffer. Returns 0 if NULL.
+/// @return Capacity in bytes, or 0 if buf is NULL.
+size_t emb_ring_buffer_capacity(const emb_ring_buffer_t *buf);
 
 /// Write a record to the ring buffer. Evicts old records if necessary.
 ///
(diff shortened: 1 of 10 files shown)


Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,371 @@
...
+                      (emb_vm_size_t)buf->capacity * 2);
+    free(buf);
+}
+
+bool emb_ring_buffer_write(emb_ring_buffer_t *buf,
+                           uint64_t timestamp_ns,
+                           const uintptr_t *frames,
+                           size_t frame_count)
+{
+    if (buf == NULL || frames == NULL || frame_count > UINT32_MAX) {
```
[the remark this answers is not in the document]
kstenerud: If we allowed this, the writer could fill the buffer with zero-length stack reads, which we don't want. We're only interested in times when something produced useful information.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,371 @@
...
+    while (pos < data_byte_count) {
+        const emb_ring_record_header_t *header = (emb_ring_record_header_t *)(data_region + pos);
+        const uint64_t seq = header->seq;
+        const size_t frame_count = header->frame_count;
+
+        if ((seq & 1) != 0) {
+            // An odd seq means that a write was in progress when memcpy reached this header.
+            // Any data after this point is unusable.
+            break;
+        }
```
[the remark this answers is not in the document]
kstenerud: The invariant "The writer will not write often enough to cause more than one torn read for any given reader." (1ms) prevents this from realistically happening. A `memcpy` takes microseconds, which is why such a large grace period (1ms) was chosen as a lower bound. The actual use case is a period of 100ms.

Checking for multiple tears during a read would complicate the code for a situation that won't happen in practice.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+    size_t page = vm_page_size;
+    return (size + page - 1) & ~(page - 1);
+}
+
+emb_ring_buffer_t *emb_ring_buffer_create(size_t capacity_bytes)
+{
+    const size_t capacity = round_size_up_to_page_boundary(capacity_bytes);
+    if (capacity == 0) {
+        return NULL;
+    }
```
ArielDemarco, line 76: `capacity_bytes` here is effectively unbounded (anything that fits in `size_t`). Given that you already have defensive sanity limits in the buffer (e.g. `unbelievable_frame_count`), would it be worth adding a similar upper cap here for consistency? Or do you think that's a responsibility of the one that instantiates the ring buffer?

kstenerud: The plan was for this to be the low level interface, where it's assumed you've already sanity checked your sizes. We could put a defensive check here, but then the question would be: what size is considered to be too big? If we start capturing other data-over-time, that will change.

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
```diff
@@ -0,0 +1,122 @@
...
+///
+/// Invariants:
+/// - Only one concurrent writer
+/// - Minimum 1ms between writes
+/// - Less than 1 million stack trace frames
+///
+/// @param buf The ring buffer.
+/// @param timestamp_ns Monotonic timestamp (nanoseconds).
+/// @param frames Array of frame addresses to copy.
+/// @param frame_count Number of frames in the array. Must not exceed UINT32_MAX.
```
fnewberg, line 160: the other PR had the default stacktrace depth limit set to 512, so there's no valid scenario where we'd get anywhere near this, right?

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+// Internal record header layout (stored in the ring buffer).
+typedef struct {
+    _Atomic uint32_t seq;  // Seqlock: odd = writing, even = stable.
+    uint32_t frame_count;  // Number of frames in this record.
+    uint64_t timestamp_ns; // Monotonic timestamp.
+} emb_ring_record_header_t;
+
+static const size_t header_size = sizeof(emb_ring_record_header_t);
+static const size_t frame_size = sizeof(((emb_ring_record_t *)0)->frames[0]);
+static const size_t unbelievable_frame_count = 1000000;
```
fnewberg: is this the per-stacktrace frames? if so, should we consider making this smaller, e.g. 10k or is there a compelling reason to keep it this big?

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+    // A single record must fit within the buffer capacity.
+    // Should never happen since we check against unbelievable_frame_count,
+    // but good just in case a future change allows it to slip through.
+    if (record_size_needed > buf->capacity) {
+        return false;
+    }
+
+    // Current write position (monotonically increasing byte offset).
+    uint64_t write_pos = atomic_load_explicit(&buf->write_pos, memory_order_relaxed);
+    uint64_t oldest_pos = atomic_load_explicit(&buf->oldest_pos, memory_order_acquire);
```
fnewberg: [claude] **Acquire fences on writer-side loads of `oldest_pos`**  The writer is the only thing that updates `oldest_pos`, so the acquire is redundant. Use `memory_order_relaxed` for both writer-side loads. Not a correctness bug — unnecessary fence cost on the async-safe hot path the design is trying to keep cheap.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+
+        if (frame_count > unbelievable_frame_count) {
+            return not_found;
+        }
+
+        if (timestamp_ns >= start_ns) {
+            break;
+        }
+
+        pos += record_size(frame_count);
```
fnewberg: not necessary now, but could we make scanning faster by approximating jump size based on how far off `timestamp_ns` was from `start_ns`

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+                               uint64_t end_ns)
+{
+    const buffer_range not_found = {0};
+    const size_t capacity = buf->capacity;
+    uint8_t *data = buf->data;
+    buffer_range range = {0};
+    uint64_t pos = low_pos;
+
+    // Find start (leading edge of first record where time >= start_ns)
+    while (pos < high_pos) {
```
fnewberg: [claude]

**Reader walks live records in `find_range` without seqlock validation** — `emb_ring_buffer.c:240-241, 263-264`. The seqlock protocol marks `seq` odd before the writer overwrites a slot's `frame_count`, `timestamp_ns`, and frames. But `find_range` reads `header->frame_count` and `header->timestamp_ns` directly with no parity check. A reader  sandwiched between `seq |= 1` and the writer's `frame_count` store can compute `pos += record_size(stale_frame_count)` and walk into garbage. The post-walk snapshot recheck at lines 307-311 only catches changes to `oldest_pos` / `write_pos`, not torn header reads at stable positions. Validate the seqlock
 inside `find_range`

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+    // The loop condition: oldest_pos + capacity < write_pos + record_size
+    // This means the oldest record would be overwritten.
+    while (oldest_pos + buf->capacity < write_pos + record_size_needed) {
+        // Read the header of the oldest record to determine its size.
+        size_t offset = oldest_pos % buf->capacity;
+        emb_ring_record_header_t *old_header = (emb_ring_record_header_t *)(buf->data + offset);
+
+        // Read frame_count (relaxed is safe: we're the single writer and this
+        // is an old committed record).
+        size_t old_frame_count = old_header->frame_count;
```
fnewberg, line 245: should we add a `unbelievable_frame_count` check here like we have in other places?
```suggestion
        size_t old_frame_count = old_header->frame_count;
        
        // Defense against torn / corrupt frame_count: if the value would advance
        // oldest_pos past write_pos, the record header is corrupt. Bail rather
        // than silently corrupt the ring further.
        if (old_frame_count > unbelievable_frame_count) {
            return false;
        }
```

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+        // Buffer is empty.
+        return result;
+    }
+    
+    buffer_range range = find_range(buf, initial_start_pos, initial_end_pos, start_ns, end_ns);
+
+    const uint64_t snapshot_start_pos = atomic_load_explicit(&buf->oldest_pos, memory_order_acquire);
+    const uint64_t snapshot_end_pos = atomic_load_explicit(&buf->write_pos, memory_order_acquire);
+    
+    if(snapshot_start_pos != initial_start_pos || snapshot_end_pos != initial_end_pos) {
```
fnewberg: [claude] (don't know if the assertion that multi-ms gaps being routine is correct)

**Single-shot snapshot retry is insufficient under preemption** . "we won't get torn again" because of the 1ms write cadence only holds if the reader thread isn't preempted for >1ms between the two `find_range` calls. Under iOS thermal/QoS pressure, multi-millisecond preemption gaps are routine. Replace with a bounded retry loop (e.g., up to 4 iterations) or explicit detection that returns a status the caller can act on.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+                      &max_prot,
+                      VM_INHERIT_NONE);
+    if (kr != KERN_SUCCESS) {
+        goto fail;
+    }
+ 
+    remap_succeeded = true;
+ 
+    // Sanity-check: VM_FLAGS_FIXED should guarantee the address didn't move.
+    if (second_half != region + half) {
```
fnewberg, line 123: [claude] 

**`vm_remap` partial-failure path leaks the second mapping** — `emb_ring_buffer.c:103-105` → `fail:` at 120-132. If `vm_remap` returns `KERN_SUCCESS` but parks the second half at a different address (`second_half != region + half`), cleanup deallocates `region + half` (which isn't mapped) and ignores the actual mapping the kernel chose. Capture the address `vm_remap` actually returned and deallocate that.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+#endif
+
+static size_t round_size_up_to_page_boundary(size_t size)
+{
+    size_t page = vm_page_size;
+    return (size + page - 1) & ~(page - 1);
+}
+
+emb_ring_buffer_t *emb_ring_buffer_create(size_t capacity_bytes)
+{
```
fnewberg: we get no feedback on failure modes. we could do something like this


```suggestion
emb_ring_buffer_t *emb_ring_buffer_create(size_t capacity_bytes, kern_return_t *kr_out)
{
    if (kr_out != NULL) *kr_out = KERN_SUCCESS;
```

and then handle errors like this

```
      kern_return_t kr = emb_vm_allocate(...);
      if (kr != KERN_SUCCESS) {
          if (kr_out != NULL) *kr_out = kr;
          return NULL;
      }
```

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+    }
+    // Deallocate the full 2 × capacity virtual region (both the original
+    // allocation and the remap share the same address range).
+    emb_vm_deallocate(mach_task_self(),
+                      (emb_vm_address_t)buf->data,
+                      (emb_vm_size_t)buf->capacity * 2);
+    free(buf);
+}
+
+bool emb_ring_buffer_write(emb_ring_buffer_t *buf,
```
fnewberg: Define an `emb_ring_status_t` enum and either return that so we can tell failures apart

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+        // vm_remap failed. VM_FLAGS_OVERWRITE may have already torn out the
+        // second half before failing, so we can only safely free the first half.
+        // The second half is either still part of the original allocation (and
+        // this covers it) or already gone.
+        emb_vm_deallocate(mach_task_self(), region, region_size);
+    }
+    return NULL;
+}
+
+void emb_ring_buffer_destroy(emb_ring_buffer_t *buf)
```
fnewberg, line 162: would this ever be used in production or is this just for tests? if used in prod, should we add some protection to prevent destroying while a reader is mid-read, e.g. with something like `_Atomic int active_readers` and have destroy spin/abort if non-zero?

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,393 @@
...
+        // Publish the new oldest_pos.
+        atomic_store_explicit(&buf->oldest_pos, oldest_pos, memory_order_release);
+    }
+
+    // Now that space is guaranteed, mark the record as "writing" via seqlock.
+    size_t offset = write_pos % buf->capacity;
+    emb_ring_record_header_t *header = (emb_ring_record_header_t *)(buf->data + offset);
+
+    // Set seq to odd (writing).
+    atomic_store_explicit(&header->seq, buf->next_seq | 1, memory_order_release);
```
fnewberg: `header-> seq` is s `uint32_t` and `buf->next_seq` is a `uint64_t`. should we make them be the same type? or at least explicitly cast here?

Tests/EmbraceProfilingSamplerTests/RingBufferConcurrencyTests.swift
```diff
@@ -0,0 +1,485 @@
...
+                buf: buf, readerCount: 4,
+                validate: { result, fc in
+                    for i in 0..<result.count {
+                        if result.records[i].frame_count > 10_000 {
+                            fc.record("wrap-around: record[\(i)].frame_count is corrupt")
+                        }
+                    }
+                },
+                writerBlock: { _ in
+                    // 60 writes with a one-page buffer forces many wrap-arounds.
```
fnewberg: nitpick: we declare buffers on L203 as being 4 pages

fnewberg (changes requested):

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
kstenerud: No, the extra capability is just being comfortably paranoid. Having deeper than 100 would be extreme.

The comment here exists because internally it uses a 32-bit variable, so it's physically incapable of more than `UINT32_MAX`. In a later PR I put in a configuration that controls it all, and keeps it within reasonable limits.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: This gets moved out to a config value in the next PR in order to make testing easier, and also give better control. Defaults to 500, and can go up to 1024 which is what the C layer enforces.

https://github.com/embrace-io/embrace-apple-sdk/pull/579/changes#diff-3dbde12cf9221285b3bde3e1a5afe9a910e521b3874c23f891260673ec190fc3R34

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: I thought about that, but since these are variable-sized, there's no way to know how far to jump beyond 1 record in order to be guaranteed you're on an actual header.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: Yes, it's used here:

* https://github.com/embrace-io/embrace-apple-sdk/pull/579/changes#diff-9f3e9295c0b7bce8d515720853a4034ca84b86ff09a1b9a32d5f47c2cfb50160R215
* https://github.com/embrace-io/embrace-apple-sdk/pull/579/changes#diff-9f3e9295c0b7bce8d515720853a4034ca84b86ff09a1b9a32d5f47c2cfb50160R235

The caller always makes sure it's not in use before calling.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: Fixed by backporting from https://github.com/embrace-io/embrace-apple-sdk/pull/579, but has moved to line 225. Line 225 uses `memory_order_relaxed` for both writer-side loads of `oldest_pos`, with an explanatory comment.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: This was also fixed by the backport. The THREE-LAYER CORRECTNESS MODEL comment at the top of Phase 1 in `emb_ring_buffer_read_range` explains the full correctness argument: primary safety comes from the `oldest_pos` + seq_cst fence protocol, with the seqlock seq field as advisory defense-in-depth.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: `unbelievable_frame_count` is gone in the latest version because it had become a redundant check. There's only one writer, so it can't corrupt itself. Originally `unbelievable_frame_count` had been a hedge against a very tiny race window in the read flow that was about as unlikely as cosmic rays flipping a bit. But it still technically existed thus the check ;-)

Now that hole is fully closed, so `unbelievable_frame_count` is no longer used.

The guard is still there. Line 252 checks `old_record_size > buf->capacity` and evicts everything on corruption. A valid record can never exceed capacity (enforced on the write path), so this catches any corrupted header.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: The newer code uses a three-phase approach (labeled Phase 1/2/3 in the source): scan live headers to find the range, copy it, then post-validate against `oldest_pos`. If eviction advanced into the copy, we skip the evicted prefix and count only the stable tail, so no retry needed.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: `VM_FLAGS_FIXED | VM_FLAGS_OVERWRITE` guarantees the kernel places the mapping at the exact requested address, making the `second_half != region + half` branch unreachable on a correctly functioning kernel. The comment at that line explains this, and the `fail:` path handles both the remap-succeeded and remap-failed cases independently.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: I've added explicit casts.

Tests/EmbraceProfilingSamplerTests/RingBufferConcurrencyTests.swift
kstenerud: Comment is gone now after the backport.

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
```diff
@@ -0,0 +1,185 @@
...
+static inline size_t emb_ring_record_size(uint32_t frame_count) {
+    return sizeof(emb_ring_record_header_t) + (size_t)frame_count * sizeof(uintptr_t);
+}
+
+/// Result from a read operation.
+typedef struct {
+    size_t records_offset; // Offset in bytes to the first matching record.
+    size_t record_count;   // Number of matching records written to the output buffer.
+    size_t total_bytes;    // Total size in bytes of the matching record set.
+} emb_ring_read_result_t;
```
fnewberg, line 109: do you think there is value in adding an error enum and distinguishing between the 4 cases that can cause premature return?

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,539 @@
...
+}
+
+/// Read a record header directly from the live ring buffer at the given
+/// absolute position. The double-mapping guarantees contiguous access.
+static inline emb_ring_record_header_t read_live_header(emb_ring_buffer_t *buf,
+                                                         uint64_t absolute_pos) {
+    size_t offset = (size_t)(absolute_pos % buf->capacity);
+    const emb_ring_record_header_t *hdr =
+        (const emb_ring_record_header_t *)(buf->data + offset);
+    return *hdr;
```
fnewberg: nitpick: i don't think this has any implementation issues, but TSan could complain about the copying of an atomic var without using atoimic operations

```suggestion
    const emb_ring_record_header_t *hdr =
        (const emb_ring_write_header_t *)(buf->data + offset);
    emb_ring_record_header_t out;
    // seq is _Atomic on the writer side; load it explicitly with acquire to satisfy C11
    out.seq = atomic_load_explicit(&whdr->seq, memory_order_acquire);
    out.frame_count = whdr->frame_count;
    out.timestamp_ns = whdr->timestamp_ns;
    return out;
```

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
```diff
@@ -0,0 +1,539 @@
...
+        size_t offset = oldest_pos % buf->capacity;
+        emb_ring_write_header_t *old_header = (emb_ring_write_header_t *)(buf->data + offset);
+        size_t old_frame_count = old_header->frame_count;
+        size_t old_record_size = record_size(old_frame_count);
+
+        // Guard against corrupted headers (e.g. bit flip, stale VM page).
+        // A valid record is always <= capacity (enforced by the write path).
+        // On corruption, evict everything to prevent cascading garbage reads.
+        if (old_record_size > buf->capacity) {
+            oldest_pos = write_pos;
```
fnewberg, line 252: for things like this, which we wouldn't expect to ever happen, do you think we should have some observability mechanism, e.g. we write something to files that we pick up on next SDK startup? i think these are things that aren't critical, but it'd be helpful to know if they happen at some appreciable rate when millions of devices are running this

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
```diff
@@ -0,0 +1,185 @@
...
+/// Read records from the ring buffer within a time range.
+///
+/// Scans the live buffer to locate the matching time range, then copies only
+/// the matching records into the caller-provided output buffer. This avoids
+/// copying the entire ring buffer contents for small time ranges.
+///
+/// Walk the results starting at `output + result.offset`, casting to
+/// `emb_ring_record_header_t *`, reading `frame_count`, advancing by
+/// `emb_ring_record_size(frame_count)`, and repeating for `result.count`
+/// records.
```
fnewberg: The struct fields are `records_offset` and `record_count` (defined at `:87-88`), not `offset` and `count`.

```suggestion
/// Walk the results starting at `output + result.records_offset`, casting to
/// `emb_ring_record_header_t *`, reading `frame_count`, advancing by
/// `emb_ring_record_size(frame_count)`, and repeating for `result.record_count`
/// records.
```

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: Added `kern_return_t *kr_out` out-param (nullable). Each failure path sets it: `KERN_INVALID_ARGUMENT` for zero capacity, the vm_allocate/vm_remap kern_return_t on Mach failures, `KERN_FAILURE` for the unreachable placement check, and `KERN_RESOURCE_SHORTAGE` for calloc failure.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: This now returns `emb_ring_write_status_t` with four values: `EMB_RING_WRITE_OK`, `EMB_RING_WRITE_BAD_ARGS`, `EMB_RING_WRITE_RECORD_TOO_LARGE`, and `EMB_RING_WRITE_CORRUPTION_DETECTED`. The corruption path (previously a silent evict-everything-and-continue) now aborts the write and returns `CORRUPTION_DETECTED`.

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
kstenerud: I've added `emb_ring_read_status_t status` field to `emb_ring_read_result_t`, with five values: `OK`, `BAD_ARGS`, `RESETTING`, `EMPTY`, and `TRUNCATED`. Every return path in `emb_ring_buffer_read_range` sets it.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: `read_live_header` now reads `seq` via `atomic_load_explicit(..., memory_order_acquire)` from the underlying `emb_ring_write_header_t`, then copies `frame_count` and `timestamp_ns` non-atomically. The acquire on `seq` establishes the happens-before needed to observe the stable frame data.

Sources/EmbraceProfilingSampler/source/emb_ring_buffer.c
kstenerud: The corruption path now returns `EMB_RING_WRITE_CORRUPTION_DETECTED` (and aborts the write) rather than silently recovering. Callers can log or count it as they see fit.

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
kstenerud: Updated to `result.records_offset` and `result.record_count`.

ferdraghi (approved):

fnewberg (approved):

PR #556, opened 2026-04-07
[EMBR-10283] Feature: New Framerate based Hang Detection

ferdraghi:
- Revamps the main thread Hang detection feature to base it on framerate using CADisplayLink.
- Removes EmbraceClock because it's been rendered redundant for the new approach.
- WatchOS is unsuported for now since CADisplayLink is not available on the platform. Support for the watch will be revisited in the future.
- Macos is supported starting on macos 14 Sonoma.


Sources/EmbraceConfiguration/HangLimits.swift
```diff
@@ -30,6 +35,9 @@ import Foundation
         guard let other = object as? Self else {
             return false
         }
-        return hangPerSession == other.hangPerSession && samplesPerHang == other.samplesPerHang && reportsWatchdogEvents == other.reportsWatchdogEvents
+        return hangThreshold == other.hangThreshold
+            && hangPerSession == other.hangPerSession
+            && samplesPerHang == other.samplesPerHang
```
ArielDemarco: You're not really using `samplesPerHang`. You might wanna remove it?

ferdraghi: you're right, it's a leftover from a previous implementation. I'll remove it.

Sources/EmbraceCore/Capture/Hang/Watchdog/HangObserver.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+
+import Foundation
+
+/// Protocol for objects that observe hang detection events.
+///
+/// - `hangStarted`: Called when a hang is first detected.
+/// - `hangUpdated`: Called periodically while a hang persists (not used by `FrameRateMonitor`).
+/// - `hangEnded`: Called when the hang resolves.
+public protocol HangObserver: AnyObject {
```
NachoEmbrace: Do we need this to be public?

Sources/EmbraceCore/Capture/Hang/Watchdog/FrameRateMonitor.swift
```diff
@@ -0,0 +1,134 @@
...
+    /// tick). A delay exceeding `threshold` is reported as a completed hang to
+    /// the `hangObserver`.
+    ///
+    /// This approach is dynamic-rate–safe: using `targetTimestamp` rather than a
+    /// fixed frame duration means ProMotion, Low Power Mode, and
+    /// `preferredFrameRateRange` transitions never produce false positives.
+    /// `CADisplayLink` also pauses automatically in the background, so suspend
+    /// gaps are excluded without any extra bookkeeping.
+    @available(macOS 14.0, *)
+    public final class FrameRateMonitor {
```
NachoEmbrace: I dont think we want this to be public

NachoEmbrace (changes requested):

Sources/EmbraceIO/Capture/CaptureServicesOptionsBuilder.swift
```diff
@@ -121,6 +121,15 @@ public class CaptureServicesOptionsBuilder: NSObject {
...
         }
 
+        // hang
+        #if !os(watchOS)
+            if #available(macOS 14.0, *) {
+                if map[.hang] == nil {
+                    map[.hang] = true
+                }
+            }
+        #endif
```
NachoEmbrace: I assume this was intentional, but wanted to double check.
Do we actually want to include this capture service as a default?

NachoEmbrace (approved):

Sources/EmbraceIO/Capture/CaptureServicesOptionsBuilder.swift
NachoEmbrace: Per our discussion today, we're removing this from the defaults for now.

PR #562, opened 2026-04-16
[EMBR-11398] Feature: Ordered uploads

NachoEmbrace:
Redesigns the EmbraceUploadInternal module to deliver spans and logs in order, with a cache-first architecture that guarantees data durability before signaling success to callers.

## Architecture

* **Coordination queue** — A single serial DispatchQueue serializes all state: cache reads/writes, in-flight tracking, and queue-fill decisions. It never executes network requests.

* **Three upload queues** — spansQueue and logsQueue are serial OperationQueues (one operation at a time, dependency-chained) guaranteeing ordered delivery. attachmentsQueue is concurrent (no ordering constraint).

* **Cache-first flow** — Every upload call follows the same path:
   1. Dispatch onto the coordination queue
   2. Validate input, write to CoreData cache
   3. Fire the completion callback (data is durable at this point)
   4. Call fillQueue to create upload operations from cached records

* **fillQueue** is the single mechanism for creating upload operations. It checks queue capacity against queueLimit, fetches cached records sorted by date (excluding in-flight IDs), chains dependencies for ordered types, and  
  enqueues. It runs after every cache write and after every operation completion, draining the cache as capacity opens up.

* **Operation lifecycle** — On success or failure, the record is deleted from cache. On cancellation, the record is preserved for replay on the next process launch via retryCachedData().                                        
                                   
## Key changes
- Ordered delivery for spans and logs via serial queues with dependency chaining                                                
- Cache-first: completion fires after cache write, not after upload
- Retry semantics: unlimited retries by default (automaticRetryCount: -1), 250ms exponential backoff, only .badURL/.unsupportedURL are non-retriable (all 4xx/5xx are retried)
- Queue capacity: queueLimit caps in-flight operations per type; overflow stays in cache                                        
- Removed attemptCount from CoreData entity and maximumAmountOfRetries from options                                             
- Fixed AsyncOperation.cancel() not calling finish() (operation lifecycle bug)

### DIFF
--- a/Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
+++ b/Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
@@ -10,7 +10,8 @@ import Foundation
 
 enum EmbraceUploadOperationResult: Equatable {
     case success
-    case failure(retriable: Bool)
+    case failure
+    case cancelled
 }
 
 typealias EmbraceUploadOperationCompletion = (_ result: EmbraceUploadOperationResult, _ attemptCount: Int) -> Void
@@ -64,7 +65,8 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
         task?.cancel()
         task = nil
 
-        completion?(.failure(retriable: true), attemptCount)
+        completion?(.cancelled, attemptCount)
+        finish()
     }
 
     override func execute() {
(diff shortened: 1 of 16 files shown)


Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
```diff
@@ -153,36 +158,19 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
...
+                .badURL:
                 return false
             default:
                 return true
             }
         }
 
-        // handle HTTP status codes:
-        // retry only if is an error (client/server) and statusCode is not 429
+        // all 4xx and 5xx are retriable
```
ferdraghi, line 171: Should ALL 4xx codes be retriable though? I'm thinking codes like 404 wouldn't make sense to retry, but maybe this is handled somewhere else and these codes wouldn't be expected at this stage?

NachoEmbrace: Our backend never sends any 4xx other than 429.
So it's safe to assume that any 4xx error we get is due to bad connection / proxying / vpn / etc issues.

ferdraghi (approved):

PR #574, opened 2026-04-23
[EMBR-11792] Feature: Feat/embr 11792 remove attributes from breadcrumbs

NachoEmbrace:
Also added missing tests for breacdrumbs and push notification events


ferdraghi (approved):

ferdraghi:
Echoing Fredric's suggestion to deprecate these removals on the current 6.x sdk, probably a good idea to do it with this as well.

NachoEmbrace:
> Echoing Fredric's suggestion to deprecate these removals on the current 6.x sdk, probably a good idea to do it with this as well.

Already done a while ago

PR #575, opened 2026-04-23
[EMBR-11793] Feature: Removing userName and userEmail metadata keys

NachoEmbrace:


ferdraghi (approved):

Sources/EmbraceIO/EmbraceIO+Metadata.swift
```diff
@@ -5,32 +5,8 @@
 /// User QoL accessors
 extension EmbraceIO {
 
-    /// Set a 'name' for the current user.
-    /// Will be set permanently until explicitly unset via the `clearUserProperties()` method.
-    /// - Note: No validation is done on the identifier. Be sure it matches or can be mapped to a record in your system
-    public var userName: String? {
```
fnewberg, line 11: should we mark these methods being removed as being deprecated in our next 6.x release?

ferdraghi: probably not a bad idea

PR #579, opened 2026-04-24
Add lock-free main thread profiling sampler

kstenerud:
## Summary

Builds on #538 (stack walker) and #555 (ring buffer) to add the main thread profiling sampler and integrate everything into the Swift `ProfilingEngine`.

The sampler runs a dedicated worker thread that periodically suspends the main thread via Mach APIs, walks the frame-pointer chain, writes the sample to the lock-free ring buffer, then immediately resumes, keeping suspension windows in the microsecond range.

### Sampler (`emb_sampler`)
- Atomic state machine (STOPPED → STARTING → RUNNING → STOPPING, plus ZOMBIE/REAPING/FAULTED) governs lifecycle and prevents concurrent start/stop races
- Drift recovery shortens subsequent intervals when sampling overruns, floored at `min_sampling_interval_ms` to avoid starving the main thread
- Optional fallback walker callback for when frame-pointer walk yields too few frames (e.g. leaf functions). This is where we can plug KSCrash in when the new release is out of beta.
- Terminal FAULTED state intentionally leaks resources to prevent cascading failures when runtime is broken

### Ring buffer improvements
- Read path rewritten: scans live buffer for time range, copies only matching records, validates post-copy. Eliminates heap allocation on the read path
- Added `reset()` with Dekker-style mutual exclusion for buffer reuse across sessions
- Corruption guard and ARM64 memory fence after eviction to prevent torn reads

### Stack walker improvements
- Stack bounds now passed by caller instead of resolved via `pthread_get_stackaddr_np`/`pthread_get_stacksize_np`, which are not async-signal-safe and risk deadlock when the target thread is suspended
- Added 16-byte frame pointer alignment check for ARM64

### `ProfilingEngine` (Swift)
- Replaced prototype fake data generation with real sampler integration
- Uses atomic CAS gate instead of `os_unfair_lock`. If the main thread holds the lock when suspended by the sampler worker, any waiter would deadlock since the kernel knows the owner is suspended
- Fail-fast `Result` enums (`StartResult`, `RetrieveResult`) instead of throwing
- Two-pass sample retrieval populates a single flat frame array shared across all samples (40–60% memory reduction vs. per-sample arrays)
- `ProfilingConfiguration` made public with sensible defaults (10 Hz, 1 MB buffer, 500 max frames)

### Other
- `StackCapture.swift` deleted; functionality moved into sampler worker thread
- Comprehensive test coverage: sampler lifecycle, concurrency, fallback walker, ring buffer reset/corruption, and Swift integration

## Checklist


Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -14,54 +15,270 @@
...
+        /// Atomic gate. CAS-based, no kernel ownership tracking.
+        private let gateIsAcquired: UnsafeMutablePointer<emb_atomic_bool_t>
     #endif
 
-    /// Returns `true` if the profiling engine is currently running.
-    public var isRunning: Bool {
+    /// Sleep interval between gate acquisition attempts (microseconds).
+    private static let gateSleepUs: useconds_t = 100
+    /// Maximum cumulative actual uptime before giving up (nanoseconds). 10ms.
+    private static let gateMaxSleepNs: UInt64 = 10_000_000
```
fnewberg: nitpick: why `Us` for `gateSleepUs` and `Ns` for `gateMaxSleepNs`?

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -14,54 +15,270 @@
...
+                }
+                readBuffer = UnsafeMutableRawPointer(newReadBuffer)
+                readBufferSize = capacity
+            }
+
+            let config = emb_sampler_config_t(
+                sampling_interval_ms: configuration.samplingIntervalMs,
+                min_sampling_interval_ms: configuration.minSamplingIntervalMs,
+                max_frames: configuration.maxFramesPerSample,
+                min_frames: 0,
```
fnewberg: do we have configurability that we do not expect to need since these two settings are hard-coded?

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -4,6 +4,7 @@
 
 #if !os(watchOS)
     import Darwin
+    import EmbraceAtomicsShim
```
fnewberg, line 7: add

```
subs.dependency "EmbraceIO/EmbraceAtomicsShim"
```

to `EmbraceProfiling` section of `EmbraceIO.podspec`

Sources/EmbraceProfilingSampler/source/emb_sampler.c
```diff
@@ -1,36 +1,437 @@
...
+        // BEGIN ASYNC-SAFE SECTION
+        // - Do NOT call sampler_checked_transition
+        // - Do NOT emit logs
+        // - Do NOT allocate
+        // - Do NOT lock
+
+        uint64_t timestamp_ns = clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW);
+
+        size_t frame_count = 0;
+        emb_stack_walk(g_main_thread,
```
fnewberg: should we handle the failure case here?

Sources/EmbraceProfilingSampler/source/emb_sampler.c
```diff
@@ -1,36 +1,437 @@
...
+        config.max_frames == 0 || config.min_sampling_interval_ms > config.sampling_interval_ms) {
+        os_log_debug(emb_profiling_log(), "emb_sampler_start failed: invalid config");
+        sampler_checked_transition(EMB_SAMPLER_STARTING_OR_STOPPING_MASK, EMB_SAMPLER_STOPPED);
+        return EMB_SAMPLER_START_ERROR;
+    }
+
+    if (config.max_frames > EMB_MAX_STACK_FRAMES) {
+        os_log_debug(emb_profiling_log(),
+                     "Clamping max_frames from %u to %d", config.max_frames, EMB_MAX_STACK_FRAMES);
+        config.max_frames = EMB_MAX_STACK_FRAMES;
```
fnewberg: nitpick: we are changing the config value here, so wouldn't the `memcmp(&config, &g_config, sizeof(config)) == 0)` check on L289 fail even if we passed in the same config twice since the first one was mutated? should we move this check to be before L 282?

Sources/EmbraceProfilingSampler/source/emb_sampler.c
```diff
@@ -1,36 +1,437 @@
...
+/// Returns true on first successful CAS.
+/// If no CAS succeeds, (i.e. actual state is not in the mask), logs error, faults, and returns false.
+static inline bool sampler_checked_transition(int valid_from_mask, emb_sampler_state_t desired) {
+    int actual = sampler_cas_multi(valid_from_mask, desired);
+    if (actual == CAS_SUCCESS) {
+        return true;
+    }
+    os_log_error(emb_profiling_log(),
+                 "State violation: expected mask 0x%x, actual 0x%x", valid_from_mask, actual);
+    sampler_fault("state machine violation");
```
fnewberg:
```suggestion
     // Don't clobber an existing fault reason with a less-informative one.
     if (actual != EMB_SAMPLER_FAULTED) {
        sampler_fault("state machine violation");
     }    
```

Sources/EmbraceProfilingSampler/include/emb_sampler.h
```diff
@@ -1,38 +1,207 @@
...
 #endif
 
-/// Opaque sampler handle.
-typedef struct emb_sampler *emb_sampler_t;
+/// Hard upper limit on stack frames per sample.
+#define EMB_MAX_STACK_FRAMES 1024
+
+/// Sampler lifecycle states.
+///
+/// Transitions:
```
fnewberg, line 67: nitpick: missing these possible transitions

- STOPPING → STOPPED due to start setup-failed 
- RUNNING → FAULTED via ring-buffer write failure
- STARTING → FAULTED via "unhandled state" /
    "state machine violation"

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -14,54 +15,270 @@
...
+                if let reason = faultReason {
+                    return .faulted(reason: reason)
+                }
+                return .samplerStartFailed
+            }
         #endif
     }
 
-    /// Stop the profiling engine.
+    /// Request the profiling engine to stop.
```
fnewberg, line 298: [claude]

    `stop()` tells callers to poll `isCapturing` to know when the worker has stopped. But `isCapturing` only returns `true` for `EMB_SAMPLER_RUNNING` — it flips to `false` the moment `stop()` advances state to STOPPING, while the worker thread is still alive:

    - **STOPPING** — worker may be in `mach_wait_until` between samples, or executing one last cycle (with a Mach-suspended main thread).
    - **ZOMBIE** — worker has exited but `pthread_join` hasn't run yet.

    A caller following the doc would conclude "stopped" and could safely destroy resources the worker is still touching. The deinit comment at `:386` already points to `emb_sampler_is_active()`, so the right answer exists —it's just not exposed in Swift.

    Suggested fix — add an `isActive` property mirroring the C-side semantics, and reword `stop()`:

    ```swift
    /// Returns `true` while the worker thread is alive — covers RUNNING,
    /// STOPPING, ZOMBIE, and REAPING. Polling until this returns `false`
    /// is the correct way to wait for the worker to fully exit.
    public var isActive: Bool {
        #if os(watchOS)
            return false
        #else
            return emb_sampler_is_active()
        #endif
    }
    ```

    ```swift
    /// This is non-blocking: it signals the worker thread to exit and
    /// returns immediately. Poll ``isActive`` to determine when the worker
    /// has actually stopped. ``isCapturing`` flips to `false` as soon as
    /// stop is requested — use it for "should I be running?" checks, not
    /// teardown coordination.
    ```

Sources/EmbraceProfilingSampler/include/emb_sampler.h
```diff
@@ -1,38 +1,207 @@
...
+/// Returns ERROR on invalid config, null buffer, thread creation failure,
+/// or if the sampler is in a permanently faulted state.
+emb_sampler_start_result_t emb_sampler_start(emb_ring_buffer_t *buffer, emb_sampler_config_t config);
 
-/// Start sampling. Returns true on success.
-bool emb_sampler_start(emb_sampler_t sampler);
+/// Request the sampler to stop.
+///
+/// This is non-blocking and fire-and-forget: it signals the worker thread
+/// to exit and returns immediately with no feedback. The worker thread
```
fnewberg, line 191: any value in passing through the return value from `sampler_cas_multi` that this calls?

Sources/EmbraceProfilingSampler/source/emb_sampler.c
```diff
@@ -1,36 +1,437 @@
...
+// ---------------------------------------------------------------------------
+// TEST-ONLY APIs
+// ---------------------------------------------------------------------------
+
+bool emb_sampler_reset_for_testing(void) {
+    emb_sampler_state_t state = sampler_get_state();
+    if (state != EMB_SAMPLER_STOPPED && state != EMB_SAMPLER_FAULTED) {
+        return false;
+    }
+    g_fault_reason = NULL;
```
fnewberg, line 505: [claude suggestion]

```suggestion

    // From FAULTED: the production leak philosophy skipped pthread_join and
    // free(g_stack_frames) to avoid risking deadlock or UAF in a broken
    // runtime. Tests need to recover these resources to avoid unbounded
    // growth across runs. By the time we get here all fault paths have
    // returned NULL from the worker (or will on its next loop-state check),
    // and no further access to g_stack_frames occurs after that exit.
    //
    // From STOPPED: cleanup_previous_session has already cleared
    // g_stack_frames and g_buffer; these assignments are idempotent.
    free(g_stack_frames);
    g_stack_frames = NULL;
    g_buffer = NULL;
    g_fault_reason = NULL;
```

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -14,54 +15,270 @@
...
+        #if !os(watchOS)
+            self.gateIsAcquired = .allocate(capacity: 1)
+            emb_atomic_bool_init(gateIsAcquired, false)
+        #endif
+    }
+
+    #if !os(watchOS)
+    /// Try to acquire the gate.
+    /// Sleeps ``gateSleepUs`` between attempts, up to ``gateMaxSleepNs``
+    /// cumulative actual elapsed time, then one final attempt (in case actual
```
fnewberg: nitpick: i don't think the code does the final check. update the doc or the code so they are aligned

Sources/EmbraceProfilingSampler/source/emb_sampler.c
```diff
@@ -1,36 +1,437 @@
...
+    return (emb_sampler_state_t)atomic_load_explicit(&g_state, memory_order_acquire);
+}
+
+/// Returns CAS_SUCCESS (0) on success, or the actual state on failure.
+/// Enum values are all non-zero (power-of-2), so 0 is unambiguous.
+static inline int sampler_cas_state(emb_sampler_state_t expected, emb_sampler_state_t desired) {
+    int e = expected;
+    if (atomic_compare_exchange_strong_explicit(&g_state, &e, desired,
+                                                memory_order_release,
+                                                memory_order_relaxed)) {
```
fnewberg: nitpick: [claude]

Relaxed failure-load doesn't synchronize-with the worker's RUNNING release, so the case at :288-289 can read stale g_buffer/g_config and return a false START_OK (matching a previous session's config). Practically benign on arm64/x86_64 but TSan-flagged. memory_order_acquire here is free in codegen.

```suggestion
                                                memory_order_acquire)) {
```

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -70,110 +287,107 @@ public final class ProfilingEngine {
...
+    }
+
+    // Note: ProfilingEngine is a singleton; deinit never runs.
+    // If changed to non-singleton, deinit would need to:
+    //   1. Call emb_sampler_stop(), then poll emb_sampler_is_active() until
+    //      false (with a timeout to avoid unbounded blocking).
+    //   2. Destroy the ring buffer: emb_ring_buffer_destroy(ringBuffer)
+    //   3. Deallocate readBuffer: if let ptr = readBuffer { free(ptr) }
+    //   4. Deinitialize and deallocate the atomic gate:
+    //      emb_atomic_bool_deinit(gateIsAcquired)
```
fnewberg: nitpick: i don't think this function exists

Sources/EmbraceProfilingSampler/source/emb_sampler.c
```diff
@@ -1,36 +1,437 @@
...
+                       g_config.max_frames,
+                       &frame_count);
+
+        if (frame_count < g_config.min_frames && fallback_walker != NULL) {
+            bool is_truncated = false;
+            int result = fallback_walker(g_main_thread,
+                                         g_stack_frames,
+                                         (int)g_config.max_frames,
+                                         &is_truncated);
+            frame_count = result < 0 ? 0 : (size_t)result;
```
fnewberg: [claude] Fallback walker returns negative on explicit failure (vs `0` = "no frames found"). Currently collapsed to `frame_count = 0` — indistinguishable from the common "thread in syscall" case. With a pluggable walker like KSCrash, real failures get attributed to idle-thread telemetry.

Consider a per-session counter (resets in `emb_sampler_start`, mirrors the existing `get_state` / `get_fault_reason` getters):

```c
static _Atomic uint32_t g_fallback_failures = 0;

if (frame_count < g_config.min_frames && fallback_walker != NULL) {
    bool is_truncated = false;
    int result = fallback_walker(g_main_thread,
                                 g_stack_frames,
                                 (int)g_config.max_frames,
                                 &is_truncated);
    if (result < 0) {
        atomic_fetch_add_explicit(&g_fallback_failures, 1, memory_order_relaxed);
        frame_count = 0;
    } else {
        frame_count = (size_t)result;
    }
}
```

`uint32_t emb_sampler_get_fallback_failure_count(void)` exposes it. Pairs with the `emb_stack_walk` return-discarded finding at `:202` — same shape, same counter pattern.

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -14,54 +15,270 @@
...
+/// `os_unfair_lock` when suspended, any thread that tries to acquire that
+/// lock will deadlock. The kernel knows the owner is suspended and will
+/// never schedule the waiter. With a plain atomic CAS gate, waiters spin
+/// briefly and time out, avoiding deadlock. This two-layer design (Swift
+/// atomic gate + C atomic state machine + ring buffer seqlock) is intentional.
+public final class ProfilingEngine: @unchecked Sendable {
+    public static let shared = ProfilingEngine()
+
+    /// Result of calling ``start(configuration:)``.
+    public enum StartResult: Equatable {
```
fnewberg:
```suggestion
    public enum StartResult: Equatable, Sendable {
```

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -14,54 +15,270 @@
...
+        case configMismatch
+        /// The sampler has entered an unrecoverable faulted state (bug or broken runtime).
+        /// The associated string describes the fault reason.
+        case faulted(reason: String)
+        /// Profiling is not supported on this platform (e.g. watchOS).
+        case notSupported
+    }
+
+    /// Result of calling ``retrieveSamples(from:through:)``.
+    public enum RetrieveResult {
```
fnewberg:
```suggestion
    public enum RetrieveResult: Equatable, Sendable {
```

Sources/EmbraceProfilingSampler/source/emb_sampler.c
```diff
@@ -1,36 +1,437 @@
...
+        // END ASYNC-SAFE SECTION
+        // ====================================================================
+
+        // Write the captured stack trace to the ring buffer.
+        // In normal operation this always succeeds because emb_sampler_start
+        // validates that max_frames fits within the buffer capacity. A failure
+        // here indicates a logic error (e.g. buffer was destroyed externally).
+        if (frame_count > 0) {
+            if (!emb_ring_buffer_write(g_buffer, timestamp_ns, g_stack_frames, frame_count)) {
+                sampler_fault("ring buffer write failed (buffer corrupt or misconfigured)");
```
fnewberg: Consider adding a test for this FAULTED path. The only worker-fault path covered today is `thread_suspend` failure (`SamplerTests.swift:322`); a regression that silently swallowed write errors here would go unnoticed.

fnewberg:
Do we have any tests that exercise reading while the writer is running? The current `retrieveSamples` tests appear to only exercise the API after `stop`.

Tests/EmbraceProfilingSamplerTests/SamplerTests.swift
```diff
@@ -0,0 +1,662 @@
...
+
+            // Clear override before assertions (cleanup).
+            emb_sampler_set_main_thread_for_testing(0, nil)
+
+            let records = testReadRange(buf, 0, UInt64.max)
+            XCTAssertGreaterThanOrEqual(records.count, 1,
+                "Should have captured at least 1 sample from the test thread")
+            for record in records {
+                XCTAssertGreaterThan(record.frame_count, 0,
+                    "Samples should contain frames from the test thread stack")
```
fnewberg: [claude] Tightening so this actually proves the redirect. The test thread is in a 20-deep recursion before settling into the spin loop, so the current `> 0` check would pass even if the redirect silently fell back to a different thread (which would also produce frames, just from a different stack). Bumping the floor to 18 catches that case while leaving slack for system frames and rare partial walks.

```suggestion
                XCTAssertGreaterThanOrEqual(record.frame_count, 18,
                    "Sample depth should reflect the test thread's ~20-deep recursion; a low count suggests the redirect didn't take effect and we sampled a different thread")
```

Tests/EmbraceProfilingSamplerTests/SamplerTests.swift
```diff
@@ -0,0 +1,662 @@
...
+            DispatchQueue.global().async {
+                bgResult = emb_sampler_start(buf, config)
+                group.leave()
+            }
+            group.wait()
+
+            XCTAssertEqual(bgResult, EMB_SAMPLER_START_ERROR,
+                "Start from non-main thread without cached main thread should fail")
+        }
+    }
```
fnewberg, line 782: [claude] Lock down "FAULTED is terminal" across every entry point — the leak philosophy at `emb_sampler.c:111-127` depends on it. Currently only `start()` from FAULTED is asserted; `stop()` and `is_active()` aren't. A regression that broadens the stop CAS mask to include FAULTED, or moves FAULTED out of `is_active()`'s default branch, would silently let resources be freed on a broken runtime.

```suggestion
        func test_faulted_isStickyAcrossAllOperations() {
            let buf = makeBuffer()
            defer { emb_ring_buffer_destroy(buf) }

            emb_sampler_inject_fault_for_testing("injected for stickiness test")
            XCTAssertEqual(emb_sampler_get_state(), EMB_SAMPLER_FAULTED)

            // stop() must not transition out of FAULTED.
            emb_sampler_stop()
            XCTAssertEqual(emb_sampler_get_state(), EMB_SAMPLER_FAULTED,
                "stop() from FAULTED must be a no-op")

            // is_active() reports false (sampler is dead, not running).
            XCTAssertFalse(emb_sampler_is_active(),
                "is_active() must return false from FAULTED")

            // start() returns ERROR.
            let config = emb_sampler_config_t(
                sampling_interval_ms: 50, min_sampling_interval_ms: 10,
                max_frames: 64, min_frames: 0, fallback_walker: nil)
            XCTAssertEqual(emb_sampler_start(buf, config), EMB_SAMPLER_START_ERROR,
                "start() from FAULTED must return ERROR")

            XCTAssertEqual(emb_sampler_get_state(), EMB_SAMPLER_FAULTED,
                "FAULTED is terminal across all operations")
        }
    }
```

Sources/EmbraceProfiling/ProfilingSample.swift
```diff
@@ -2,32 +2,47 @@
...
+/// ``ProfilingResult/frames(for:)`` for convenient per-sample access.
 public struct ProfilingSample: Equatable, Hashable, Sendable {
     /// `CLOCK_MONOTONIC_RAW` timestamp in nanoseconds when this sample was captured.
     public let timestamp: UInt64
 
-    /// Raw return addresses from the captured stack trace.
+    /// Index range into the shared frames array for this sample's return addresses.
+    public let frameRange: Range<Int>
+
+    public init(timestamp: UInt64, frameRange: Range<Int>) {
```
fnewberg: [claude] Tightening the public surface — there's no caller path that constructs `ProfilingSample`. SDK customers consume samples via `retrieveSamples`, never produce them. Internal tests can use `@testable import`. The current public init exists as a side effect of `public let` fields, not as a deliberate API decision. Locking this down pre-release; going public→internal post-release is a source break for any customer who happened to use it.

```suggestion
    init(timestamp: UInt64, frameRange: Range<Int>) {
```

Sources/EmbraceProfiling/ProfilingSample.swift
```diff
@@ -2,32 +2,47 @@
...
-    /// The method used to capture this stack trace.
-    public let unwindMethod: StackUnwindMethod
+    /// Access the frame addresses for a specific sample.
+    public func frames(for sample: ProfilingSample) -> ArraySlice<UInt> {
+        frames[sample.frameRange]
+    }
 
-    public init(timestamp: UInt64, frames: [UInt], unwindMethod: StackUnwindMethod) {
-        self.timestamp = timestamp
+    public init(samples: [ProfilingSample], frames: [UInt]) {
```
fnewberg: [claude] Same reasoning as the `ProfilingSample.init` comment — `ProfilingResult` is constructed by `retrieveSamples`, not by consumers. Tightening pre-release.

```suggestion
    init(samples: [ProfilingSample], frames: [UInt]) {
```

Sources/EmbraceProfilingSampler/include/emb_sampler.h
```diff
@@ -1,38 +1,207 @@
...
+/// the suspended thread holds the malloc lock).
+///
+/// Designed to match the signature of ksbt_captureBacktraceFromSuspendedMachThread
+/// (see kstenerud/KSCrash#816).
+///
+/// @param thread        Mach thread port (already suspended).
+/// @param frames_out    Output buffer for frame addresses.
+/// @param max_frames    Capacity of frames_out.
+/// @param is_truncated  Set to true if the stack exceeded max_frames.
+/// @return The number of frames captured, or 0 on failure. Must be positive.
```
fnewberg: the sampler at `emb_sampler.c:215` actually treats any `<= 0` value as failure (`result < 0 ? 0 : (size_t)result`)

```suggestion
/// @return Number of frames captured (>0 on success); <= 0 indicates failure.
```

Sources/EmbraceProfilingSampler/include/emb_sampler.h
```diff
@@ -1,38 +1,207 @@
...
+    EMB_SAMPLER_START_CONFIG_MISMATCH, // Already running with a different config or buffer.
+    EMB_SAMPLER_START_ERROR,           // Permanent failure (invalid config, null buffer, faulted, etc).
+} emb_sampler_start_result_t;
 
-/// Create a sampler. Returns NULL on failure.
-emb_sampler_t emb_sampler_create(void);
+/// Start sampling the main thread into the given ring buffer.
+///
+/// May be called from any thread. Main thread info is captured at load
+/// time via a constructor. If main thread is unavailable, returns ERROR.
```
fnewberg, line 164: The current doc only describes the typical case (constructor ran on main → call from any thread). The actual contract has a conditional second case: if the SDK is loaded from a background thread (e.g. dlopen), the first `start()` must be on main to seed the cache — otherwise it returns `ERROR` and the caller has to grep `os_log_debug` to figure out why.

```suggestion
/// Main thread info is cached at load time by a constructor when the SDK
/// is loaded on main (typical case). If the SDK is loaded from a
/// background thread, call start() from the main thread at least once
/// to seed the cache; subsequent calls may then come from any thread.
/// Returns ERROR if main thread info is unavailable.
```

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -70,110 +287,107 @@ public final class ProfilingEngine {
     /// Time values are in `CLOCK_MONOTONIC_RAW` nanoseconds, matching timekeeping used in EmbraceClock.swift.
     ///
     /// Samples are returned in chronological order. The engine does not need to be running.
-    /// Samples persist after ``stop()`` is called.
+    /// Samples persist after ``stop()`` is called, but are cleared on ``start(configuration:)``.
```
fnewberg: Aligning with implementation

```suggestion
    /// Samples persist after ``stop()`` is called, and persist across calls to ``start(configuration:)`` that match the current session. Samples are cleared when ``start(configuration:)`` is called while the sampler is not already active.
```

Sources/EmbraceProfiling/ProfilingConfiguration.swift
```diff
@@ -1,15 +1,52 @@
...
+public struct ProfilingConfiguration: Sendable {
     /// Sampling interval in milliseconds.
-    let samplingIntervalMs: UInt32 = 100
+    /// Recommend: 100ms
+    /// Minimum: 10ms
+    public let samplingIntervalMs: UInt32
+
+    /// Minimum sampling interval when recovering from drift (ms).
+    /// Prevents back-to-back sampling from starving the main thread.
+    /// Recommend: around 10% of samplingIntervalMs
```
fnewberg: The "around 10%" recommendation is fine but doesn't state the actual constraint. `isValid` (`:46`) requires `minSamplingIntervalMs < samplingIntervalMs / 2`, so the real upper bound is 50%, not 10%. State the constraint explicitly so callers know what's enforced:

```suggestion
    /// Minimum sampling interval when recovering from drift (ms).
    /// Prevents back-to-back sampling from starving the main thread.
    /// Must be > 0 and strictly less than `samplingIntervalMs / 2`.
    /// Recommend: around 10% of `samplingIntervalMs`.
```

Sources/EmbraceProfiling/ProfilingConfiguration.swift
```diff
@@ -1,15 +1,52 @@
...
+//  Copyright © 2026 Embrace Mobile, Inc. All rights reserved.
 //
 
-/// Internal configuration for the profiling engine.
-struct ProfilingConfiguration {
+/// Configuration for the profiling engine.
+///
+/// All properties have sensible defaults. Pass a custom configuration to
+/// ``ProfilingEngine/start(configuration:)`` to override.
+public struct ProfilingConfiguration: Sendable {
```
fnewberg: All stored properties are `Equatable`/`Hashable`-conformant `UInt32`s, so synthesis is free. Matches the convention used by `ProfilingSample` and `ProfilingResult` and lets consumers cache or compare configurations:

```suggestion
public struct ProfilingConfiguration: Equatable, Hashable, Sendable {
```

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: I've changed it so both constants are now in nanoseconds (`gateSleepNs` and `gateTimeoutNs`). `usleep` takes the converted value.

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: `min_frames` is now a configurable field on `ProfilingConfiguration` (default 3). `fallback_walker` will be wired up in a follow-up PR once KSCrash 2.6.0 ships.

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: Done

Sources/EmbraceProfilingSampler/source/emb_sampler.c
kstenerud: This one is intentionally not faulted. `emb_stack_walk` failure produces `frame_count = 0`, which causes the sample to be skipped by the `if (frame_count > 0)` guard below. This is a transient condition (e.g. thread caught mid-prologue) and not worth aborting the session over. We can't log here either because we're in the async-safe section.

Sources/EmbraceProfilingSampler/source/emb_sampler.c
kstenerud: Yes indeed!

`max_frames` clamping now happens at the top of `emb_sampler_start`, before the idempotency `memcmp`, so a second call with the same original config correctly returns `OK`.

Sources/EmbraceProfilingSampler/source/emb_sampler.c
kstenerud: Done

Sources/EmbraceProfilingSampler/include/emb_sampler.h
kstenerud: Added them all

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: Added an `isActive` computed property (wrapping `emb_sampler_is_active()`) and updated the `stop()` doc to direct callers there when they need to know the worker is fully done (e.g. before destroying the ring buffer). `isCapturing` semantics are unchanged (it means "actively capturing samples right now.")

Sources/EmbraceProfilingSampler/include/emb_sampler.h
kstenerud: Not really... `stop()` is fire-and-forget by design. Callers who need to know when the worker is done poll `isActive`. The CAS result from `sampler_cas_multi` doesn't add actionable information: if the CAS succeeded the caller still has to poll, and if it failed (already stopping, faulted, etc.) the state is already observable.

Sources/EmbraceProfilingSampler/source/emb_sampler.c
kstenerud: added to `emb_sampler_reset_for_testing`.

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: The loop already tries before sleeping, so the first iteration after deadline expiry naturally gets one last attempt before the deadline check fires. I've removed that part of the comment so it doesn't get confusing.

Sources/EmbraceProfilingSampler/source/emb_sampler.c
kstenerud: Updated

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: Updated the comment. It's `gateIsAcquired.deallocate()` now.

Sources/EmbraceProfilingSampler/source/emb_sampler.c
kstenerud: The KSCrash fallback walker contract only returns values >= 0. Negative is not a documented error condition, just a defensive guard in our code against misuse. Both `< 0` and `0` correctly result in a skipped sample, which is the right behavior in both cases. Updated the `@return` doc on `emb_fallback_stack_walker_fn` to reflect the actual contract.

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: Added

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: Added

Sources/EmbraceProfilingSampler/source/emb_sampler.c
kstenerud: Added `test_stop_whenFaulted_isNoOp` and `test_isCapturing_falseWhenFaulted`. A direct test for the write-failure fault path requires forcing a ring buffer write to fail while the sampler is live. We'd need to add some kind of injection point to test that.

Tests/EmbraceProfilingSamplerTests/SamplerTests.swift
kstenerud: Tightened to `>= 15` frames, which cannot be satisfied by a normal idle main thread stack but is comfortably met by the 20-deep test thread recursion.

Tests/EmbraceProfilingSamplerTests/SamplerTests.swift
kstenerud: Added `test_stop_whenFaulted_isNoOp` and `test_isCapturing_falseWhenFaulted`. An existing test already checks for false when faulted.

Sources/EmbraceProfiling/ProfilingSample.swift
kstenerud: `ProfilingSample.init` and `ProfilingResult.init` are now internal.

Sources/EmbraceProfiling/ProfilingSample.swift
kstenerud: done

Sources/EmbraceProfilingSampler/include/emb_sampler.h
kstenerud: Fixed

Sources/EmbraceProfilingSampler/include/emb_sampler.h
kstenerud: Added a second paragraph to the `emb_sampler_start` doc covering the background-thread / dlopen case.

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: done

Sources/EmbraceProfiling/ProfilingConfiguration.swift
kstenerud: done

Sources/EmbraceProfiling/ProfilingConfiguration.swift
kstenerud: done

kstenerud:
> Do we have any tests that exercise reading while the writer is running? The current `retrieveSamples` tests appear to only exercise the API after `stop`.

We have the following read-while-writing tests:
- `test_retrieveSamples_whileRunning_succeeds` retrieves while the sampler is actively writing and asserts both that samples are present and that timestamps are in order.
- `test_concurrentRetrieveWhileRunning_noCorruption` goes further: it hammers 50 concurrent retrieveSamples calls against the live writer and verifies no timestamp inversions or out-of-bounds frame ranges. Both tests call stop() only after the assertions.

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -14,54 +15,361 @@
...
+            let config = emb_sampler_config_t(
+                sampling_interval_ms: configuration.samplingIntervalMs,
+                min_sampling_interval_ms: configuration.minSamplingIntervalMs,
+                max_frames: configuration.maxFramesPerSample,
+                min_frames: configuration.minFramesPerSample,
+                fallback_walker: nil,
+                start_paused: configuration.startPaused
+            )
+
+            switch emb_sampler_start(ringBuffer, config) {
```
ferdraghi: do `ringBuffer` and `readBuffer` need to be deallocated if this method returns anything other than ok? I'm not sure if there could be a scenario where `needsNewBuffer` is true and then this method doesn't return ok though...

ferdraghi (approved):

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: It will leak, but only temporarily (until the next call to `start`). I'll tighten it up though so it doesn't trip up a refactor later.

fnewberg (approved):

PR #586, opened 2026-05-06
[EMBR-12006] Feature: Adding UserSessionController

NachoEmbrace:
NOTE: https://github.com/embrace-io/embrace-apple-sdk/pull/613 should be merged into this

Introduces the in-memory user-session controller that owns the lifecycle decision tree and the cold-start reconstruction from the latest persisted `SessionRecord`. Session parts (`EmbraceSession`) now route through the controller at start time so each part is attached to the correct user session, and the `emb.session.upload_index` counter increments once per user-session creation rather than once per part.

No public-facing behavior change beyond the `userSession*` columns being populated on every new `SessionRecord` and the counter incrementing on a different cadence. The manual `endUserSession` API and the heartbeat-driven max-duration detector land in a follow-up.

## Architectural changes

- New `UserSessionController` with `bootstrap()`, `attachPart(state:startTime:)`, `markForegroundPartEnded(at:)`, `endActiveUserSession(reason:at:)`, `canManuallyEnd(now:)`. Single-mutex state (snapshot + manual-end timestamp); injected `dateProvider` for tests.
- `SessionController` calls `attachPart` before inserting the part record and stamps the resulting user-session columns. Calls `markForegroundPartEnded` on foreground-part end. Removed the per-part counter increment.
- `Embrace.init` constructs the controller and calls `bootstrap()` before `sessionLifecycle.startSession()` and `UnsentDataHandler.sendUnsentData` — the previous process's last part is the source for reconstruction and would otherwise be uploaded and deleted first.
- New notifications on `UserSessionController`: `embraceUserSessionDidStart`, `embraceUserSessionDidEnd` (both async on main, payload is the `EmbraceUserSession` snapshot).
- Renamed `SessionController` notifications to reflect that they're part-level: `embraceSessionDidStart` → `embraceSessionPartDidStart`, `embraceSessionWillEnd` → `embraceSessionPartWillEnd`. Internal consumers updated.
- Renamed `UserSession` (the defaults struct in `EmbraceSemantics`) → `UserSessionSemantics`


Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -13,21 +13,28 @@ import Foundation
 #endif
 
 extension Notification.Name {
-    public static let embraceSessionDidStart = Notification.Name("embrace.session.did_start")
-    public static let embraceSessionWillEnd = Notification.Name("embrace.session.will_end")
+    /// Posted just after a session part starts. The `object` is the newly-created `EmbraceSession`
+    /// (the part). For user-session lifecycle, listen to `embraceUserSessionDidStart` instead.
+    public static let embraceSessionPartDidStart = Notification.Name("embrace.session.part.did_start")
```
fnewberg, line 18: `Examples/EmbraceIOTestApp/EmbraceIOTestApp/Tests/UploadedPayloads/ViewModels/UploadedSessionPayloadTestViewModel.swift:77,82` still references the removed/renamed `.embraceSessionDidStart` / `.embraceSessionWillEnd`

Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -13,21 +13,28 @@ import Foundation
...
+
+    /// Posted just before a session part ends. The `object` is the `EmbraceSession` (the part)
+    /// that is about to end. For user-session lifecycle, listen to `embraceUserSessionDidEnd`.
+    public static let embraceSessionPartWillEnd = Notification.Name("embrace.session.part.will_end")
 }
 
-/// The source of truth for sessions. Provides the CRUD functionality for a given EmbraceSession
+/// The source of truth for session parts. A "part" is one contiguous foreground/background
```
fnewberg: nitpick: to avoid this being read as "combined foreground and background"

```suggestion
/// The source of truth for session parts. A "part" is one contiguous foreground or background
```

Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -166,7 +177,13 @@ class SessionController: SessionControllable {
                 spanId: span.context.spanId,
                 startTime: startTime,
                 coldStart: isColdStart,
-                sessionNumber: sessionNumber
+                sessionNumber: sessionPartNumber,
```
fnewberg, line 275: should we rename `sessionNumber` to avoid confusion?

Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -243,15 +260,23 @@ class SessionController: SessionControllable {
             // )
         }
 
-        // update session end time and clean exit
+        // update session end time and clean exit. For foreground parts, also record the
```
fnewberg: explicit call out whether this is updating `session part end time...` or `user session end time...`

Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -243,15 +260,23 @@ class SessionController: SessionControllable {
...
+        // foreground-end timestamp on the part record AND on the in-memory user-session
+        // snapshot so the next `attachPart` can compute the inactivity cutoff.
+        let isForeground = inProgressSession.state == SessionState.foreground
         let sessionToUpload: EmbraceSession? = storage?.updateSession(
             session: inProgressSession,
             endTime: now,
-            cleanExit: true
+            cleanExit: true,
+            userSessionLastForegroundEnd: isForeground ? now : nil
```
fnewberg, line 372:
```suggestion
            // nil will not overwrite previously-set userSessionLastForegroundEnd value
            userSessionLastForegroundEnd: isForeground ? now : nil
```

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,309 @@
...
+    /// the active user session — its `lastForegroundPartEnd` is cleared (no inactivity cutoff
+    /// applies during a part) and `partIndex` is bumped.
+    ///
+    /// - Parameters:
+    ///   - state: The state of the part being created (foreground/background).
+    ///   - startTime: The wall-clock start time of the new part.
+    /// - Returns: The user-session snapshot the new part will belong to.
+    @discardableResult
+    func attachPart(state: SessionState, startTime: Date) -> EmbraceUserSession {
+        return _state.withLock { state in
```
fnewberg: we are shadowing the `state` arg here. nothing in the function uses the `state: SessionState` arg. should it be used or should we remove the arg? e.g. should `snapshot.cleared()` be gated on this being a foreground session part?

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,309 @@
...
+/// - `inactivityTimeout`: time after the most recent foreground-part end before the user
+///   session expires.
+///
+/// The controller is the single source of truth for "which user session does this part belong to?".
+/// It is consumed by `SessionController` (which calls `attachPart` before creating a part record)
+/// and exposes `bootstrap()` for cold-start reconstruction from the latest persisted `SessionRecord`.
+///
+/// All cutoff math uses wall-clock `Date()`; `dateProvider` is injected for tests.
+///
+/// - Important: `bootstrap()` MUST run before unsent data is uploaded — the previous process's
```
fnewberg, line 41: This lists only `sendUnsentData` as an ordering hazard, but `Embrace.swift:172` names two: `sessionLifecycle.startSession()` is the other one — it routes into `attachPart`, which sees an empty snapshot and starts a fresh user session, losing the persisted one before `bootstrap()` ever reads it.  Expand this comment to cover both.

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,309 @@
...
+                return startNewUserSession(state: &state, at: startTime)
+            }
+
+            // Active user session should be terminated -> end current and start new one
+            if let reason = Self.expiryReason(snapshot: snapshot, at: startTime) {
+                internalEndUserSession(snapshot: snapshot, reason: reason, at: startTime)
+                return startNewUserSession(state: &state, at: startTime)
+            }
+
+            // Active userr session still valid -> update
```
fnewberg:
```suggestion
            // Active user session still valid -> update
```

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,309 @@
...
+        }
+    }
+
+    // MARK: - Manual-end rate limit
+
+    /// Returns `true` if a manual end-user-session call is allowed at `now` (i.e., at least 5s
+    /// have elapsed since the last allowed manual end). Records `now` as the last allowed call
+    /// when it returns `true`. The 5-second window prevents customers from overwhelming the
+    /// backend by calling the manual end API in a tight loop.
+    func canManuallyEnd(now: Date) -> Bool {
```
fnewberg, line 224: should we leverage `dateProvider` in the function rather than passing in `now`?

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,309 @@
...
+            maxDuration: maxDuration,
+            inactivityTimeout: inactivityTimeout,
+            lastForegroundPartEnd: latest.userSessionLastForegroundEnd,
+            partIndex: max(latest.userSessionPartIndex, 1),
+            endTime: nil,
+            terminationReason: nil
+        )
+
+        if Self.expiryReason(snapshot: snapshot, at: dateProvider()) != nil {
+            return
```
fnewberg, line 126: When the reconstructed snapshot is already expired, this returns silently — no `embraceUserSessionDidEnd` is posted, and `backfillTerminationReasonOnLatestPart` is never called for
the prior part. Mid-session expiry (via `attachPart` at L152-156) goes through `internalEndUserSession` and gets both. Cold-start expiry doesn't, so the same user-session boundary
produces different telemetry depending on whether the app happened to be running when it crossed.

If this is intentional — e.g., the prior process was expected to deliver these and a cold-start replay would just double-count — note this in the doc-comment at L84-86 . Otherwise, route this path through `internalEndUserSession` with the computed reason so the two expiry paths match.

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /// Stamps the termination reason on the just-closed part record (if any) and posts the
+    /// internal notification. Caller holds the mutex.
+    private func internalEndUserSession(
+        snapshot: ImmutableUserSession,
+        reason: TerminationReason,
+        at endTime: Date
+    ) {
+        sessionController?.backfillTerminationReasonOnLatestPart(reason)
```
fnewberg, line 269: [from Claude] 

Called while holding `_state`'s mutex (via `attachPart` / `endActiveUserSession`). Today's no-op stub makes this safe, but once the stub starts mutating `SessionController` state it will need `SessionController.lock` (an `UnfairLock`, i.e. non-reentrant `os_unfair_lock`). Two failure modes appear:

 1. `startSession` path: A (`SessionController.lock`) → B (`_state`) → A inside B = same-thread re-acquire of a non-reentrant lock → UB/hang.
 2. `endActiveUserSession` path: B → A combined with `startSession`'s A → B prefix = classic lock-order inversion across threads.

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,309 @@
...
+
+    // MARK: - Manual-end rate limit
+
+    /// Returns `true` if a manual end-user-session call is allowed at `now` (i.e., at least 5s
+    /// have elapsed since the last allowed manual end). Records `now` as the last allowed call
+    /// when it returns `true`. The 5-second window prevents customers from overwhelming the
+    /// backend by calling the manual end API in a tight loop.
+    func canManuallyEnd(now: Date) -> Bool {
+        _state.withLock { state in
+            if let last = state.lastManualEnd, now.timeIntervalSince(last) < 5.0 {
```
fnewberg: nitpick: define a constant for this so it feels less like a magic number

Tests/EmbraceCoreTests/Session/SessionControllerTests.swift
```diff
@@ -447,52 +458,61 @@ final class SessionControllerTests: XCTestCase {
...
 
-        // then sessionNumber continues from 6
         XCTAssertEqual(session?.sessionNumber, 6)
 
-        let resource = storage.fetchRequiredPermanentResource(key: SessionController.sessionNumberKey)
+        let resource = storage.fetchRequiredPermanentResource(key: SessionController.sessionPartNumberKey)
         XCTAssertEqual(resource?.value, "6")
     }
 
+    func test_startSession_persistsUserSessionColumnsOnRecord() throws {
```
fnewberg, line 858: any other tests worth adding, e.g. checking session part number behavior number across user sessions?

Tests/EmbraceCoreTests/Session/UserSessionControllerTests.swift
```diff
@@ -0,0 +1,318 @@
...
+        wait(for: [drained], timeout: 1)
+
+        XCTAssertEqual(startCount, 1, "Joining an existing user session must not post a start notification")
+    }
+
+    func testAttachPart_expiredByMaxDuration_endsAndCreatesNew() {
+        let controller = makeController()
+        let first = controller.attachPart(state: .foreground, startTime: now)
+
+        let endNotification = expectation(forNotification: .embraceUserSessionDidEnd, object: nil)
```
fnewberg: check the notification payload, not that we just got a notification. also applies to L119 and L134

```suggestion
        let endNotification = expectation(forNotification: .embraceUserSessionDidEnd, object: nil)
        let endNotification = expectation(forNotification: .embraceUserSessionDidEnd, object: nil) { notif in
            guard let session = notif.object as? EmbraceUserSession else { return false }
            XCTAssertEqual(session.id, first.id)
            XCTAssertEqual(session.terminationReason, .maxDurationReached)
            XCTAssertNotNil(session.endTime)
            return true
       }
```

Sources/EmbraceSemantics/UserSession/UserSessionSemantics.swift
```diff
@@ -4,7 +4,7 @@
 
 import Foundation
 
-public struct UserSession {
+public struct UserSessionSemantics {
```
fnewberg, line 7: nitpick: i think this could be an enum
```suggestion
public enum UserSessionSemantics {
```

Sources/EmbraceCore/Session/UserSessionController.swift
fnewberg: i think this may have been addressed in #613

Sources/EmbraceCore/Session/UserSessionController.swift
fnewberg: Follow-up addressed on #613 — leaving a concrete comment there now that `backfillTerminationReasonOnLatestPart` exists to wire into.

Sources/EmbraceCore/Session/SessionController.swift
NachoEmbrace: Test apps are not addressed on these PRs, will be done later on separated PRs.

Sources/EmbraceCore/Session/SessionController.swift
NachoEmbrace: Like I said in other PRs, if we're doing this I'd rather do all of it together in one single PR (explicitly renaming session to session part everywhere)

Sources/EmbraceCore/Session/UserSessionController.swift
NachoEmbrace: https://github.com/embrace-io/embrace-apple-sdk/pull/613 addresses this

Sources/EmbraceCore/Session/SessionController.swift
fnewberg: yeah, good point

Sources/EmbraceCore/Session/UserSessionController.swift
NachoEmbrace: We can't because the date is being passed to 2 function calls and we want it to be the same
I will, however, clean up Embrace.endUserSession so it's cleaner.

Sources/EmbraceCore/Session/UserSessionController.swift
NachoEmbrace: Good catch

Tests/EmbraceCoreTests/Session/SessionControllerTests.swift
NachoEmbrace: I added a lot more tests, including ones around session part number / index

fnewberg:
@NachoEmbrace should we still have `emb.session.upload_index` [here](https://github.com/embrace-io/embrace-apple-sdk/blob/f92b966d4d679db17211baaa973f034bd277acef/Examples/EmbraceIOTestApp/EmbraceIOTestApp/Tests/Startup/Tests/MetadataResourceTest.swift#L38)? seems like we've moved on from that key elsewhere

Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -13,21 +13,28 @@ import Foundation
...
-///     Just after a session starts. See ``Notification.Name.embraceSessionDidStart``
-///     Just before a session will end. See ``Notification.Name.embraceSessionWillEnd``
+///     Just after a part starts. See ``Notification.Name.embraceSessionPartDidStart``
+///     Just before a part will end. See ``Notification.Name.embraceSessionPartWillEnd``
 ///
 
 class SessionController: SessionControllable {
 
-    static let sessionNumberKey = "emb.session.upload_index"
+    static let sessionPartNumberKey = "emb.session_part_number"
```
fnewberg, line 37: since we rename the key here, we will reset this counter. not the end of the world, but how would you feel about adding something like this?

```
  private static let legacyCounterKey = "emb.session.upload_index"

  private func migrateLegacyCounterIfNeeded(storage: EmbraceStorage) {
      // If the new key already has a row, migration already ran (or this is a
      // fresh install that started on 7.0). Skip.
      guard storage.fetchRequiredPermanentResource(key: Self.counterKey) == nil else {
          return
      }
      guard let legacy = storage.fetchRequiredPermanentResource(key: Self.legacyCounterKey),
            let legacyValue = EMBInt(legacy.value)
      else {
          return  // Genuine fresh install or unparsable legacy row — start at 1 via normal path.
      }
      storage.setPermanentResource(key: Self.counterKey, value: String(legacyValue))
      // Optional: also delete the legacy row to keep the metadata table clean.
      // storage.deletePermanentResources(keys: [Self.legacyCounterKey])
  }
```

Sources/EmbraceCore/Capture/CaptureServices.swift
```diff
@@ -179,6 +189,10 @@ final class CaptureServices {
         for service in services { service.onSessionWillEnd() }
     }
 
+    @objc func onUserSessionDidEnd(notification: Notification) {
+        crashReporter?.currentUserSessionId = nil
```
fnewberg, line 205: i think we have a race condition on account of the following
 
- `UserSessionController` posts both user-session notifications on the main queue asynchronously:
  - start: `Sources/EmbraceCore/Session/UserSessionController.swift:242`
  - end: `Sources/EmbraceCore/Session/UserSessionController.swift:258`
- `SessionController` posts part-start synchronously right after session creation:
  - `Sources/EmbraceCore/Session/SessionController.swift:287`
- `CaptureServices` sets `currentUserSessionId` on part-start, but clears it on user-session-end:
  - set: `Sources/EmbraceCore/Capture/CaptureServices.swift:183`
  - clear: `Sources/EmbraceCore/Capture/CaptureServices.swift:193`
  - comment claiming safe repopulation: `Sources/EmbraceCore/Capture/CaptureServices.swift:109`

so on rollover, the new `currentUserSessionId` can be written first (sync part-start) and then overwritten to `nil` later when the queued `.embraceUserSessionDidEnd` handler runs. that creates a user session attribution gap until another part-start occurs.

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,347 @@
...
+    /// If the latest part has no user-session columns (legacy v6 row) the snapshot is left
+    /// empty and the next `attachPart` call starts a new user session from scratch.
+    ///
+    /// If the reconstructed user session is already expired by the time bootstrap runs, the
+    /// termination is routed through `internalEndUserSession` — same path as a mid-session
+    /// expiry detected by `attachPart`. The termination reason is backfilled onto the prior
+    /// part record (no-op if the prior process already stamped one) and the user-session-end
+    /// notification fires, so cold-start expiry produces the same telemetry as mid-session
+    /// expiry instead of silently dropping the snapshot.
+    func bootstrap() {
```
fnewberg: is there a compelling reason for this to not be part of `init()` given the statements we make about this having to run before certain things? e.g. could we do something like this

1. .call a  helper from `init`, e.g. `reconstructFromStorage(...)`, that returns an optional snapshot.
2. initialize `_state` from the helper functions result.
3. remove public `bootstrap()` and its ordering contract from callers.
4. inject the dependency needed for termination-reason backfill up front so reconstruction/expiry paths don’t rely on post-init weak wiring.

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,347 @@
...
+            startTime: startTime,
+            maxDuration: maxDuration,
+            inactivityTimeout: inactivityTimeout,
+            lastForegroundPartEnd: lastForegroundPartEnd,
+            partIndex: partIndex,
+            endTime: endTime,
+            terminationReason: reason
+        )
+    }
+}
```
fnewberg, line 361: * This call runs while `_state` is held. 
* `attachPart` acquires `_state.withLock` at `UserSessionController.swift:154`, and the expiry branch calls `internalEndUserSession` before releasing it (`UserSessionController.swift:163`). 
* `backfillTerminationReasonOnLatestPart` performs two CoreData operations (`fetchLatestSession` + `updateSession`), so the mutex can be held for the full storage round-trip.

Concurrent readers of `currentUserSession` — especially the heartbeat path in `checkUserSessionMaxDurationExpiry` and the background-split check in `applyBackgroundSplitIfNeeded` — both call `userSessionController?.currentUserSession` and contend on this same lock. They can block for the duration of CoreData I/O.

Suggested fix: capture the termination inputs/snapshot inside the lock, release the lock, then perform `backfillTerminationReasonOnLatestPart` (and any related side effects) outside the lock.

Note: `endActiveUserSession` currently follows the same lock-held call path into `internalEndUserSession` (`UserSessionController.swift:200`–`205`), so both call sites should be updated.

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,347 @@
...
+            startTime: startTime,
+            maxDuration: maxDuration,
+            inactivityTimeout: inactivityTimeout,
+            lastForegroundPartEnd: nil,
+            partIndex: 1,
+            endTime: nil,
+            terminationReason: nil
+        )
+        state.session = snapshot
+        DispatchQueue.main.async {
```
fnewberg, line 252: This notification is enqueued before `SessionController` has called `storage.addSession()`. If `addSession` returns `nil`, the notification fires, claiming a new user session has started, but the part that would be its first record was never written.

`SessionController._session.session` is `nil` at that point, and the next successful part starts at `partIndex: 2`, silently skipping index 1.

`embraceSessionPartDidStart` handles this correctly — it's gated on `session != nil` after `addSession` returns (`SessionController.swift:286`). `embraceUserSessionDidStart` should follow the same pattern: remove the post from `startNewUserSession`, have `attachPart` signal to its caller
whether a new user session was started, and let `SessionController.startSession` post it only after `addSession` is confirmed non-nil.

Sources/EmbraceCore/Session/SessionController.swift
NachoEmbrace: I'm not a big fan of this simply because it'll be a fetch that runs on startup every time
If we still think keeping the number is worth it then sure

Sources/EmbraceCore/Capture/CaptureServices.swift
NachoEmbrace: Addressed by using the user session start notification instead
I also changed the part notifications to be dispatched async on the main queue, for consistency

Sources/EmbraceCore/Session/UserSessionController.swift
NachoEmbrace: What's the reason for the change?
This is all internal code, there's no risk.

fnewberg: feels safer to have the behavior enforced in code rather than stating things like "must run before X" in comments

Sources/EmbraceCore/Session/SessionController.swift
fnewberg: not a huge deal. we're really looking more at numbers being consecutive than absolute values

Sources/EmbraceCore/Session/UserSessionController.swift
NachoEmbrace: What call are you referring too exactly? The comment is highlighting the entire file...

Sources/EmbraceCore/Session/UserSessionController.swift
NachoEmbrace: This is not entirely true.
addSession will always return something, regardless of what the internal async core data operation does.
That being said, it is true that the in-memory state can get out of sync if the core data operation fails.
This is not something that was introduced here, it's a compromise that we've been living with for a long time.
If we're addressing this, it will be in a different PR.

Sources/EmbraceCore/Session/UserSessionController.swift
fnewberg: sorry, this comment should have been anchored to line 255

Sources/EmbraceCore/Session/UserSessionController.swift
NachoEmbrace: The only reason it's done this way is because both UserSessionController and SessionController reference each other.
bootstrap can't be called until both references are set.
There are ways out of this but I don't think the resulting code will be cleaner.
So given that this is all internal mechanisms of the SDK, I think it's fine to leave it this way even if it's not a very elegant architecture.

NachoEmbrace:
> @NachoEmbrace should we still have `emb.session.upload_index` [here](https://github.com/embrace-io/embrace-apple-sdk/blob/f92b966d4d679db17211baaa973f034bd277acef/Examples/EmbraceIOTestApp/EmbraceIOTestApp/Tests/Startup/Tests/MetadataResourceTest.swift#L38)? seems like we've moved on from that key elsewhere

nope, will remove

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -0,0 +1,364 @@
...
+    func bootstrap() {
+        guard let storage = storage else {
+            return
+        }
+
+        let latest = storage.fetchLatestSession()
+        guard
+            let latest = latest,
+            let userSessionId = latest.userSessionId,
+            let userSessionStartTime = latest.userSessionStartTime
```
ferdraghi, line 103: Is it worth guarding that `userSessionTerminationReason` is nil here since there's a reason already set so we don't need to add one? (unless I misunderstood what happens after the guard here)

ferdraghi (approved):

Sources/EmbraceCore/Session/UserSessionController.swift
NachoEmbrace: This is the code that restores the last user session, so there wont be a termination reason.
At the end of the function it is checking the user session should expire and sets the reason there.
If the app crashed, that will be handled in UnsentDataHandler a bit later, and the termination reason is updated to crashed.

PR #587, opened 2026-05-06
[EMBR-12039] Feature: Removing OSLogStore usage to generate critical logs file

NachoEmbrace:
Replaces `OSLogStore`-based critical log export with direct file writes. The previous design used `os_log` to write entries and then synchronously read them back through `OSLogStore.getEntries`, which under memory pressure tripped a recursive `NSException`/`couldNotInstantiate` crash and a synchronous-XPC hang vector.

## Changes

- **Two-file state machine**: `pending-logs` receives `.startup` lines; promoted via atomic `rename(2)` to `critical-logs` on the first `.critical`. No file is left behind for healthy runs (requirement: only generate the upload file when something actually went wrong).
- **POSIX writes** (`open`/`write`/`fsync`/`close`) instead of `FileHandle`. Avoids `NSException` on failure (Swift can't catch it), drops the iOS 13.4+ availability gate, and removes the Foundation/ObjC bridge from the hot path.
- **`fsync` on `.critical`** for crash durability; skipped on `.startup` to keep boot cheap.
- **Boot-time orphan cleanup** in `UnsentDataHandler.sendCriticalLogs` deletes any leftover `pending-logs`.
- **Removed**: `OSLogStore` usage, the `com.embrace.logger.export` queue, `lastExportDate`/`exporting` cursor state, the iOS 15 / tvOS 15 availability gate, and the dual `OSLog` channel split.
- Tests rewritten around file-state assertions (no more `OSLogStore` reads).

### DIFF
--- a/Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
+++ b/Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
@@ -96,7 +96,6 @@ class DefaultInternalLogger: BaseInternalLogger {
             let fd = state.fileDescriptor
             guard fd >= 0 else { return }
 
-  
             let available = customExportByteCountLimit - state.bytesWritten
             if available <= 0 {
                 state.limitReached = true
@@ -104,7 +103,7 @@ class DefaultInternalLogger: BaseInternalLogger {
             }
             let chunk = data.prefix(available)
 
-            let written = data.withUnsafeBytes { (buffer: UnsafeRawBufferPointer) -> Int in
+            let written = chunk.withUnsafeBytes { (buffer: UnsafeRawBufferPointer) -> Int in
                 guard let base = buffer.baseAddress else { return -1 }
                 return write(fd, base, buffer.count)
             }
@@ -115,6 +114,10 @@ class DefaultInternalLogger: BaseInternalLogger {
             }
 
             state.bytesWritten += written
+            if chunk.count < data.count {
+                // we just wrote up to the cap; further writes are no-ops
+                state.limitReached = true
+            }
 
             if level == .critical {
                 _ = fsync(fd)
--- a/Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
+++ b/Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
@@ -2,6 +2,7 @@
 //  Copyright © 2024 Embrace Mobile, Inc. All rights reserved.
 //
 
+import Darwin
 import Foundation
 import OSLog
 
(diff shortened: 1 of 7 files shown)


Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
```diff
@@ -13,132 +14,172 @@ import OSLog
...
+            }
+
+            if written < 0 {
+                print("Error writing critical logs file: errno \(errno)")
                 return
             }
 
-            self.exporting = true
-            defer { self.exporting = false }
+            state.bytesWritten += written
```
fnewberg: while unlikely in most cases, `write()` can return fewer bytes than requested. but we're only in this code path because the app is logging `.critical`, so things may not be that healthy when we hit this code it might make sense to cover that case. might be worthwhile handling that case

```suggestion
            let written = data.withUnsafeBytes { (buffer: UnsafeRawBufferPointer) -> Int in
                guard let base = buffer.baseAddress else { return -1 }
                var offset = 0
                while offset < buffer.count {
                    let n = write(fd, base.advanced(by: offset), buffer.count - offset)
                    if n < 0 {
                        if errno == EINTR { continue }
                        return offset > 0 ? offset : -1
                    }
                    if n == 0 { break }
                    offset += n
                }
                return offset
            }
    
            if written < 0 {
                print("Error writing critical logs file: errno \(errno)")
                return
            }
            
            state.bytesWritten += written
```

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
```diff
@@ -13,132 +14,172 @@ import OSLog
...
+                openFile(at: url, state: &state)
+            }
+
+            let fd = state.fileDescriptor
+            guard fd >= 0 else { return }
+
+            guard state.bytesWritten + data.count <= customExportByteCountLimit else {
+                state.limitReached = true
                 return
             }
```
fnewberg: this would drop an initial long line. truncate instead of dropping it

```suggestion
  
            let available = customExportByteCountLimit - state.bytesWritten
            if available <= 0 {
                state.limitReached = true
                return
            }
            let chunk = data.prefix(available)
```

Tests/EmbraceCoreTests/Internal/Logs/Loggers/DefaultInternalLoggerTests.swift
```diff
@@ -6,164 +6,199 @@ import EmbraceCommonInternal
...
             }
         }
+
+        group.wait()
+
+        // After at least one .critical the file must exist
+        XCTAssertTrue(FileManager.default.fileExists(atPath: criticalURL.path))
+
+        let dump = (try? String(contentsOf: criticalURL)) ?? ""
+        // every line must be terminated; trailing newline means split keeps empty element
```
fnewberg, line 271: also add an assert that you get the number of lines you expected

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
```diff
@@ -13,132 +14,172 @@ import OSLog
...
-    init(exportFilePath: URL?, exportByCountLimit: Int = 1000, exportCategory: String = "custom-export") {
-        self.exportFilePath = exportFilePath
-        self.customExportByteCountLimit = exportByCountLimit
-        self.customExportCategory = exportCategory
+    private let state = EmbraceMutex(State())
 
-        defaultLogger = OSLog(subsystem: subsystem, category: defaultCategory)
-        customExportLogger = OSLog(subsystem: subsystem, category: customExportCategory)
+    init(
+        pendingFilePath: URL?,
```
fnewberg, line 54: is there a use case for where these values are not set?

Tests/EmbraceCoreTests/Internal/Logs/Loggers/DefaultInternalLoggerTests.swift
```diff
@@ -6,164 +6,199 @@ import EmbraceCommonInternal
...
+        XCTAssertFalse(FileManager.default.fileExists(atPath: pendingURL.path))
     }
-}
 
-extension DefaultInternalLogger {
-    func waitForQueue() async {
-        await withCheckedContinuation { continuation in
-            queue.async {
-                continuation.resume()
+    /// Once the byte budget is exhausted, further writes are dropped — file size is bounded.
```
fnewberg: add a test that tests the suggested line truncation, if that is adopted

fnewberg (commented):
looks good. i made a number of suggestions, but the only one i feel strongly about is truncating a long initial line.

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
NachoEmbrace: No, it's passed by injection for testing purposes
In the real world it will always have a value

fnewberg (approved):

Sources/EmbraceCore/Internal/Logs/Loggers/DefaultInternalLogger.swift
NachoEmbrace: Per our discussion we are not doing this (we don't really need this level of defensiveness for the feature).

PR #601, opened 2026-05-11
[EMBR-11979] Fix: Update Traceparent handling

ferdraghi:


Sources/EmbraceCore/Capture/Network/HostAllowlistMatcher.swift
```diff
@@ -0,0 +1,27 @@
...
+    /// Returns whether `host` is permitted by `allowlist`.
+    ///
+    /// - `host` nil → false.
+    /// - `allowlist` empty → true (no filter; all hosts pass).
+    /// - Otherwise: case-insensitive match on equality (`"test.com"`) or subdomain
+    ///   (`"api.test.com"` for entry `"test.com"`). Entries are expected to be
+    ///   lowercase-normalized at Options-init time.
+    static func matches(host: String?, allowlist: [String]) -> Bool {
+        guard let host else { return false }
+        guard !allowlist.isEmpty else { return true }
```
fnewberg: nitpick: should this be the first check in the function? if `host` isn't set, arguably it doesn't matter, but if we want to be able to say that an empty `allowList` doesn't affect anything, that should probably be the first check

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService+Options.swift
```diff
@@ -19,18 +67,67 @@ extension URLSessionCaptureService {
...
             injectTracingHeader: Bool,
             requestsDataSource: URLSessionRequestsDataSource?,
-            ignoredURLs: [String]
+            ignoredURLs: [String],
+            traceparent: Traceparent = Traceparent()
         ) {
-            self.injectTracingHeader = injectTracingHeader
             self.requestsDataSource = requestsDataSource
             self.ignoredURLs = ignoredURLs
+            self.traceparent = traceparent
```
fnewberg, line 108: [claude]  Deprecated Options initializer silently discards injectTracingHeader

`URLSessionCaptureService+Options.swift`:91-100 — the deprecated init takes the old `injectTracingHeader`: parameter and drops it on the floor with no runtime signal. The one-shot deprecation log (`logDeprecatedInjectTracingHeaderOnce()`) is wired only into the property setter, so customers who migrate via the initializer path get only the compile-time `@available` warning and no log at runtime.

```suggestion
            self.traceparent = traceparent
            Options.logDeprecatedInjectTracingHeaderOnce()
```

Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig/RemoteConfigPayload.swift
```diff
@@ -135,20 +134,9 @@ public struct RemoteConfigPayload: Decodable, Equatable {
...
-            networkSpansForwardingThreshold =
-                try networkSpansForwardingContainer.decodeIfPresent(
-                    Float.self,
-                    forKey: CodingKeys.NetworkSpansForwardingCodingKeys.threshold
-                ) ?? defaultPayload.networkSpansForwardingThreshold
-        } else {
-            networkSpansForwardingThreshold = defaultPayload.networkSpansForwardingThreshold
-        }
+        // network span forwarding (new flat keys — old SDKs keep reading legacy nested key)
+        nsfThreshold = try rootContainer.decodeIfPresent(Float.self, forKey: .nsfThreshold)
```
fnewberg: should we be more defensive when decoding this and have a helper function like this to prevent errors in one or both these values from disrupting the overall parsing of the remote config?


```
  private static func decodeOptionalFloat(
      from container: KeyedDecodingContainer<CodingKeys>,
      forKey key: CodingKeys
  ) -> Float? {
      guard container.contains(key) else { return nil }
      if let value = try? container.decode(Float.self, forKey: key) {
          return value
      }
      Embrace.logger.warning(
          "Remote config key '\(key.rawValue)' is present but malformed; ignoring and falling back to default."
      )
      return nil
  }
```

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService+Options.swift
```diff
@@ -5,11 +5,59 @@
...
+                    )
+                } else if entry.hasPrefix(".") {
+                    Embrace.logger.warning(
+                        "allowedDomains entry '\(entry)' has a leading '.' — use a bare hostname like '\(entry.dropFirst())' instead; entry will be ignored."
+                    )
+                } else {
+                    result.append(entry.lowercased())
+                }
+            }
+            return result
```
fnewberg, line 63: should we add a warning or error if the final list is empty? if so, we should probably only call `validated` when the `allowedDomains` list passed in to `init` is not empty. we wouldn't want that log if users weren't specifying an allow list

Tests/EmbraceConfigInternalTests/Fixtures/remote_config.json
```diff
@@ -3,9 +3,8 @@
         "threshold": 75
     },
     "threshold": 50,
-    "network_span_forwarding": {
-        "pct_enabled": 25
-    },
+    "nsf_pct_enabled": 25,
```
fnewberg, line 6: this value should never be larger than `traceparent_injection_pct_enabled`. if there was no traceparent injected, there's no header to add to indicate NSF should be done

Sources/EmbraceCore/Capture/Network/HostAllowlistMatcher.swift
ferdraghi: yeah that's probably correct, if `allowList` must never be empty regardless of anything else, then yeah it should go first on the list. I'll fix it.

Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig/RemoteConfigPayload.swift
ferdraghi: good catch on the try failing here... although there's no need to add a helper function unless we want to explicitly log the error. If not, then we can simply do `try?` instead... If the decode fails (for example, an invalid value set on they key) then it will already return nil and since the properties are already optional, nothing will break. Let me know what you prefer.

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService+Options.swift
ferdraghi: Yeah it probably makes sense to add a warning if a non empty list was sent and then all of them were discarded. I'll add it.

Tests/EmbraceConfigInternalTests/Fixtures/remote_config.json
ferdraghi: this is just a mock config for testing purposes, you can disregard the values specified here, if anything, we should look at the tests instead that make sure the code executes as expected with the values specified here.

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService+Options.swift
```diff
@@ -5,11 +5,65 @@
...
+        /// are dropped at init time and a warning is logged. The SDK does not throw for bad entries.
+        ///
+        /// Empty list (default) = no allowlist filter; the header is sent on every captured
+        /// request that passes the other gates.
+        @objc public let allowedDomains: [String]
+
+        @objc public init(allowedDomains: [String] = []) {
+            self.allowedDomains = Traceparent.validated(allowedDomains)
+            super.init()
+        }
```
fnewberg: i'm wondering if we'd should do something like this and check `denyAll` in `shouldInjectHeader` to avoid the following scenario

- user defines one or more allowed domains
- all of the domains are invalid
- the domain list comes back as empty
- traceparents are injected into all requests when the intent was to limit to a subset of domains

```suggestion
        internal private(set) var denyAll = false

        @objc public init(allowedDomains: [String] = []) {
            let validated = Traceparent.validated(allowedDomains)
            if !allowedDomains.isEmpty && validated.isEmpty {
                self.allowedDomains = []
                self.denyAll = true
            } else {
                self.allowedDomains = validated
            }
            super.init()
        }
```

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService.swift
```diff
@@ -79,14 +80,16 @@ public final class URLSessionCaptureService: CaptureService, URLSessionTaskHandl
         }
     }
 
-    var injectTracingHeader: Bool {
-        // check remote config
-        guard Embrace.client?.config.isNetworkSpansForwardingEnabled == true else {
-            return false
-        }
+    func shouldInjectHeader(for request: URLRequest, span: Span) -> Bool {
```
fnewberg: remove unused `span` arg

Tests/EmbraceCoreTests/Capture/Network/HostAllowlistMatcherTests.swift
```diff
@@ -0,0 +1,69 @@
...
+
+final class HostAllowlistMatcherTests: XCTestCase {
+
+    // MARK: - nil host
+
+    func test_nilHost_ReturnsFalse() {
+        XCTAssertFalse(HostAllowlistMatcher.matches(host: nil, allowlist: ["test.com"]))
+    }
+
+    func test_nilHost_EmptyAllowlist_ReturnsFalse() {
```
fnewberg:
```suggestion
    func test_emptyAllowlist_ReturnsFalse() {
```

fnewberg (changes requested):

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService+Options.swift
ferdraghi: Not really sure a new variable is even needed at all here for this purpose, we could simply check that `allowedDomains` is not empty on the same place we would check for `denyAll` and achieve the same result. We're kinda doing that on `URLSessionCaptureService.swift` line 84 there's a guard waiting for the result of `HostAllowlistMatcher.matches` which receives this `allowedDomains` list as a parameter. However, this method would return `true` if `allowList` is empty... maybe we should change that to false? I mean if the list is empty then that would mean no domains are allowed?

Tests/EmbraceCoreTests/Capture/Network/HostAllowlistMatcherTests.swift
ferdraghi: now that you mention these tests, their behavior should be updated to reflect that an empty allowList with a valid host would return true, at least as it sits right now before addressing the comment above about adding a `denyAll` feature

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService+Options.swift
fnewberg: > maybe we should change that to false? I mean if the list is empty then that would mean no domains are allowed?

it is okay to not define any domains. in that case all domains should have traceparents injected. what we need to be able to tell the difference between is

- user specified no domains
- user specified domains, but they were all invalid

i don't know that we can tell that difference without adding a variable like `denyAll`, but i could be wrong

ferdraghi: Ah ok I see. There's a difference between and empty list and an invalid list. Yeah that makes sense then.

ferdraghi: Went for a different solution. Instead of adding another variable, forcing another not so trivial check, I made the list optional... a nil list means "do not apply any filters" and a not nil list means "only allow these domains" where the list being empty basically means "ignore everything".

fnewberg (approved):

PR #613, opened 2026-05-13
[EMBR-12007] Feature: Linking SesssionController with UserSessionController

NachoEmbrace:
Finishes the session logic and link the old SessionController with the new UserSessionController.

Note that this PR is pointing to https://github.com/embrace-io/embrace-apple-sdk/pull/586

### DIFF
--- a/Sources/EmbraceCore/Session/SessionController.swift
+++ b/Sources/EmbraceCore/Session/SessionController.swift
@@ -99,17 +99,20 @@ class SessionController: SessionControllable {
     /// cutoff, schedules a part-roll on the upload coordination queue so the work doesn't stall
     /// the heartbeat thread. The roll closes the current part, ends the user session with
     /// `.maxDurationReached`, and starts a new part with the same state.
-    private func checkUserSessionMaxDurationExpiry(now: Date) {
-        guard let controller = userSessionController,
-            let userSession = controller.currentUserSession
-        else {
-            return
-        }
-
-        guard now >= userSession.maxEnd else { return }
-
+    ///
+    /// The expiry decision runs INSIDE the dispatched block — not before — so that two ticks
+    /// queued back-to-back can't both roll. The second block sees the state left by the first
+    /// (the freshly-rotated user session has a far-future `maxEnd`) and bails.
+    func checkUserSessionMaxDurationExpiry(now: Date) {
         queue.async { [weak self] in
-            self?.rollPartForUserSessionExpiry(reason: .maxDurationReached, at: now)
+            guard let self = self,
+                let userSession = self.userSessionController?.currentUserSession,
+                now >= userSession.maxEnd,
+                self._session.safeValue.session != nil
+            else {
+                return
+            }
+            self.rollPartForUserSessionExpiry(reason: .maxDurationReached, at: now)
         }
     }
 
--- a/Sources/EmbraceCore/Session/SessionController.swift
+++ b/Sources/EmbraceCore/Session/SessionController.swift
@@ -106,8 +106,7 @@ class SessionController: SessionControllable {
             return
         }
 
-        let maxEnd = userSession.startTime.addingTimeInterval(userSession.maxDuration)
-        guard now >= maxEnd else { return }
+        guard now >= userSession.maxEnd else { return }
 
         queue.async { [weak self] in
             self?.rollPartForUserSessionExpiry(reason: .maxDurationReached, at: now)
@@ -133,9 +132,8 @@ class SessionController: SessionControllable {
             return false
         }
 
-        let maxEnd = userSession.startTime.addingTimeInterval(userSession.maxDuration)
         let inactivityCutoff = userSession.lastForegroundPartEnd?.addingTimeInterval(userSession.inactivityTimeout)
-        guard let cutoff = [maxEnd, inactivityCutoff].compactMap({ $0 }).min(),
+        guard let cutoff = [userSession.maxEnd, inactivityCutoff].compactMap({ $0 }).min(),
             now >= cutoff,
             cutoff > prev.startTime
         else {
--- a/Sources/EmbraceCore/Session/SessionController.swift
+++ b/Sources/EmbraceCore/Session/SessionController.swift
@@ -88,9 +88,78 @@ class SessionController: SessionControllable {
 
         self.heartbeat.callback = { [weak self] in
             let span = EmbraceMetricKitSpan.begin(name: "heartbeat")
-            self?.update(heartbeat: Date())
+            let now = Date()
+            self?.update(heartbeat: now)
+            self?.checkUserSessionMaxDurationExpiry(now: now)
             span.end()
         }
+    }
+
+    /// Triggered each heartbeat tick. If the active user session has crossed its max-duration
+    /// cutoff, schedules a part-roll on the upload coordination queue so the work doesn't stall
+    /// the heartbeat thread. The roll closes the current part, ends the user session with
+    /// `.maxDurationReached`, and starts a new part with the same state.
+    private func checkUserSessionMaxDurationExpiry(now: Date) {
+        guard let controller = userSessionController,
+            let userSession = controller.currentUserSession
+        else {
+            return
+        }
+
+        let maxEnd = userSession.startTime.addingTimeInterval(userSession.maxDuration)
+        guard now >= maxEnd else { return }
+
+        queue.async { [weak self] in
+            self?.rollPartForUserSessionExpiry(reason: .maxDurationReached, at: now)
+        }
+    }
+
+    /// If a foreground part is about to start after a background part whose user session
+    /// crossed its `maxDuration` or `inactivityTimeout` cutoff, slice the bg part at the
+    /// cutoff and synthesize a bg part `[cutoff, now]`. Returns `true` when the split was
+    /// applied — the caller must skip its own "end previous part" step because the prev part
+    /// is already closed.
+    @discardableResult
+    private func applyBackgroundSplitIfNeeded(
+        prev: EmbraceSession?,
+        newState: SessionState,
+        now: Date
+    ) -> Bool {
+        guard newState == .foreground,
+            let prev = prev,
+            prev.state == .background,
+            let userSession = userSessionController?.currentUserSession
+        else {
+            return false
+        }
+
+        let maxEnd = userSession.startTime.addingTimeInterval(userSession.maxDuration)
+        let inactivityCutoff = userSession.lastForegroundPartEnd?.addingTimeInterval(userSession.inactivityTimeout)
+        guard let cutoff = [maxEnd, inactivityCutoff].compactMap({ $0 }).min(),
+            now >= cutoff,
+            cutoff > prev.startTime
+        else {
+            return false
+        }
+
+        endSession(at: cutoff)
+        startSession(state: .background, startTime: cutoff)
+        endSession(at: now)
+        return true
+    }
+
+    /// Closes the current part at `now`, ends the user session with the given reason, then
+    /// starts a new part with the same `SessionState`. Used by the heartbeat-driven max-duration
+    /// detector and the manual `endUserSession()` API. The order — end-part → end-user-session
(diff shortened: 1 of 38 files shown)


Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -88,11 +88,80 @@ class SessionController: SessionControllable {
...
+    /// the heartbeat thread. The roll closes the current part, ends the user session with
+    /// `.maxDurationReached`, and starts a new part with the same state.
+    private func checkUserSessionMaxDurationExpiry(now: Date) {
+        guard let controller = userSessionController,
+            let userSession = controller.currentUserSession
+        else {
+            return
+        }
+
+        let maxEnd = userSession.startTime.addingTimeInterval(userSession.maxDuration)
```
fnewberg: nitpick: this seems knowable from the start of the session. should we just have computed this once instead of on each heartbeat?

Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -88,11 +88,80 @@ class SessionController: SessionControllable {
...
+        guard now >= maxEnd else { return }
+
+        queue.async { [weak self] in
```
fnewberg, line 107: is there a chance for checkUserSessionMaxDurationExpiry  function to be called again before `rollPartForUserSessionExpiry()` is processed off the queue? if so, do we need to  protect against this? e.g. consider this scenario

  1. heartbeat fires  — `checkUserSessionMaxDurationExpiry` sees expired session, enqueues roll onto queue
  2. before the roll executes, heartbeat fires again, sees the same still-expired session (user session hasn't been replaced yet), enqueues a second roll
  3. first roll executes — ends old session, starts new session
  4. second roll executes — sees info.session != nil (the new session), rolls it too which would be wrong

Sources/EmbraceIO/EmbraceIO.swift
```diff
@@ -108,17 +108,13 @@ public class EmbraceIO {
...
-    /// Forces the Embrace SDK to stop the current session, if any.
-    /// - Note: This method won't do anything if the SDK is stopped.
-    public func endCurrentSession() {
-        Embrace.client?.endCurrentSession()
+    /// Ends the current user session immediately. The current part is closed and a new part
+    /// (with the same foreground/background state) is started under a fresh user session.
+    /// - Note: This call is rate-limited to once per 5 seconds. Calls within 5 seconds of the
+    ///   previous one are ignored silently.
+    /// - Note: This method has no effect if the SDK is stopped.
+    public func endUserSession() {
```
fnewberg, line 116: consider adding test coverage for this.  i don't think we have anything for it currently

Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -88,11 +88,80 @@ class SessionController: SessionControllable {
...
+    }
+
+    /// If a foreground part is about to start after a background part whose user session
+    /// crossed its `maxDuration` or `inactivityTimeout` cutoff, slice the bg part at the
+    /// cutoff and synthesize a bg part `[cutoff, now]`. Returns `true` when the split was
+    /// applied — the caller must skip its own "end previous part" step because the prev part
+    /// is already closed.
+    @discardableResult
+    private func applyBackgroundSplitIfNeeded(
+        prev: EmbraceSession?,
```
fnewberg, line 126: nitpick: i'd consider naming this `prevSessionPart` to be more explicit

Sources/EmbraceCore/Session/UserSessionController.swift
fnewberg, line 116:
```suggestion
        if let reason = Self.expiryReason(snapshot: snapshot, at: dateProvider()) {
            sessionController?.backfillTerminationReasonOnLatestPart(reason)
```

Sources/EmbraceCore/Session/SessionController.swift
```diff
@@ -88,11 +88,80 @@ class SessionController: SessionControllable {
...
+        startSession(state: .background, startTime: cutoff)
+        endSession(at: now)
+        return true
+    }
+
+    /// Closes the current part at `now`, ends the user session with the given reason, then
+    /// starts a new part with the same `SessionState`. Used by the heartbeat-driven max-duration
+    /// detector and the manual `endUserSession()` API. The order — end-part → end-user-session
+    /// → start-part — guarantees that `attachPart` for the new part sees no active user session.
+    func rollPartForUserSessionExpiry(reason: TerminationReason, at now: Date) {
```
fnewberg, line 156: The three-step sequence implemented below (`endSession` → `endActiveUserSession` → `startSession`), has each step individually locked but the overall sequence is not atomic . Heartbeat fires on `self.queue` from L112, manual `endUserSession()` fires on `processingQueue` from `Embrace.swift:413` — separately serialized queues with no cross-queue coordination.

Some thoughts on how to address this:
 
- add a `rollInProgress` to `UserSessionController`'s `State` that gates rolling
- indicate what user session should be rolled and enforce it, e.g. something like this

```
  // Capture intent at the call site, not at the run site.

  private func checkUserSessionMaxDurationExpiry(now: Date) {
      guard let userSession = userSessionController?.currentUserSession else { return }
      let maxEnd = userSession.startTime.addingTimeInterval(userSession.maxDuration)
      guard now >= maxEnd else { return }

      let target = userSession.id                    // ← captured here, while we still see US0
      queue.async { [weak self] in
          self?.rollPartForUserSessionExpiry(reason: .maxDurationReached, at: now, expecting: target)
      }
  }

  package func endUserSession() {
      guard isSDKEnabled else { return }
      processingQueue.async { [weak self] in
          guard let self = self else { return }
          let now = Date()
          guard self.userSessionController.canManuallyEnd(now: now) else { return }
          guard let target = self.userSessionController.currentUserSession?.id else { return }
          self.sessionController.rollPartForUserSessionExpiry(reason: .manual, at: now, expecting: target)
      }
  }

  func rollPartForUserSessionExpiry(reason: TerminationReason, at now: Date, expecting target: EmbraceIdentifier) {
      let info = _session.safeValue
      guard info.session != nil else { return }
      guard userSessionController?.currentUserSession?.id == target else {
          return                                     // someone else already ended my target
      }
      let currentState = info.session?.state ?? .unknown
      endSession(at: now)
      userSessionController?.endActiveUserSession(reason: reason, at: now)
      startSession(state: currentState, startTime: now)
  }
```

Tests/EmbraceCoreTests/Session/SessionControllerTests.swift
```diff
@@ -179,6 +179,213 @@ final class SessionControllerTests: XCTestCase {
...
+        // Background part that begins before the cutoff and would normally span past it.
+        splitController.startSession(state: .background, startTime: foregroundEnd.addingTimeInterval(0.1))
+
+        // Wait long enough for the cutoff to have passed.
+        Thread.sleep(forTimeInterval: 1.2)
+        let bgEnd = Date()
+        XCTAssertGreaterThan(bgEnd, cutoffExpected)
+
+        // Now transition to foreground: bg-split fires.
+        splitController.startSession(state: .foreground, startTime: bgEnd)
```
fnewberg: since `startSession` accepts a `startTime` we should be able to avoid the sleep.

```suggestion
        let bgEnd = foregroundEnd.addingTimeInterval(1.5) // past the 1s cutoff
        // Now transition to foreground: bg-split fires.
        splitController.startSession(state: .foreground, startTime: bgEnd)
```

Sources/EmbraceCore/Session/UserSessionController.swift
```diff
@@ -244,18 +250,21 @@ final class UserSessionController {
     }
 
     /// Returns the termination reason if the snapshot has expired at `now`, else `nil`.
+    /// The cutoff boundaries are inclusive — a part starting exactly at the cutoff is treated
+    /// as expired so the bg-split path (which starts a synthetic part *at* the cutoff) cleanly
+    /// rolls into a new user session.
     /// Inactivity cutoff is only applicable when there is a `lastForegroundPartEnd`.
     private static func expiryReason(snapshot: ImmutableUserSession, at now: Date) -> TerminationReason? {
```
fnewberg, line 257: Clock-anomaly handling catches `now` < `startTime` but not `now` < `lastForegroundPartEnd`. A backward clock jump, or `bootstrap()` on a device whose clock is in the past relative to
the prior process's recorded `lastForegroundPartEnd`, leaves the inactivity check evaluating `now >= lastFgEnd + timeout` against a `lastFgEnd` that's in the future, which is always
false. The user session escapes inactivity expiry entirely; it survives until `maxDuration` eventually fires.

We can handle this by adding this to the function

```
if let lastFgEnd = snapshot.lastForegroundPartEnd, now < lastFgEnd {
    return .clockAnomaly
}
```

Tests/EmbraceCoreTests/TestSupport/TestDoubles/MockSessionController.swift
```diff
@@ -21,6 +21,14 @@ class MockSessionController: SessionControllable {
     var didCallEndSession: Bool = false
     var didCallUpdateSession: Bool = false
 
+    /// Ordered log of every `startSession`/`endSession` call (and their timestamp) for tests
+    /// that verify the bg-split sequence in `iOSSessionLifecycle.appDidBecomeActive`.
+    enum Event: Equatable {
+        case startSession(state: SessionState, startTime: Date)
+        case endSession(at: Date)
+    }
+    var callLog: [Event] = []
```
fnewberg: we append things to this, but AFAICT we never validate what `callLog` is

Sources/EmbraceCore/Session/SessionController.swift
fnewberg: more thoughts on this below as i worked my way through the PR

Sources/EmbraceCore/Session/SessionController.swift
NachoEmbrace: Like I said in the other PR, if we want to go for the rename, i'd do it at the end in a full sweep.

Sources/EmbraceCore/Session/SessionController.swift
NachoEmbrace: Maybe possible I think? but would be kinda crazy since the heartbeat interval is 5 seconds.
Would mean the async dispatch on the SessionController queue took more than 5 seconds.

Sources/EmbraceIO/EmbraceIO.swift
NachoEmbrace: UserSessionControllerTests, canManuallyEnd related tests cover the use case.

NOTE: this obviously is not a test on the actual API. We don't really have tests over the Embrace class since it causes a lot of issues. If we want to address this we can discuss it but it would be on another PR, and it's a completely separated issue.

Sources/EmbraceCore/Session/SessionController.swift
NachoEmbrace: Addressed by making endUserSession queue the work into the userSessionController queue. This fixes the potential race entirely.
Also added a test for this.

Sources/EmbraceCore/Session/UserSessionController.swift
NachoEmbrace: What's the reason for this?

Sources/EmbraceCore/Session/SessionController.swift
NachoEmbrace: Addressed and added a test.

Tests/EmbraceCoreTests/TestSupport/TestDoubles/MockSessionController.swift
NachoEmbrace: Removed, not used

Sources/EmbraceCore/Session/UserSessionController.swift
fnewberg: so we have cold-start expiry add a termination reason that mid-session expiry would do

fnewberg (approved):
added a response to the one open comment, but looks good overall

PR #617, opened 2026-05-15
[EMBR-12008] Feature: Propagating new fields to signals and payloads

NachoEmbrace:
This PR adds the new  user session related keys to our payloads and generated otel signals.

Note that this PR is pointing to https://github.com/embrace-io/embrace-apple-sdk/pull/613

### DIFF
--- a/Sources/EmbraceCommonInternal/OTelBridge/EmbraceMetadataProvider.swift
+++ b/Sources/EmbraceCommonInternal/OTelBridge/EmbraceMetadataProvider.swift
@@ -16,8 +16,7 @@ package protocol EmbraceMetadataProvider: AnyObject {
     var currentSessionId: EmbraceIdentifier? { get }
 
     /// Returns the identifier for the current user session, if any.
-    /// Stamped on spans/logs as both `session.id` (in v7, `session.id` represents the user
-    /// session, not the part) and `emb.user_session_id`.
+    /// Stamped on spans/logs as both `session.id`
     var currentUserSessionId: EmbraceIdentifier? { get }
 
     /// Returns the identifier for the current process
--- a/Sources/EmbraceCommonInternal/OTelBridge/EmbraceMetadataProvider.swift
+++ b/Sources/EmbraceCommonInternal/OTelBridge/EmbraceMetadataProvider.swift
@@ -11,8 +11,14 @@ import Foundation
 /// Protocol used to provide Embrace specific information to 3rd party OTel implementations
 package protocol EmbraceMetadataProvider: AnyObject {
 
-    /// Returns the identifier for the current Embrace session, if any
+    /// Returns the identifier for the current Embrace session part, if any.
+    /// Stamped on spans/logs as `emb.session_part_id`.
     var currentSessionId: EmbraceIdentifier? { get }
+
+    /// Returns the identifier for the current user session, if any.
+    /// Stamped on spans/logs as both `session.id` (in v7, `session.id` represents the user
+    /// session, not the part) and `emb.user_session_id`.
+    var currentUserSessionId: EmbraceIdentifier? { get }
 
     /// Returns the identifier for the current process
     var currentProcessId: EmbraceIdentifier { get }
--- a/Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
+++ b/Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
@@ -51,8 +51,7 @@ extension DefaultOTelSignalsHandler: InternalOTelSignalsHandler {
 
             // Cross-cutting attribute stamping. Every non-session-part span carries the three
             // identity keys so backend correlation works even on spans that started before
-            // an active session existed (in which case the values are empty strings, per the
-            // backend's "presence of `emb.session_part_id` means new SDK" detection).
+            // an active session existed .
             internalAttributes.setSessionIdentity(
                 userSessionId: sessionController?.currentSession?.userSessionId?.stringValue ?? "",
                 partId: sessionId?.stringValue ?? ""
--- a/Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
+++ b/Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
@@ -37,15 +37,26 @@ extension DefaultOTelSignalsHandler: InternalOTelSignalsHandler {
         var internalAttributes = [String: EmbraceAttributeValue]()
         internalAttributes.setEmbraceType(type)
 
-        // add session id if needed
+        // Resolve the part id used for the spans-to-parts storage association. The session
+        // part span itself carries its id via `keyPartId` in `attributes` (set by
+        // `SessionSpanUtils.span`); every other span inherits the active part from the
+        // session controller.
         var sessionId: EmbraceIdentifier?
         if type == .session {
-            if let value = internalAttributes[SpanSemantics.Session.keyId] as? String {
+            if let value = sanitizedAttributes[SpanSemantics.Session.keyPartId] as? String {
                 sessionId = EmbraceIdentifier(stringValue: value)
             }
         } else {
             sessionId = sessionController?.currentSession?.id
-            internalAttributes.setEmbraceSessionId(sessionId)
+
+            // Cross-cutting attribute stamping. Every non-session-part span carries the three
+            // identity keys so backend correlation works even on spans that started before
+            // an active session existed (in which case the values are empty strings, per the
+            // backend's "presence of `emb.session_part_id` means new SDK" detection).
+            internalAttributes.setSessionIdentity(
+                userSessionId: sessionController?.currentSession?.userSessionId?.stringValue ?? "",
+                partId: sessionId?.stringValue ?? ""
+            )
         }
 
         let finalAttributes = internalAttributes.merging(sanitizedAttributes) { (current, _) in current }
@@ -406,11 +417,14 @@ extension DefaultOTelSignalsHandler: EmbraceOTelDelegate {
 
     /// Attribute keys stamped by `EmbraceSpanProcessor.injectAttributes()` before the span
     /// reaches the delegate. These must always be preserved in storage regardless of the
-    /// attribute count limit.
+    /// attribute count limit. The three identity keys (`session.id`, `emb.user_session_id`,
+    /// `emb.session_part_id`) are protected so the cross-cutting stamping survives sanitization.
     static let bridgeProtectedKeys: Set<String> = [
         SpanSemantics.keyEmbraceType,
         SpanSemantics.Session.keyState,
-        SpanSemantics.keySessionId
+        SpanSemantics.keySessionId,
+        SpanSemantics.Session.keyUserSessionId,
+        SpanSemantics.Session.keyPartId
     ]
 
     public func onStartSpan(_ span: EmbraceSpan) {
--- a/Sources/EmbraceCore/Internal/OTel/Dictionary+EmbraceAttributes.swift
+++ b/Sources/EmbraceCore/Internal/OTel/Dictionary+EmbraceAttributes.swift
@@ -13,7 +13,13 @@ extension Dictionary where Key == String, Value == EmbraceAttributeValue {
         self[SpanSemantics.keyEmbraceType] = type.rawValue
     }
 
-    mutating func setEmbraceSessionId(_ sessionId: EmbraceIdentifier?) {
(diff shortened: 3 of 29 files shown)


Sources/EmbraceCommonInternal/OTelBridge/EmbraceMetadataProvider.swift
```diff
@@ -11,9 +11,15 @@ import Foundation
 /// Protocol used to provide Embrace specific information to 3rd party OTel implementations
 package protocol EmbraceMetadataProvider: AnyObject {
 
-    /// Returns the identifier for the current Embrace session, if any
+    /// Returns the identifier for the current Embrace session part, if any.
+    /// Stamped on spans/logs as `emb.session_part_id`.
     var currentSessionId: EmbraceIdentifier? { get }
```
fnewberg, line 16: would renaming this to currentSessionPartID have a large blast radius?

Sources/EmbraceCommonInternal/OTelBridge/EmbraceMetadataProvider.swift
```diff
@@ -11,9 +11,15 @@ import Foundation
...
 
+    /// Returns the identifier for the current user session, if any.
+    /// Stamped on spans/logs as both `session.id` (in v7, `session.id` represents the user
+    /// session, not the part) and `emb.user_session_id`.
```
fnewberg: no need to refer to v7 in the comments

```suggestion
    /// Stamped on spans/logs as both `session.id`
```

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
```diff
@@ -37,15 +37,26 @@ extension DefaultOTelSignalsHandler: InternalOTelSignalsHandler {
...
                 sessionId = EmbraceIdentifier(stringValue: value)
             }
         } else {
             sessionId = sessionController?.currentSession?.id
-            internalAttributes.setEmbraceSessionId(sessionId)
+
+            // Cross-cutting attribute stamping. Every non-session-part span carries the three
+            // identity keys so backend correlation works even on spans that started before
+            // an active session existed (in which case the values are empty strings, per the
+            // backend's "presence of `emb.session_part_id` means new SDK" detection).
```
```suggestion
            // an active session existed .
```

Sources/EmbraceCore/Session/SessionSpanUtils.swift
```diff
@@ -72,47 +75,71 @@ extension SpanPayload {
...
-                key: SpanSemantics.Session.keyCleanExit,
-                value: String(session.cleanExit)
+                key: SpanSemantics.Session.keyUserSessionStartTs,
+                value: String((session.userSessionStartTime ?? session.startTime).nanosecondsSince1970Truncated)
             ),
             Attribute(
-                key: SpanSemantics.Session.keyHeartbeat,
-                value: String(session.lastHeartbeatTime.nanosecondsSince1970Truncated)
+                key: SpanSemantics.Session.keyUserSessionMaxDurationSeconds,
+                value: String(session.userSessionMaxDuration ?? 0)
```
fnewberg: consistently emit ints

```suggestion
                value: String(Int(session.userSessionMaxDuration ?? 0))
```

Sources/EmbraceCore/Session/SessionSpanUtils.swift
```diff
@@ -72,47 +75,71 @@ extension SpanPayload {
...
             ),
             Attribute(
-                key: SpanSemantics.Session.keySessionNumber,
-                value: String(sessionNumber)
+                key: SpanSemantics.Session.keyUserSessionInactivityTimeoutSeconds,
+                value: String(session.userSessionInactivityTimeout ?? 0)
```
fnewberg: consistently emit ints


```suggestion
                value: String(Int(session.userSessionInactivityTimeout ?? 0))
```

fnewberg:
The changes in this PR affect the MetricKit crash → user-session linking.

We add the session ID to MetricKit payloads [here](https://github.com/embrace-io/embrace-apple-sdk/blob/715cfa0b7025831b62f9666ffe838bb6c4272cc8/Sources/EmbraceCore/Capture/MetricKit/MetricKitCrashCaptureService.swift#L52)
 
That call site builds `EmbraceLogAttributesBuilder` with `session: nil` and then invokes the 1-arg `addSessionIdentifier(_:)` overload with a historical part id. The overload resolves `userSessionId` via `currentSession?.userSessionId`, which is nil in that construction — so the log ships with `session.id` and `emb.user_session_id` as empty strings, while `emb.session_part_id` correctly carries the historical part UUID.
 
Previously the same call set `session.id` to the historical part UUID, so we need to decide if we are okay with this change , and will adjust accordingly on the backend or if we want to make SDK changes.

Sources/EmbraceCore/Session/SessionSpanUtils.swift
```diff
@@ -72,47 +75,71 @@ extension SpanPayload {
...
+                value: String(session.userSessionPartIndex)
             ),
+
+            // User-session config snapshots — read from the part record, which captured them
+            // when the user session was created. Pre-v7 rows return nil/0; emit defaults.
             Attribute(
-                key: SpanSemantics.Session.keyCleanExit,
-                value: String(session.cleanExit)
+                key: SpanSemantics.Session.keyUserSessionStartTs,
+                value: String((session.userSessionStartTime ?? session.startTime).nanosecondsSince1970Truncated)
```
fnewberg, line 114: do we want this fallback? verify this is the desired behavior that the backend team wants to see (cc: @juansc )

Sources/EmbraceCore/Internal/OTel/Dictionary+EmbraceAttributes.swift
```diff
@@ -13,7 +13,13 @@ extension Dictionary where Key == String, Value == EmbraceAttributeValue {
         self[SpanSemantics.keyEmbraceType] = type.rawValue
     }
 
-    mutating func setEmbraceSessionId(_ sessionId: EmbraceIdentifier?) {
-        self[SpanSemantics.keySessionId] = sessionId?.stringValue
+    /// Stamps the three identity keys that must appear on every span and every log:
+    /// `session.id` (= user-session UUID), `emb.user_session_id`, and `emb.session_part_id`.
+    /// All three are emitted unconditionally — empty strings when unknown — because the
+    /// backend uses the presence of `emb.session_part_id` to detect new SDKs.
+    mutating func setSessionIdentity(userSessionId: String, partId: String) {
```
fnewberg, line 20: we have this helper here, but there are three other places that write the same three keys using the same logic instead of calling it:

  - [`EmbraceSpanProcessor.swift:131-135`](https://github.com/embrace-io/embrace-apple-sdk/blob/main/Sources/EmbraceOTelBridge/Processors/EmbraceSpanProcessor.swift#L131-L135)
  - [`EmbraceLogProcessor.swift:49-51`](https://github.com/embrace-io/embrace-apple-sdk/blob/main/Sources/EmbraceOTelBridge/Processors/EmbraceLogProcessor.swift#L49-L51)
  - [`SessionSpanUtils.swift:83-88`](https://github.com/embrace-io/embrace-apple-sdk/blob/main/Sources/EmbraceCore/Session/SessionSpanUtils.swift#L83-L88)

Sources/EmbraceSemantics/Log/LogSemantics.swift
fnewberg, line 5:
```suggestion
public enum LogSemantics {
```

Sources/EmbraceSemantics/Span/SpanSemantics+Session.swift
```diff
@@ -9,12 +9,45 @@ extension EmbraceType {
 extension SpanSemantics {
     public struct Session {
```
fnewberg, line 10:
```suggestion
    public enum Session {
```

Sources/EmbraceCore/Session/SessionSpanUtils.swift
```diff
@@ -72,47 +75,71 @@ extension SpanPayload {
...
                 Attribute(
-                    key: SpanSemantics.Session.keyCrashId,
-                    value: crashId
-                ))
+                    key: SpanSemantics.Session.keyUserSessionTerminationReason,
+                    value: terminationReason.rawValue
+                )
+            )
+            attributeArray.append(
+                Attribute(key: SpanSemantics.Session.keyIsFinalSessionPart, value: "1")
```
fnewberg, line 141: @juansc was there a reason we made this `"1" instead of `"true"` which would be more consistent with other attributes like `coldStart`

Sources/EmbraceCore/Session/SessionSpanUtils.swift
NachoEmbrace: The spec didn't specify if the value should be truncated or not.
Are we sure we don't care about the decimals? (`session.userSessionMaxDuration` is a `TimeInternal` aka `Double`)

Sources/EmbraceCore/Session/SessionSpanUtils.swift
NachoEmbrace: I think this is wrong, actually.

Sources/EmbraceCore/Internal/OTel/Dictionary+EmbraceAttributes.swift
NachoEmbrace: The first 2 are in a different module, but either way, the code is not dealing with a dictionary, it is dealing with OTel types. The extension on Dictionary you are pointing to is not applicable there, even if we moved it to a place where it could be imported from the otel bridge module.

The last one in SessionSpanUtils is also not dealing with a Dictionary, so the helper can't be used.

Sources/EmbraceSemantics/Log/LogSemantics.swift
NachoEmbrace: Why?

Sources/EmbraceCommonInternal/OTelBridge/EmbraceMetadataProvider.swift
NachoEmbrace: It would be very inconsistent with the rest of the code.

I'd be okay doing this exercise at the end once everything is merged though, but it would be this change alongside many others, so the code is consistent at the end.
The blast radius will be very big, specially on tests, but if we really want to do it, we have to do it now (aka before releasing 7.0)

Sources/EmbraceSemantics/Span/SpanSemantics+Session.swift
NachoEmbrace: Why?

Sources/EmbraceCore/Session/SessionSpanUtils.swift
fnewberg: yeah, pretty sure that these values will only be ints. i don't think we'll ever want to specify 3600.5 seconds. cc: @juansc

Sources/EmbraceCommonInternal/OTelBridge/EmbraceMetadataProvider.swift
fnewberg: not a big deal. if it's a giant blast radius i'm fine keeping it as is

Sources/EmbraceSemantics/Span/SpanSemantics+Session.swift
fnewberg: i think it's just a minor thing as described [here](https://github.com/kodecocodes/swift-style-guide#constants) (search for `enum`). not a big deal

Sources/EmbraceCore/Session/SessionSpanUtils.swift
juansc: I'd rather have an unset value rather than a fallback. That itself tells us information (the SDK didn't know when this was set) rather than getting wrong information

Sources/EmbraceCore/Session/SessionSpanUtils.swift
juansc: These will always by Ints.

Sources/EmbraceCore/Session/SessionSpanUtils.swift
juansc: I didn't verify against the other values. From the code I can't tell if there's a cross-SDK standard, as it just parses all standard truthy values like 1, t, T, etc. I didn't get any pushback from Android, and the code will handle either case. You can stick to `true` if it keeps the iOS code cleaner

juansc (commented):
Thanks for the questions! Keep them coming

Sources/EmbraceCore/Session/SessionSpanUtils.swift
fnewberg: makes sense, so we should just omit the attribute?

PR #618, opened 2026-05-15
[EMBR-12009] Feature: New user session keys on crash reporters

NachoEmbrace:
New keys for crash reporters

Note that this PR is pointing to https://github.com/embrace-io/embrace-apple-sdk/pull/617

### DIFF
--- a/Sources/EmbraceCore/Capture/CaptureServices.swift
+++ b/Sources/EmbraceCore/Capture/CaptureServices.swift
@@ -104,6 +104,15 @@ final class CaptureServices {
             object: nil
         )
 
+        // Clear the crash-reporter's user-session id when the active user session ends.
+        // The next part start will repopulate it via `onSessionStart`.
+        NotificationCenter.default.addObserver(
+            self,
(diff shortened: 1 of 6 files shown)


Sources/EmbraceCore/Capture/CaptureServices.swift
```diff
@@ -177,6 +187,10 @@ final class CaptureServices {
         for service in services { service.onSessionWillEnd() }
     }
 
+    @objc func onUserSessionDidEnd(notification: Notification) {
```
fnewberg, line 190: is there any chance that processing of `onUserSessionDidEnd` and `onUserSessionDidStart` could race given the close proximity in which they are fired when a user session is force-ended? could we end up processing `onUserSessionDidEnd` after `onUserSessionDidStart`?

NachoEmbrace: Both notifications are posted into the main thread, in order. There shouldn't be any issues since we're not doing anything fancy with the queues when listening to them.

fnewberg (approved):

PR #626, opened 2026-05-20
[EMBR-12281] Chore: Examples: Update BrandGame and Benchmarks for 7.0 EmbraceIO public API

fnewberg:
On the `7.0` branch the SDK's public API surface renamed `Embrace` → `EmbraceIO` and reshaped `Options`. PR #590 already migrated `Examples/EmbraceIOTestApp/`; this PR does the same for the two remaining example apps — `Examples/BrandGame` and `Examples/Benchmarks` — so the `Build Example Apps / Build BrandGame App (26.0)` and `On Device Benchmarks / build` jobs don't fail. on 7.0-descendant PR.

## Checklist

### DIFF
--- a/Examples/Benchmarks/Benchmarks.xcodeproj/project.pbxproj
+++ b/Examples/Benchmarks/Benchmarks.xcodeproj/project.pbxproj
@@ -202,7 +202,7 @@
 			mainGroup = 2A7143B02E79E41500CCE829;
 			minimizedProjectReferenceProxies = 1;
 			packageReferences = (
-				2A7143E32E79E45700CCE829 /* XCLocalSwiftPackageReference "../../../embrace-apple-sdk" */,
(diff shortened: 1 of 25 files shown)


NachoEmbrace:
I have no idea where this "waitForAllWork" API came from, but I don't think it should come back.

That being said, I don't think the Benchmark app will work properly with these changes, so I'm assuming this API was added just for the Benchmark app. However this API looks odd and out of place in our SDK; so I think we should try to find an alternative.

This PR should disable the benchmark jobs if we want to merge it, since the Benchmark app won't be reliable.

EDIT: So I just remembered this API was hidden with `@_spi(Private)` which I guess is an OK compromise.
Modify this PR so EmbraceIO exposes a public method `waitForAllWork` (also flagged as `@_spi(Private)`) that just calls the Embrace one; and have the app use the new API.

NachoEmbrace (approved):

PR #632, opened 2026-05-21
[EMBR-12070] Fix: Fix: Guard EmbraceUploadOperation against cancel/completion race

fnewberg:
`EmbraceUploadOperation.task` (`URLSessionDataTask?`) was written from a GCD worker thread inside `sendRequest(_:retryCount:)` and read from the main thread inside `cancel()` with no synchronization. The class is `@unchecked Sendable`, so the compiler didn't catch it. PR #562 added `test_cancelKeepsRecordInCache` which made this failure observable. TSan reports `Crash: xctest at EmbraceUploadOperation.task.getter`and on a simulator without TSan, the race deadlocks.

This PR adds a `State` struct guarded by `EmbraceMutex<State>` that bundles `attemptCount`, `task`, and a new `hasFinished` one-shot gate. 

Two correctness properties land together:

- **Race-free reads/writes** of `task` and `attemptCount` across `cancel()` (main) and `sendRequest()` (worker).
- **Exactly-once completion**: whichever of `cancel()` or `handleTaskCompletion` flips `hasFinished` inside the lock fires `completion` + `finish()`; the other returns. Replaces the racy `guard !isCancelled` check at the top of `handleTaskCompletion`.

`sendRequest` uses a tentative-then-commit pattern: the post-increment `attemptCount` is computed inside one lock acquisition, the `URLSessionDataTask` is built outside the lock (so we don't hold the mutex across `URLSession.dataTask(...)`), then committed under a second acquisition. If `cancel()` flips `hasFinished` between the two, the just-built task is `cancel()`-ed without `resume()` (avoiding a leaked network call) and the function returns.

Lock discipline: foreign code (`URLSessionDataTask.cancel()`, `completion`, `finish()`) is never invoked from inside `state.withLock` — the underlying `EmbraceMutex` is non-recursive.

### How to repro the issue

Run the existing single test under TSan with hang detection. This must be run on the branch for #630 or on `main` after #630 has been merged  :

```sh
xcodebuild test -workspace .swiftpm/xcode/package.xcworkspace -scheme EmbraceIO-Package \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro' \
  -enableThreadSanitizer YES -test-iterations 50 \
  -test-timeouts-enabled YES -default-test-execution-time-allowance 60 \
  -only-testing:EmbraceUploadInternalTests/EmbraceUploadOrderedDeliveryTests/test_cancelKeepsRecordInCache
```

Iteration 1 hangs past the 60s allowance → `TEST FAILED`. On this branch, the same test completes in ~0.028s per iteration; 50 iterations clean.

### Stacking

Stacked on  PR #630 which adds `EmbraceHTTPMock.clearRequests()` to the three upload-tests classes so
`-test-iterations` runs no longer accumulate static-dict state. Without it, the new `test_completionFiresExactlyOnceWhenCancelledAfterSuccess` and `test_genericRequestHeaders` collide in the full suite. Merge the mock-fix PR first.

## Checklist

### DIFF
--- a/Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
+++ b/Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
@@ -76,16 +76,16 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
     override func cancel() {
         super.cancel()
 
-        let (taskToCancel, attemptCount, shouldFire) = state.withLock {
-            guard !$0.hasFinished else {
-                return (nil, $0.attemptCount, false)
+        let (taskToCancel, attemptCount, shouldFire) = state.withLock { s -> (URLSessionDataTask?, Int, Bool) in
+            guard !s.hasFinished else {
+                return (nil, s.attemptCount, false)
             }
-            
+
             let taskToCancel = s.task
-            $0.task = nil
-            $0.hasFinished = true
-            
-            return (taskToCancel, $0.attemptCount, true)
+            s.task = nil
+            s.hasFinished = true
+
+            return (taskToCancel, s.attemptCount, true)
         }
 
         taskToCancel?.cancel()
@@ -127,20 +127,18 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
                 )
             })
 
-        // If operation finished meanwhile, cancel the new task immediately.
+        // If operation finished meanwhile, drop the unstarted task; it was never resumed.
         let started = state.withLock {
             guard !$0.hasFinished else { return false }
-            
+
             $0.attemptCount = nextAttemptCount
             $0.task = newTask
-            
+
             return true
         }
 
         if started {
             newTask.resume()
-        } else {
-            newTask.cancel()
         }
     }
 
@@ -184,10 +182,10 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
         }
 
         let (shouldFire, attemptCount) = state.withLock {
-            guard !$0.hasFinished else { 
+            guard !$0.hasFinished else {
                 return (false, $0.attemptCount)
             }
-            
+
             $0.hasFinished = true
             return (true, $0.attemptCount)
         }
--- a/Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
+++ b/Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
@@ -76,14 +76,16 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
     override func cancel() {
         super.cancel()
 
-        let (taskToCancel, attemptCount, shouldFire) = state.withLock { s -> (URLSessionDataTask?, Int, Bool) in
-            guard !s.hasFinished else {
-                return (nil, s.attemptCount, false)
+        let (taskToCancel, attemptCount, shouldFire) = state.withLock {
+            guard !$0.hasFinished else {
+                return (nil, $0.attemptCount, false)
             }
+            
             let taskToCancel = s.task
-            s.task = nil
-            s.hasFinished = true
-            return (taskToCancel, s.attemptCount, true)
+            $0.task = nil
+            $0.hasFinished = true
+            
+            return (taskToCancel, $0.attemptCount, true)
         }
 
         taskToCancel?.cancel()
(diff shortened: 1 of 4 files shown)


Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
```diff
@@ -55,18 +68,29 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
...
-        finish()
+        let (taskToCancel, attemptCount, shouldFire) = state.withLock { s -> (URLSessionDataTask?, Int, Bool) in
+            guard !s.hasFinished else {
+                return (nil, s.attemptCount, false)
+            }
+            let taskToCancel = s.task
+            s.task = nil
+            s.hasFinished = true
+            return (taskToCancel, s.attemptCount, true)
+        }
```
NachoEmbrace:
```suggestion
        let (taskToCancel, attemptCount, shouldFire) = state.withLock {
            guard !$0.hasFinished else {
                return (nil, $0.attemptCount, false)
            }
            
            let taskToCancel = s.task
            $0.task = nil
            $0.hasFinished = true
            
            return (taskToCancel, $0.attemptCount, true)
        }
```

Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
```diff
@@ -102,7 +125,19 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
...
             })
 
-        task?.resume()
+        // If operation finished meanwhile, cancel the new task immediately.
+        let started = state.withLock { s -> Bool in
+            guard !s.hasFinished else { return false }
+            s.attemptCount = nextAttemptCount
+            s.task = newTask
+            return true
+        }
```
NachoEmbrace:
```suggestion
        let started = state.withLock {
            guard !$0.hasFinished else { return false }
            
            $0.attemptCount = nextAttemptCount
            $0.task = newTask
            
            return true
        }
```

Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
```diff
@@ -102,7 +125,19 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
...
+
+        if started {
+            newTask.resume()
+        } else {
+            newTask.cancel()
```
NachoEmbrace: This makes me nervous. The overall problem kinda makes sense, but I don't like the solution if it explicitly adds code that cancels the operation (we never did this before).

I'm also not entirely sure if we actually have to do this since the only way for `started` to be false here, is if `state.hasFinished` is false, which only happens if:
- The operation was cancelled already (makes no sense to cancel again)
- The request actually finished (makes no sense that this could happen at all, but even if it did, the operation is about to finish, no reason to cancel it).

Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
```diff
@@ -112,17 +147,15 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
         request: URLRequest,
         retryCount: Int
     ) {
-        // If the operation was cancelled, cancel() already fired the completion and called finish().
-        // Just return to avoid double-completion and double-finish.
-        guard !isCancelled else {
-            return
-        }
+        // Fast path: operation already completed or cancelled.
+        if state.withLock({ $0.hasFinished }) { return }
```
NachoEmbrace:
```suggestion
        guard !state.safeValue.hasFinished else { return }
```

Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
```diff
@@ -136,21 +169,26 @@ class EmbraceUploadOperation: AsyncOperation, @unchecked Sendable {
...
-            completion?(.failure, attemptCount)
+            result = .failure
         }
 
-        finish()
+        let (shouldFire, attemptCount) = state.withLock { s -> (Bool, Int) in
+            guard !s.hasFinished else { return (false, s.attemptCount) }
+            s.hasFinished = true
+            return (true, s.attemptCount)
+        }
```
NachoEmbrace:
```suggestion
        let (shouldFire, attemptCount) = state.withLock {
            guard !$0.hasFinished else { 
                return (false, $0.attemptCount)
            }
            
            $0.hasFinished = true
            return (true, $0.attemptCount)
        }
```

NachoEmbrace (changes requested):

Sources/EmbraceUploadInternal/Operations/EmbraceUploadOperation.swift
fnewberg: yeah, agreed. i'll get rid of the `else` clause

NachoEmbrace (approved):

PR #644, opened 2026-06-10
fix: Remove proxy only URLSessionDelegate proxy methods

gregiOS:
## Description. 

https://github.com/embrace-io/embrace-apple-sdk/pull/567/changes introduced changes that were targeted to fix the crash of intercepting URLSessionDelegate methods. While the fix from `nonatomic` to `atomic` property of `originalDelegate` seems valid, adding extra proxies methods introduce a regression into our app.
The regression is connected with intercepting `didReceiveChallenge` methods, that for some reason sometimes are not proxied into the consumer delegate but into initially injected `EmbraceDummyURLSessionDelegate`.
Due that some endpoints SSL pinning challenges are not passed into correct consumer delegate and that results with failing those.

This MR is just removing those methods as my understanding is those are kind of redundant, because if those methods are not implemented in the New/Legacy `EMBURLSessionNewDelegateProxy` those will be passed to the originalDelegate either by: 
```
- (BOOL)respondsToSelector:(SEL)sel
{
    // If we implement it directly (instance methods below), advertise YES.
    if ([super respondsToSelector:sel]) {
        return YES;
    }
    // Otherwise mirror the original delegate’s capabilities.
    return [self.originalDelegate respondsToSelector:sel];
}
```
or

```
- (id)forwardingTargetForSelector:(SEL)sel
{
    // We can't call `-respondsToSelector:` from here.
    if (sel == @selector(URLSession:task:didCompleteWithError:) ||
        sel == @selector(URLSession:task:didFinishCollectingMetrics:) ||
        sel == @selector(URLSession:dataTask:didReceiveData:) ||
        sel == @selector(URLSession:didBecomeInvalidWithError:)) {

        return nil;
    }

    id forwardingTarget = nil;

    forwardingTarget = [super forwardingTargetForSelector:sel];
    if (forwardingTarget) {
        return forwardingTarget == self ? nil : forwardingTarget;
    }

    // is the original doing any forwarding?
    forwardingTarget = [self.originalDelegate forwardingTargetForSelector:sel];
    if (forwardingTarget) {
        return forwardingTarget;
    }

    if ([self.originalDelegate respondsToSelector:sel]) {
        return self.originalDelegate;
    }

    return nil;
}
```

Adding a screenshot showing that somehow the selector of `didReceiveChallange` was attempted to be passed to 
`EmbraceDummyURLSessionDelegate `

Screenshot 2026-06-10 at 09 20 00

## Checklist

### DIFF
--- a/Sources/EmbraceObjCUtilsInternal/source/LegacyProxy/EMBURLSessionLegacyDelegateProxy.m
+++ b/Sources/EmbraceObjCUtilsInternal/source/LegacyProxy/EMBURLSessionLegacyDelegateProxy.m
@@ -14,8 +14,6 @@
 #define DID_BECOME_INVALID_WITH_ERROR @selector(URLSession:didBecomeInvalidWithError:)
 #define DID_RECEIVE_RESPONSE @selector(URLSession:dataTask:didReceiveResponse:completionHandler:)
 #define WILL_PERFORM_REDIRECTION @selector(URLSession:task:willPerformHTTPRedirection:newRequest:completionHandler:)
-#define DID_RECEIVE_CHALLENGE @selector(URLSession:didReceiveChallenge:completionHandler:)
-#define DID_RECEIVE_TASK_CHALLENGE @selector(URLSession:task:didReceiveChallenge:completionHandler:)
 
 @interface EMBURLSessionLegacyDelegateProxy ()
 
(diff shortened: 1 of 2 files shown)


ferdraghi:
Thanks for bringing this issue to our attention. While it is true that these changes introduced a regression breaking the challenge logic, the solution is not to remove all these methods. They were introduced for a reason and removing them risks creating crashes, problems to apps using 3rd party networking frameworks and more. 

We decided to go for a different approach and will be addressing the core issue you're having on [this PR instead](https://github.com/embrace-io/embrace-apple-sdk/pull/648).

Closing this for now.

PR #652, opened 2026-06-22
Add crash-resilient file-backed profiling persistence and recovery.

kstenerud:
## Summary

Adds optional file-backed persistence to `EmbraceProfiling` so sampling-profiler data survives abrupt process death (crash, OOM, or Jetsam) and can be recovered on the next launch. In-memory profiling (the default) is unchanged.

## Motivation

The sampling profiler previously held samples only in memory, so anything captured before an abrupt termination was lost. This adds a persistence path whose samples are flushed to disk by the kernel even on SIGKILL/Jetsam, plus a read-only recovery API to retrieve them next launch.

## How it works

- Samples are written into a double-mapped ("magic ring") `MAP_SHARED` region backed by a per-session file. The single-writer / multi-reader lock-free ring buffer is unchanged; only its backing store and control block moved into the mapping.
- On process death the kernel flushes the dirty `MAP_SHARED` pages to the file. (Validated on the macOS host and a physical device across all kill modes, including Jetsam.)
- The file carries a frozen 16-byte tail identity (magic + format version) and a versioned trailer; `format_version == 0` doubles as the transactional "disregard" marker and the clean-stop tombstone.
- Hardening: a `PROT_NONE` guard page bounds runaway writes; the metadata page is `mprotect`-locked write-once; on iOS the file uses data-protection Class B (CompleteUnlessOpen) so it keeps accepting writes while the device is locked.
- Recovery is strictly read-only. It validates the trailer/control block and walk records with torn-tail / zeroed-page / overrun stop conditions, discarding anything inconsistent. It never writes or deletes.

## API

- `start(configuration:directory:sessionId:)`: Pass a `directory` for file-backed mode (default `nil` = in-memory, unchanged). `sessionId` is an opaque 16-byte per-launch id.
- `recoverableSessions(in:) -> [RecoverableSession]`: Cheap enumeration (file metadata + a 16-byte identity peek, no samples loaded). Each handle exposes `sessionId`, `lastModified`, `byteSize`, and `status` (`.recoverable` / `.finalized`).
- `recover(_:) -> SessionRecoveryResult`: Recover a single session's samples.
- `delete(_:)`: Remove a session's file. The host app owns retention; the SDK never auto-deletes.
- `flush()`: `msync` ahead of a possible background Jetsam (call on background/terminate).
- `finalizeStorage()`: Write the clean-stop tombstone after the worker has stopped.
- `ProfilingSample` gains `threadState` (captured pre-suspend).

## Integration notes

- On a clean stop, call `finalizeStorage()` once the worker has exited (poll `isActive`); otherwise the session is listed as recoverable next launch: harmless, and `delete()`-able.
- Recovery calls are read-only and intended to run off the main thread at launch; process one handle at a time to bound memory.
- Retrieve any live samples before `finalizeStorage()`. Finalizing discards the on-disk copy.

## Compatibility

- In-memory mode is the default and behaves exactly as before; no change for existing callers.
- watchOS remains a no-op (file-backed APIs return empty / `.notSupported`).

## Testing

- 260 unit tests, 0 failures.
- Host + on-device kill-tests confirm samples survive SIGKILL/Jetsam and clean finalize.
- AddressSanitizer: clean. ThreadSanitizer: lock-free concurrency tests pass with only the inherent, by-design seqlock read/write pattern flagged (safe via sequence re-validation).
- iOS (arm64) and macOS builds verified.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,4 +86,6 @@ Examples/Benchmarks/Benchmarks.xcworkspace/xcuserdata
 # Claude
 .claude/
 claude/
+
+# Embrace upload cache (EmbraceUpload diskPath: "embrace") created at repo root when tests run
 /embrace/
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,4 @@ Examples/Benchmarks/Benchmarks.xcworkspace/xcuserdata
 # Claude
 .claude/
 claude/
+/embrace/
--- a/Sources/EmbraceProfiling/ProfilingEngine.swift
+++ b/Sources/EmbraceProfiling/ProfilingEngine.swift
@@ -2,6 +2,8 @@
 //  Copyright © 2026 Embrace Mobile, Inc. All rights reserved.
 //
 
+import Foundation  // URL / FileManager / String(format:) — used in the public API and file-backed paths
+
 #if !os(watchOS)
     import Darwin
     import EmbraceAtomicsShim
@@ -51,6 +53,9 @@ public final class ProfilingEngine: @unchecked Sendable {
         /// The sampler has entered an unrecoverable faulted state (bug or broken runtime).
         /// The associated string describes the fault reason.
         case faulted(reason: String)
+        /// File-backed storage setup failed (open/ftruncate/mmap/mprotect). The associated
+        /// string describes the cause (e.g. an errno). Only returned for file-backed starts.
+        case storageSetupFailed(reason: String)
         /// Profiling is not supported on this platform (e.g. watchOS).
         case notSupported
     }
@@ -76,6 +81,11 @@ public final class ProfilingEngine: @unchecked Sendable {
         internal var ringBuffer: UnsafeMutablePointer<emb_ring_buffer_t>?
         internal var readBuffer: UnsafeMutableRawPointer?
         internal var readBufferSize: Int = 0
+        /// File-backed store (persistent mode). nil for in-memory mode. When set, `ringBuffer`
+        /// points into the store's mapping and is owned by the store, not the engine.
+        internal var store: OpaquePointer?
+        /// Filename of the active file-backed session (so `recoverableSessions(in:)` skips it). nil if in-memory.
+        internal var activeSessionFileName: String?
         /// Atomic gate. CAS-based, no kernel ownership tracking.
         private let gateIsAcquired: UnsafeMutablePointer<emb_atomic_bool_t>
     #endif
@@ -183,8 +193,23 @@ public final class ProfilingEngine: @unchecked Sendable {
     /// - Parameter configuration: Profiling parameters. Defaults are suitable
     ///   for most use cases (10 Hz, 1MB buffer).
     /// - Returns: A ``StartResult`` indicating the outcome.
+    /// Start sampling.
+    ///
+    /// - Parameters:
+    ///   - configuration: profiling parameters.
+    ///   - directory: when non-nil, samples are persisted to a file-backed buffer in this directory
+    ///     (created if needed) so they survive a crash/Jetsam; recover them next launch via
+    ///     `recoverableSessions(in:)` + `recover(_:)`. When nil, the buffer is in-memory (the default).
+    ///   - sessionId: opaque 16-byte id for the file-backed session (becomes the filename). Must be
+    ///     unique per launch. Ignored for in-memory mode; a random id is used if nil.
+    ///
+    /// - Note: For a file-backed session, on a *clean* stop call ``finalizeStorage()`` after the worker
+    ///   has exited (poll ``isActive`` until `false`). Otherwise that session looks crash-like and is
+    ///   listed by ``recoverableSessions(in:)`` on the next launch.
     @discardableResult
-    public func start(configuration: ProfilingConfiguration = ProfilingConfiguration()) -> StartResult {
+    public func start(configuration: ProfilingConfiguration = ProfilingConfiguration(),
+                      directory: URL? = nil,
+                      sessionId: [UInt8]? = nil) -> StartResult {
         #if os(watchOS)
             return .notSupported
         #else
@@ -203,56 +228,42 @@ public final class ProfilingEngine: @unchecked Sendable {
 
             let wasActive = emb_sampler_is_active()
 
-            // If the sampler is already active, skip buffer manipulation
-            // (the worker thread may be writing to it) and let
-            // emb_sampler_start() report the accurate state.
-            let needsNewBuffer: Bool
+            // Decide the backing. If the sampler is already active, leave it untouched (the worker
+            // may be writing) and let emb_sampler_start() report the accurate state.
+            let createdNewBacking: Bool
             if wasActive {
-                needsNewBuffer = false
-            } else if let existing = ringBuffer {
-                if configuration.bufferCapacityBytes == activeConfiguration?.bufferCapacityBytes {
-                    // Retry briefly: the Dekker protocol in reset can fail if a
-                    // concurrent reader is briefly inside read_range. Under the
-                    // Swift gate this is unlikely, but defensive retries are cheap.
+                createdNewBacking = false
+            } else if let directory {
+                // FILE-BACKED: a new session is a new file — always fresh, no reuse. Drop any prior
+                // backing (in-memory buffer or a previous store) first.
+                tearDownBacking()
+                if let failure = setUpFileBackedBuffer(configuration: configuration,
+                                                       directory: directory, sessionId: sessionId) {
+                    return failure
+                }
+                createdNewBacking = true
+            } else {
+                // IN-MEMORY. If we were previously file-backed, drop the store first.
+                if store != nil { tearDownBacking() }
+                if let existing = ringBuffer,
+                    configuration.bufferCapacityBytes == activeConfiguration?.bufferCapacityBytes {
+                    // Same-capacity restart: reuse the buffer. Retry briefly — the Dekker protocol in
+                    // reset can fail if a concurrent reader is briefly inside read_range.
                     var resetOK = false
                     for _ in 0..<3 {
-                        if emb_ring_buffer_reset(existing) {
-                            resetOK = true
-                            break
-                        }
+                        if emb_ring_buffer_reset(existing) { resetOK = true; break }
                         usleep(200)
                     }
-                    if !resetOK {
-                        return .samplerBusy
-                    }
-                    needsNewBuffer = false
+                    if !resetOK { return .samplerBusy }
+                    createdNewBacking = false
                 } else {
-                    emb_ring_buffer_destroy(existing)
-                    if let ptr = readBuffer { free(ptr) }
-                    activeConfiguration = nil
-                    ringBuffer = nil
-                    readBuffer = nil
-                    readBufferSize = 0
-                    needsNewBuffer = true
+                    // First start, or capacity changed → fresh in-memory buffer.
+                    if ringBuffer != nil { tearDownBacking() }
+                    guard setUpInMemoryBuffer(configuration: configuration) else {
+                        return .bufferCreationFailed
+                    }
+                    createdNewBacking = true
                 }
-            } else {
-                needsNewBuffer = true
-            }
-
-            if needsNewBuffer {
-                guard let buffer = emb_ring_buffer_create(Int(configuration.bufferCapacityBytes), nil) else {
-                    return .bufferCreationFailed
-                }
-                ringBuffer = buffer
-
-                let capacity = emb_ring_buffer_capacity(buffer)
-                guard let newReadBuffer = malloc(capacity) else {
-                    emb_ring_buffer_destroy(buffer)
-                    ringBuffer = nil
-                    return .bufferCreationFailed
-                }
-                readBuffer = UnsafeMutableRawPointer(newReadBuffer)
-                readBufferSize = capacity
             }
 
             let config = emb_sampler_config_t(
@@ -266,12 +277,8 @@ public final class ProfilingEngine: @unchecked Sendable {
 
             let startResult = emb_sampler_start(ringBuffer, config)
 
-            if startResult != EMB_SAMPLER_START_OK, needsNewBuffer {
-                emb_ring_buffer_destroy(ringBuffer)
-                if let ptr = readBuffer { free(ptr) }
-                ringBuffer = nil
-                readBuffer = nil
-                readBufferSize = 0
+            if startResult != EMB_SAMPLER_START_OK, createdNewBacking {
+                tearDownBacking()
             }
 
             switch startResult {
@@ -292,6 +299,130 @@ public final class ProfilingEngine: @unchecked Sendable {
                 }
                 return .samplerStartFailed
             }
+        #endif
+    }
+
+    #if !os(watchOS)
+    /// Tear down whatever backing exists (in-memory ring buffer or file-backed store) + the read buffer.
+    /// For a store, this frees the attached ring-buffer wrapper and unmaps the file; it does NOT delete
+    /// the file or write a tombstone (use `finalizeStorage()` for a clean stop).
+    private func tearDownBacking() {
+        if let s = store {
+            emb_profile_store_destroy(s)  // frees the attached ring-buffer wrapper + unmaps
+            store = nil
+            ringBuffer = nil
+        } else if let existing = ringBuffer {
+            emb_ring_buffer_destroy(existing)
+            ringBuffer = nil
+        }
+        if let ptr = readBuffer { free(ptr) }
+        readBuffer = nil
+        readBufferSize = 0
+        activeConfiguration = nil
+        activeSessionFileName = nil
+    }
+
+    /// Create a fresh in-memory ring buffer + read buffer. Returns false on allocation failure.
+    private func setUpInMemoryBuffer(configuration: ProfilingConfiguration) -> Bool {
+        guard let buffer = emb_ring_buffer_create(Int(configuration.bufferCapacityBytes), nil) else {
+            return false
+        }
+        let capacity = emb_ring_buffer_capacity(buffer)
+        guard let rb = malloc(capacity) else {
+            emb_ring_buffer_destroy(buffer)
+            return false
+        }
+        ringBuffer = buffer
+        readBuffer = UnsafeMutableRawPointer(rb)
+        readBufferSize = capacity
+        return true
+    }
+
+    /// Create a file-backed store + read buffer in `directory`. Returns nil on success, or a failing
+    /// `StartResult` on error (the engine is left with no backing).
+    private func setUpFileBackedBuffer(configuration: ProfilingConfiguration,
+                                       directory: URL,
+                                       sessionId: [UInt8]?) -> StartResult? {
+        do {
+            try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true)
+        } catch {
+            return .storageSetupFailed(reason: "mkdir failed: \(error.localizedDescription)")
+        }
+
+        let sid = sessionId ?? (0..<16).map { _ in UInt8.random(in: 0...255) }
+        guard sid.count == 16 else {
+            return .storageSetupFailed(reason: "sessionId must be 16 bytes")
+        }
+        let fileName = sid.map { String(format: "%02x", $0) }.joined() + ".embprof"
+        let path = directory.appendingPathComponent(fileName).path
+
+        var err: Int32 = 0
+        let storePtr: OpaquePointer? = path.withCString { cpath in
+            sid.withUnsafeBufferPointer { sidBuf in
+                emb_profile_store_create(cpath, Int(configuration.bufferCapacityBytes), sidBuf.baseAddress, &err)
+            }
+        }
+        guard let storePtr else {
+            return .storageSetupFailed(reason: "store create failed (errno \(err))")
+        }
+        guard let buf = emb_profile_store_buffer(storePtr) else {
+            emb_profile_store_destroy(storePtr)
+            return .storageSetupFailed(reason: "store buffer unavailable")
+        }
+        let capacity = emb_ring_buffer_capacity(buf)
+        guard let rb = malloc(capacity) else {
+            emb_profile_store_destroy(storePtr)
+            return .storageSetupFailed(reason: "read-buffer allocation failed (\(capacity) bytes)")
+        }
+        store = storePtr
+        ringBuffer = buf
+        readBuffer = UnsafeMutableRawPointer(rb)
+        readBufferSize = capacity
+        activeSessionFileName = fileName
+        return nil
+    }
+    #endif
+
+    /// Flush persisted samples to disk (`msync(MS_ASYNC)`). No-op in in-memory mode. The CaptureService
+    /// wrapper calls this on `willResignActive` / `didEnterBackground` (and `willTerminate`) so samples
+    /// are durable before a possible background Jetsam kill. Cheap, non-blocking, off the hot path.
+    ///
+    /// - Note: Best-effort — it skips silently if the engine's gate is briefly contended (e.g. a
+    ///   concurrent `start`/`retrieveSamples`). `MAP_SHARED` pages are still flushed by the kernel on
+    ///   normal termination; this only matters for an abrupt Jetsam within that small window.
+    /// - Note: `msync` runs concurrently with the live sampler writing the same mapping — this is by
+    ///   design. A page captured mid-write is just re-flushed later, and recovery's seqlock/torn-tail
+    ///   handling discards any partial record, so a torn flush is never observable as corruption.
+    public func flush() {
+        #if !os(watchOS)
+            // Gate-protected: `store` is torn down (destroyed + niled) under the gate by start()/reset.
+            // Without this, a background/terminate-thread flush could msync a freed/unmapped store.
+            guard acquireGate() else { return }
+            defer { releaseGate() }
+            if let s = store { emb_profile_store_flush(s) }
+        #endif
+    }
+
+    /// Mark the file-backed session cleanly finalized (writes the version-0 tombstone), so recovery
+    /// reports nothing for it. No-op in in-memory mode. Call on a clean stop, after the worker has
+    /// exited (poll `isActive` until false) — the wrapper sequences this.
+    ///
+    /// - Important: This DISCARDS the session's on-disk samples — the tombstone marks the *entire* file
+    ///   "disregard". Only call it once you have retrieved everything you need (e.g. via
+    ///   ``retrieveSamples(from:through:)``); there is no recovery after finalize.
+    /// - Important: This is a NO-OP while sampling is still active. Writing the tombstone before the
+    ///   worker has drained would silently discard not-yet-reported samples, so if called early it
+    ///   leaves the session recoverable (reported next launch) rather than lost. Conversely, if you
+    ///   never call it on a clean stop, that session is listed by ``recoverableSessions(in:)`` next
+    ///   launch (a harmless false positive you can ``delete(_:)``).
+    public func finalizeStorage() {
+        #if !os(watchOS)
+            // Gate-protected for the same reason as flush(): guards against a concurrent store teardown.
+            guard acquireGate() else { return }
+            defer { releaseGate() }
+            // Refuse to tombstone a live buffer (see note above) — leave it recoverable instead.
+            guard !emb_sampler_is_active() else { return }
+            if let s = store { emb_profile_store_finalize(s) }
         #endif
     }
 
@@ -441,8 +572,8 @@ public final class ProfilingEngine: @unchecked Sendable {
                 // Defense-in-depth: reject implausibly large frame_count values.
                 // The C read path already bounds-checks via record_size > (write_pos -
                 // scan_pos), but a corruption check is cheap.
-                guard header.frame_count <= UInt32(EMB_MAX_STACK_FRAMES) else { break }
-                let recordSize = Int(emb_ring_record_size(header.frame_count))
+                guard Int(header.frame_count) <= Int(EMB_MAX_STACK_FRAMES) else { break }
+                let recordSize = Int(emb_ring_record_size(UInt32(header.frame_count)))
                 guard offset + recordSize <= validEnd else { break }
                 totalFrames += Int(header.frame_count)
                 offset += recordSize
@@ -458,9 +589,9 @@ public final class ProfilingEngine: @unchecked Sendable {
             for _ in 0..<result.record_count {
                 guard offset + headerSize <= validEnd else { break }
                 let header = rawBuffer.load(fromByteOffset: offset, as: emb_ring_record_header_t.self)
-                guard header.frame_count <= UInt32(EMB_MAX_STACK_FRAMES) else { break }
+                guard Int(header.frame_count) <= Int(EMB_MAX_STACK_FRAMES) else { break }
                 let frameCount = Int(header.frame_count)
-                let recordSize = Int(emb_ring_record_size(header.frame_count))
+                let recordSize = Int(emb_ring_record_size(UInt32(header.frame_count)))
                 guard offset + recordSize <= validEnd else { break }
 
                 let framesStart = frames.count
@@ -474,7 +605,8 @@ public final class ProfilingEngine: @unchecked Sendable {
 
                 samples.append(ProfilingSample(
                     timestamp: header.timestamp_ns,
-                    frameRange: framesStart..<frames.count
+                    frameRange: framesStart..<frames.count,
+                    threadState: ThreadState(rawValue: header.thread_state) ?? .unknown
                 ))
 
                 offset += recordSize
--- a/Sources/EmbraceProfiling/ProfilingEngine.swift
+++ b/Sources/EmbraceProfiling/ProfilingEngine.swift
@@ -32,6 +32,9 @@ import Foundation  // URL / FileManager / String(format:) — used in the public
 public final class ProfilingEngine: @unchecked Sendable {
     public static let shared = ProfilingEngine()
 
+    /// File extension (without the leading dot) for a persisted profiling session file.
+    internal static let sessionFileExtension = "embprof"
+
     /// Result of calling ``start(configuration:)``.
     public enum StartResult: Equatable, Sendable {
         /// Successfully started the sampler.
@@ -203,6 +206,9 @@ public final class ProfilingEngine: @unchecked Sendable {
     ///   - sessionId: opaque 16-byte id for the file-backed session (becomes the filename). Must be
     ///     unique per launch. Ignored for in-memory mode; a random id is used if nil.
     ///
+    /// - Important: A file-backed start (`directory != nil`) performs synchronous filesystem work on the
+    ///   calling thread (`createDirectory`, `ftruncate`, `mmap`) and never dispatches internally — call
+    ///   it off the main thread. In-memory mode (the default) does no filesystem I/O.
     /// - Note: For a file-backed session, on a *clean* stop call ``finalizeStorage()`` after the worker
     ///   has exited (poll ``isActive`` until `false`). Otherwise that session looks crash-like and is
     ///   listed by ``recoverableSessions(in:)`` on the next launch.
@@ -353,7 +359,7 @@ public final class ProfilingEngine: @unchecked Sendable {
         guard sid.count == 16 else {
             return .storageSetupFailed(reason: "sessionId must be 16 bytes")
         }
-        let fileName = sid.map { String(format: "%02x", $0) }.joined() + ".embprof"
+        let fileName = sid.map { String(format: "%02x", $0) }.joined() + "." + Self.sessionFileExtension
         let path = directory.appendingPathComponent(fileName).path
 
         var err: Int32 = 0
@@ -381,7 +387,7 @@ public final class ProfilingEngine: @unchecked Sendable {
         activeSessionFileName = fileName
         return nil
     }
-    #endif
+    #endif  // !os(watchOS)
 
     /// Flush persisted samples to disk (`msync(MS_ASYNC)`). No-op in in-memory mode. The CaptureService
     /// wrapper calls this on `willResignActive` / `didEnterBackground` (and `willTerminate`) so samples
--- a/Sources/EmbraceProfiling/ProfilingEngine.swift
+++ b/Sources/EmbraceProfiling/ProfilingEngine.swift
@@ -6,8 +6,13 @@ import Foundation  // URL / FileManager / String(format:) — used in the public
 
 #if !os(watchOS)
     import Darwin
-    import EmbraceAtomicsShim
-    import EmbraceProfilingSampler
+
+    // Under CocoaPods every subspec compiles into the single EmbraceIO module, so these
+    // sibling modules must not be imported by name — see EmbraceIO.podspec.
+    #if !EMBRACE_COCOAPOD_BUILDING_SDK
+        import EmbraceAtomicsShim
+        import EmbraceProfilingSampler
+    #endif
 #endif
 
 /// A sampling-based profiling engine that captures main thread stack traces
@@ -154,7 +159,8 @@ public final class ProfilingEngine: @unchecked Sendable {
     }
 
     #if !os(watchOS)
-    /// Try to acquire the gate.
+    /// Try to acquire the backing gate — the CAS flag guarding the engine's backing store and buffers
+    /// (`ringBuffer`, `readBuffer`, `store`, …). Unrelated to the C sampler's own run-state machine.
     /// Sleeps ``gateSleepNs`` between attempts, up to ``gateTimeoutNs``
     /// cumulative actual elapsed time.
     ///
@@ -163,7 +169,7 @@ public final class ProfilingEngine: @unchecked Sendable {
     /// 2-3 CAS attempts within the 10ms deadline rather than the ~100 attempts
     /// that the arithmetic suggests. This is acceptable because contention is
     /// expected to be extremely rare (only start/retrieve overlap).
-    internal func acquireGate() -> Bool {
+    internal func acquireBackingGate() -> Bool {
         let deadlineNs = clock_gettime_nsec_np(CLOCK_UPTIME_RAW) + Self.gateTimeoutNs
         while true {
             var expected: Bool = false
@@ -177,8 +183,8 @@ public final class ProfilingEngine: @unchecked Sendable {
         }
     }
 
-    /// Release the gate.
-    internal func releaseGate() {
+    /// Release the backing gate.
+    internal func releaseBackingGate() {
         emb_atomic_bool_store(gateIsAcquired, false, .release)
     }
     #endif
@@ -227,12 +233,20 @@ public final class ProfilingEngine: @unchecked Sendable {
                 return .faulted(reason: reason)
             }
 
-            guard acquireGate() else {
+            guard acquireBackingGate() else {
                 return .operationInProgress
             }
-            defer { releaseGate() }
+            defer { releaseBackingGate() }
 
             let wasActive = emb_sampler_is_active()
+
+            // KNOWN RACE (dormant; no production caller of start() yet): is_active() is true for
+            // STOPPING, and this Swift gate doesn't block the C worker's autonomous STOPPING→ZOMBIE
+            // transition. If the worker finishes here, emb_sampler_start() below reaps the zombie and
+            // wins a fresh CAS, so a genuinely new session reports .alreadyActive — and in file-backed
+            // mode skips the setUpFileBackedBuffer branch, dropping the caller's sessionId. The real fix
+            // is to have the C layer distinguish "fresh start" from "already running" in its return
+            // instead of re-reading is_active() here; tracked as a follow-up.
 
             // Decide the backing. If the sampler is already active, leave it untouched (the worker
             // may be writing) and let emb_sampler_start() report the accurate state.
@@ -403,8 +417,8 @@ public final class ProfilingEngine: @unchecked Sendable {
         #if !os(watchOS)
             // Gate-protected: `store` is torn down (destroyed + niled) under the gate by start()/reset.
             // Without this, a background/terminate-thread flush could msync a freed/unmapped store.
-            guard acquireGate() else { return }
-            defer { releaseGate() }
+            guard acquireBackingGate() else { return }
+            defer { releaseBackingGate() }
             if let s = store { emb_profile_store_flush(s) }
         #endif
     }
@@ -424,8 +438,8 @@ public final class ProfilingEngine: @unchecked Sendable {
     public func finalizeStorage() {
         #if !os(watchOS)
             // Gate-protected for the same reason as flush(): guards against a concurrent store teardown.
-            guard acquireGate() else { return }
-            defer { releaseGate() }
+            guard acquireBackingGate() else { return }
+            defer { releaseBackingGate() }
             // Refuse to tombstone a live buffer (see note above) — leave it recoverable instead.
             guard !emb_sampler_is_active() else { return }
             if let s = store { emb_profile_store_finalize(s) }
@@ -546,8 +560,8 @@ public final class ProfilingEngine: @unchecked Sendable {
                 return .faulted(reason: reason)
             }
 
-            guard acquireGate() else { return .busy }
-            defer { releaseGate() }
+            guard acquireBackingGate() else { return .busy }
+            defer { releaseBackingGate() }
 
             guard let ringBuffer, let readBuffer else {
                 return .notStarted
--- a/Sources/EmbraceProfiling/ProfilingRecovery.swift
+++ b/Sources/EmbraceProfiling/ProfilingRecovery.swift
@@ -5,7 +5,10 @@
 import Foundation
 
 #if !os(watchOS)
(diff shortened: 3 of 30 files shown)


Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -183,8 +193,23 @@ public final class ProfilingEngine: @unchecked Sendable {
...
+    ///     unique per launch. Ignored for in-memory mode; a random id is used if nil.
+    ///
+    /// - Note: For a file-backed session, on a *clean* stop call ``finalizeStorage()`` after the worker
+    ///   has exited (poll ``isActive`` until `false`). Otherwise that session looks crash-like and is
+    ///   listed by ``recoverableSessions(in:)`` on the next launch.
     @discardableResult
-    public func start(configuration: ProfilingConfiguration = ProfilingConfiguration()) -> StartResult {
+    public func start(configuration: ProfilingConfiguration = ProfilingConfiguration(),
+                      directory: URL? = nil,
+                      sessionId: [UInt8]? = nil) -> StartResult {
```
fnewberg, line 224: can we modify the `sessionId ` or is it immutable once the profile engine has started?

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -203,56 +228,42 @@ public final class ProfilingEngine: @unchecked Sendable {
 
             let wasActive = emb_sampler_is_active()
```
fnewberg, line 241: nitpick:  we take a snapshot of `wasActive` here under the gate, but the worker can transition active→inactive any time after this read. while not a memory-safety issue given what we do later, on lines 287 and 293 we use this potentially stale `wasActive` to choose .alreadyActive vs .started / .samplerBusy. If the worker stops in the window between here and line 278, a call that actually (re)started a session can still report .alreadyActive (or vice-versa). purely cosmetic today, but any caller that branches
 on the `StartResult` would be misled.

not sure if this is worth addressing given the narrow window and limited value, but thought i'd mention it

Sources/EmbraceProfiling/ProfilingEngine.swift
fnewberg, line 201: File-backed mode (directory != nil) does synchronous, blocking syscalls on the calling thread —createDirectory, ftruncate, and several mmaps — and start() never dispatches internally. 


```suggestion
    ///
    /// - Important: File-backed mode (`directory != nil`) does synchronous filesystem work on the
    ///   calling thread (`createDirectory`, `ftruncate`, `mmap`) and never dispatches internally.
    ///   Call it off the main thread for file-backed starts In-memory mode (the default) does no 
    ///   filesystem I/O.
    ///
```

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -183,8 +193,23 @@ public final class ProfilingEngine: @unchecked Sendable {
     /// - Parameter configuration: Profiling parameters. Defaults are suitable
     ///   for most use cases (10 Hz, 1MB buffer).
     /// - Returns: A ``StartResult`` indicating the outcome.
+    /// Start sampling.
+    ///
+    /// - Parameters:
+    ///   - configuration: profiling parameters.
+    ///   - directory: when non-nil, samples are persisted to a file-backed buffer in this directory
```
fnewberg, line 209: do you think we should use a consistent directory and rely on the filenames being unique? seems like that would make it easier to do clean up and be slightly less overhead when starting the sampler

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -295,6 +302,130 @@ public final class ProfilingEngine: @unchecked Sendable {
...
+            return .storageSetupFailed(reason: "read-buffer allocation failed (\(capacity) bytes)")
+        }
+        store = storePtr
+        ringBuffer = buf
+        readBuffer = UnsafeMutableRawPointer(rb)
+        readBufferSize = capacity
+        activeSessionFileName = fileName
+        return nil
+    }
+    #endif
```
fnewberg:
```suggestion
    #endif  // !os(watchOS)
```

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -295,6 +302,130 @@ public final class ProfilingEngine: @unchecked Sendable {
...
+    ///   concurrent `start`/`retrieveSamples`). `MAP_SHARED` pages are still flushed by the kernel on
+    ///   normal termination; this only matters for an abrupt Jetsam within that small window.
+    /// - Note: `msync` runs concurrently with the live sampler writing the same mapping — this is by
+    ///   design. A page captured mid-write is just re-flushed later, and recovery's seqlock/torn-tail
+    ///   handling discards any partial record, so a torn flush is never observable as corruption.
+    public func flush() {
+        #if !os(watchOS)
+            // Gate-protected: `store` is torn down (destroyed + niled) under the gate by start()/reset.
+            // Without this, a background/terminate-thread flush could msync a freed/unmapped store.
+            guard acquireGate() else { return }
```
fnewberg: do you think it would be clearer if we named these `acquireStateGate` and `releaseStateGate` to make it clear that they have nothing to do with the actual capture of samples?

Sources/EmbraceProfiling/ProfilingEngine.swift
```diff
@@ -295,6 +302,130 @@ public final class ProfilingEngine: @unchecked Sendable {
...
+            try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true)
+        } catch {
+            return .storageSetupFailed(reason: "mkdir failed: \(error.localizedDescription)")
+        }
+
+        let sid = sessionId ?? (0..<16).map { _ in UInt8.random(in: 0...255) }
+        guard sid.count == 16 else {
+            return .storageSetupFailed(reason: "sessionId must be 16 bytes")
+        }
+        let fileName = sid.map { String(format: "%02x", $0) }.joined() + ".embprof"
```
fnewberg: nitpick: should we define `embrof` as a constant since it's used in multiple places

Sources/EmbraceProfiling/ProfilingRecovery.swift
```diff
@@ -0,0 +1,165 @@
...
+            if acquireGate() { activeName = activeSessionFileName; releaseGate() }
+
+            var sessions: [RecoverableSession] = []
+            for url in entries where url.pathExtension == "embprof" && url.lastPathComponent != activeName {
+                let peek = url.path.withCString { emb_profile_peek($0) }
+                let status: RecoverableSession.Status
+                switch peek {
+                case EMB_PROFILE_PEEK_RECOVERABLE: status = .recoverable
+                case EMB_PROFILE_PEEK_FINALIZED: status = .finalized
+                default: continue  // not ours / unreadable → skip
```
fnewberg: if we ended up with corrupted/unknown files in this directory, how would we ever find out so that we could delete them? should we consider some method to bubble up those files/paths?

fnewberg: should we consider adding something that treats the dir as a FIFO, so we can put a max limit on number of retained files and delete the oldest ones first if we hit a limit?

Sources/EmbraceProfiling/ProfilingEngine.swift
fnewberg: thoughts on also including the creation timestamp in the filename or are we better off just getting that data for each file if we want a creation-time-ordered list?

Sources/EmbraceProfiling/ProfilingRecovery.swift
```diff
@@ -0,0 +1,165 @@
...
+        /// Has (or may have) un-reported samples — worth recovering.
+        case recoverable
+        /// Cleanly finalized (the app stopped normally); nothing to recover, safe to delete.
+        case finalized
+    }
+
+    /// Hex session id (the file-name stem).
+    public let sessionId: String
+    /// When the file was last written (its most recent sample or the finalize tombstone).
+    public let lastModified: Date
```
fnewberg, line 31: nitpick: should we consider making this and `byteSize` nillable since we have a scenario further down where we do

```
 lastModified: values?.contentModificationDate ?? .distantPast,
byteSize: values?.fileSize ?? 0,
```

if these were nillable we could tell the difference between `0` and "wasn't able to get filesize"

Sources/EmbraceProfilingSampler/include/emb_profile_store.h
```diff
@@ -0,0 +1,167 @@
...
+/// File-backed persistence for the profiling ring buffer.
+///
+/// `emb_profile_store` owns a per-session file and the memory mapping over it, and
+/// hands back an `emb_ring_buffer_t` (via `emb_ring_buffer_attach`) wired to the
+/// mapped data region + a mapped footer control block. The sampler writes into that
+/// ring buffer exactly as it does for an in-memory buffer; the kernel flushes the
+/// dirty `MAP_SHARED` pages to the file, so samples survive abrupt process death and
+/// can be recovered on the next launch.
+///
+/// See `design/PROFILING-DISK-FORMAT.md` for the on-disk format. Recovery (reading a
```
fnewberg: is this a file you meant to commit?

Sources/EmbraceProfilingSampler/include/emb_profile_store.h
```diff
@@ -0,0 +1,167 @@
...
+// mid-mutation (create/reset) and as the clean-shutdown / finalized tombstone. Because the identity
+// layout is frozen, version 0 means "ignore" regardless of any future format change.
+#define EMB_PROFILE_FORMAT_INVALID 0u
+
+/// Write-once static descriptor (written at create).
+typedef struct {
+    uint64_t created_uptime_ns;   // CLOCK_MONOTONIC_RAW at creation
+    uint64_t created_wall_ns;     // wall clock at creation (session correlation)
+    uint8_t  session_id[16];      // opaque 128-bit id supplied by the caller
+    uint64_t image_table_offset;  // 0 if absent (always 0 in v1)
```
fnewberg: are these for some future feature?

Sources/EmbraceProfilingSampler/include/emb_profile_store.h
```diff
@@ -0,0 +1,167 @@
...
+    uint64_t image_table_bytes;   // 0 if absent (always 0 in v1)
+} emb_profile_descriptor_t;
+
+/// FROZEN identity — the LAST bytes of the file. Same layout in every format version, forever.
```
fnewberg, line 52: why put this at the end rather than at the start?

Sources/EmbraceProfilingSampler/include/emb_profile_store.h
```diff
@@ -0,0 +1,167 @@
...
+
+/// Version-specific trailer — sits immediately before the identity struct. v1 layout.
+typedef struct {
+    uint64_t footer_offset;       // == data_capacity; where the footer begins
+    uint64_t data_capacity;       // ring data region size in bytes (page-aligned)
+    uint32_t page_size;           // page size used at creation
+    uint16_t record_header_bytes; // sizeof(emb_ring_record_header_t), validated on recovery
+    uint16_t pointer_bytes;       // sizeof(uintptr_t) == 8
+    uint32_t footer_bytes;        // total footer size
+    uint32_t trailer_bytes;       // sizeof(this struct) for THIS version
```
fnewberg: is this included so that an SDK that supports v2 could read v1 or why would we not just get this with `sizeof`?

Sources/EmbraceProfilingSampler/include/emb_profile_store.h
```diff
@@ -0,0 +1,167 @@
...
+void emb_profile_store_finalize(emb_profile_store_t *store);
+
+/// `msync(MS_ASYNC)` the mapped data + footer, so samples are durable before a possible
+/// background Jetsam kill. Cheap, non-blocking, off the hot path. NOT async-safe (call from
+/// the main/notification thread on background/terminate).
+void emb_profile_store_flush(emb_profile_store_t *store);
+
+/// Tear down the store: destroy the ring buffer wrapper, unmap the reservation, close the
+/// file. Does not delete the file (Embrace owns deletion). Safe to call with NULL.
+void emb_profile_store_destroy(emb_profile_store_t *store);
```
fnewberg, line 123: should we note if this does or doesn't require a call to `emb_profile_store_flush` before being called?

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
```diff
@@ -60,19 +61,63 @@ typedef enum {
 extern "C" {
 #endif
 
+/// Control block holding the writer/reader positions. Relocated out of
+/// emb_ring_buffer_t so it can be backed either by a small anonymous allocation
+/// (in-memory buffers) or by a file-backed mapped footer page (persistent buffers,
+/// PROFILING-DISK-FORMAT.md §3.1) — the same struct, only the backing differs.
```
fnewberg: remove claude-generated reference to local file 🙂

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
```diff
@@ -60,19 +61,63 @@ typedef enum {
...
 typedef struct {
-    uint8_t *data;               // Double-mapped virtual region (2 × capacity).
-    size_t capacity;             // Usable buffer size in bytes (page-aligned).
-    uint64_t next_seq;           // Writer-local seqlock counter (single-writer).
     _Atomic uint64_t write_pos;  // Monotonically increasing write position.
     _Atomic uint64_t oldest_pos; // Position of the oldest surviving record.
+    uint64_t next_seq;           // Writer-local seqlock counter (single-writer).
+    _Atomic uint32_t status_flags; // Reserved (file-backed lifecycle is signaled via the footer
+                                 // format_version, not here — PROFILING-DISK-FORMAT.md §8).
```
fnewberg: same comment as above

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
```diff
@@ -60,19 +61,63 @@ typedef enum {
...
                                  // both sides use seq_cst to prevent ARM64 store-buffer
                                  // reordering. See emb_ring_buffer_reset and read_range.
 } emb_ring_buffer_t;
 
+/// Hard upper limit on stack frames per sample. Defined here (not in emb_sampler.h)
+/// because the ring write path validates `frame_count` against it; emb_sampler.h
+/// includes this header and consumes the value from here.
+#define EMB_MAX_STACK_FRAMES 1024
+
+/// Main-thread run state captured per sample (see THREAD-STATE.md). Values mirror
```
fnewberg: fix reference to uncommitted doc

Sources/EmbraceProfiling/ProfilingSample.swift
```diff
@@ -7,16 +7,34 @@
 /// Frame addresses are stored in a flat array shared across all samples in a
 /// ``ProfilingResult``. Use ``frameRange`` to index into that array, or use
 /// ``ProfilingResult/frames(for:)`` for convenient per-sample access.
+/// Run state of the sampled (main) thread at the moment a sample was captured.
+///
+/// Raw values mirror the Mach `TH_STATE_*` constants (and the C
+/// `emb_thread_run_state_t`). `.unknown` (255) means the state could not be
+/// captured (e.g. `thread_info` failed).
+public enum ThreadState: UInt8, Sendable {
```
fnewberg, line 15: this class and `emb_thread_run_state_t` (`emb_ring_buffer.h:98`) both maintain these constants.  to avoid them drifting, should we add a small test anchoring the Swift enum to the same source of truth the C enum already uses, so both independently mirror TH_STATE_* and can't drift? e.g.


```
func test_threadState_rawValuesMatchMachConstants() {
    XCTAssertEqual(ThreadState.running.rawValue,         UInt8(TH_STATE_RUNNING))
    XCTAssertEqual(ThreadState.stopped.rawValue,         UInt8(TH_STATE_STOPPED))
    XCTAssertEqual(ThreadState.waiting.rawValue,         UInt8(TH_STATE_WAITING))
    XCTAssertEqual(ThreadState.uninterruptible.rawValue, UInt8(TH_STATE_UNINTERRUPTIBLE))
    XCTAssertEqual(ThreadState.halted.rawValue,          UInt8(TH_STATE_HALTED))
}
```

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
```diff
@@ -84,16 +129,23 @@ typedef struct {
...
+    uint64_t timestamp_ns;  // Monotonic timestamp (nanoseconds).
 } emb_ring_record_header_t;
 
 // The record layout assumes 64-bit pointers (uintptr_t == 8 bytes).
 // The Swift layer reads frame data as UInt (also 8 bytes on 64-bit).
 // Catch any hypothetical 32-bit platform at compile time.
 _Static_assert(sizeof(uintptr_t) == 8,
                "Ring buffer record layout requires 64-bit pointers");
+// The header must stay exactly 16 bytes: seq(4) + frame_count(2) + thread_state(1)
+// + flags(1) + timestamp_ns(8). The on-disk format (PROFILING-DISK-FORMAT.md §3.2)
```
fnewberg: remove local file reference

Sources/EmbraceProfilingSampler/include/emb_sampler.h
```diff
@@ -59,8 +59,8 @@
 extern "C" {
 #endif
 
-/// Hard upper limit on stack frames per sample.
-#define EMB_MAX_STACK_FRAMES 1024
+// EMB_MAX_STACK_FRAMES is defined in emb_ring_buffer.h (included above) — the ring
```
fnewberg: do we need this comment?

Sources/EmbraceProfilingSampler/include/emb_stack_walker.h
```diff
@@ -56,13 +56,17 @@ extern "C" {
 /// @param frames_out   Output buffer for frame addresses.
 /// @param max_frames   Capacity of frames_out.
 /// @param count_out    On return, the number of frames captured.
+/// @param is_truncated Optional (may be NULL). On return, true if the walk stopped
```
fnewberg, line 59: what's the value of having this be nullable? where do we need `null` where `true` and `false` don't suffice?

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <time.h>
+#include <unistd.h>
+
+// Data-protection Class B == NSFileProtectionCompleteUnlessOpen. The protection-class enum constants
+// (PROTECTION_CLASS_*) are NOT exported by the public SDK headers — only F_SETPROTECTIONCLASS is — so
+// we use the documented integer value directly.
+#define EMB_PROTECTION_CLASS_B 2
```
fnewberg, line 24: were there more protection class definitions at one point? why do we have `B` and nothing else?

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+    size_t page;              // vm page size used at creation
+    void *footer;             // start of the mapped footer ([control page][metadata page]); for flush
+    void *meta_page;          // the write-once metadata page (mprotect-toggled for version writes)
+    emb_profile_ident_t *ident; // frozen identity within the metadata page
+    emb_ring_buffer_t *ring;  // attached ring buffer (owns_resources == false)
+};
+
+/// Write `version` into the frozen identity, briefly toggling the write-once metadata
+/// page to RW. The version field is the file's transactional validity/tombstone marker
+/// (PROFILING-DISK-FORMAT.md §4.2): 0 = "disregard"; a real version = recoverable.
```
fnewberg: remove reference to local file

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+static void store_set_version(emb_profile_store_t *store, uint64_t version)
+{
+    // Toggle the write-once metadata page to RW just for this write. If the unprotect fails, the page
+    // is still read-only — writing would SIGBUS and crash the host app — so bail and leave the current
+    // version intact. Worst case the file keeps version 1 and is re-reported as a crash next launch
+    // (a harmless false positive), never a fault.
```
fnewberg: can you elaborate on why this would be harmless?

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+    emb_profile_ident_t *ident; // frozen identity within the metadata page
+    emb_ring_buffer_t *ring;  // attached ring buffer (owns_resources == false)
+};
+
+/// Write `version` into the frozen identity, briefly toggling the write-once metadata
+/// page to RW. The version field is the file's transactional validity/tombstone marker
+/// (PROFILING-DISK-FORMAT.md §4.2): 0 = "disregard"; a real version = recoverable.
+static void store_set_version(emb_profile_store_t *store, uint64_t version)
+{
+    // Toggle the write-once metadata page to RW just for this write. If the unprotect fails, the page
```
fnewberg: is this failure ever expected or are we just being defensive?

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+        *errno_out = EINVAL;
+        return NULL;
+    }
+
+    const size_t page = (size_t)vm_page_size;
+    const size_t data_capacity = round_up_to_page(capacity_bytes, page);
+    // Footer = one RW control-block page + one write-once metadata page.
+    const size_t footer_bytes = 2 * page;
+    const size_t guard_bytes = page;
+    const off_t  file_size = (off_t)(data_capacity + footer_bytes);
```
fnewberg, line 78: does this mean we should expect each session file to be (3*16kb + data capacity), so maybe somewhere in the 100-200kB range?

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+    if (base == MAP_FAILED) { *errno_out = errno; close(fd); return NULL; }
+
+    uint8_t *b = (uint8_t *)base;
+    // Lower half + aliased upper half: both MAP_SHARED over file offset 0, so a record that
+    // wraps past the end is written through the alias into the single file region.
+    if (mmap(b, data_capacity, PROT_READ | PROT_WRITE,
+             MAP_SHARED | MAP_FIXED, fd, 0) == MAP_FAILED) { goto fail_mmap; }
+    if (mmap(b + data_capacity, data_capacity, PROT_READ | PROT_WRITE,
+             MAP_SHARED | MAP_FIXED, fd, 0) == MAP_FAILED) { goto fail_mmap; }
+    // Guard page at b + 2*data_capacity stays PROT_NONE (never mapped over) — §4.1 Layer 1.
```
fnewberg: remove local doc reference

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+    store->data_capacity = data_capacity;
+    store->footer_bytes  = footer_bytes;
+    store->page          = page;
+    store->footer        = footer;
+    store->meta_page     = meta;
+    store->ident         = ident;
+    store->ring          = ring;
+    return store;
+
+fail_mmap:
```
fnewberg: would it more correct to just call this `cleanup`? e.g. we can end up here if a `calloc` fails

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+
+emb_ring_buffer_t *emb_profile_store_buffer(emb_profile_store_t *store)
+{
+    return store ? store->ring : NULL;
+}
+
+bool emb_profile_store_reset(emb_profile_store_t *store)
+{
+    if (store == NULL) { return false; }
+    // Transactional reset (PROFILING-DISK-FORMAT.md §4.2 / audit B3): bracket the mutation with the
```
fnewberg: remove local doc reference

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+    bool ok = emb_ring_buffer_reset(store->ring);  // memset data + reset positions (reader-safe)
+    store_set_version(store, EMB_PROFILE_FORMAT_VER);  // consistent again (also restores on reader-blocked reset)
+    return ok;
+}
+
+void emb_profile_store_finalize(emb_profile_store_t *store)
+{
+    if (store == NULL) { return; }
+    // Clean-stop tombstone (audit B2): version → 0 marks the file "finalized" so recovery reports
+    // nothing for it (its samples were already drained live). We never delete — Embrace owns deletion.
```
fnewberg: what does "drained live" mean?

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+}
+
+void emb_profile_store_flush(emb_profile_store_t *store)
+{
+    if (store == NULL) { return; }
+    // Async flush of dirty pages to the file ahead of a possible background Jetsam kill. The lower data
+    // half covers the file data region (the upper half aliases the same file pages); the footer holds the
+    // control block + metadata. MS_ASYNC: schedule writeback, don't block.
+    msync(store->base, store->data_capacity, MS_ASYNC);
+    msync(store->footer, store->footer_bytes, MS_ASYNC);  // exact mapped footer base (no guard-size coupling)
```
fnewberg, line 218: would we be better of synching the footer data first?

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+        goto done;  // too small to be valid → CORRUPT
+    }
+
+    // --- frozen identity at EOF − 16: magic + version gate ---
+    emb_profile_ident_t ident;
+    if (!pread_exact(fd, &ident, sizeof(ident), file_size - (off_t)sizeof(ident))) {
+        status = EMB_PROFILE_RECOVER_IO_ERROR; goto done;
+    }
+    if (ident.magic != EMB_PROFILE_FILE_MAGIC) { status = EMB_PROFILE_RECOVER_NOT_OURS; goto done; }
+    if (ident.format_version == EMB_PROFILE_FORMAT_INVALID) { status = EMB_PROFILE_RECOVER_FINALIZED; goto done; }
```
fnewberg, line 299: should we have a dedicated format version for finalized? i don't think it's a logical issue, but the readability of "invalid means finalized" isn't the clearest

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+    if (tr.page_size != (uint32_t)vm_page_size ||
+        tr.record_header_bytes != (uint16_t)sizeof(emb_ring_record_header_t) ||
+        tr.pointer_bytes != (uint16_t)sizeof(uintptr_t) ||
+        tr.data_capacity == 0 ||
+        (tr.data_capacity % (uint64_t)vm_page_size) != 0 ||        // must be page-aligned
+        tr.data_capacity > (uint64_t)file_size ||                  // can't exceed the real file (anti-overflow)
+        tr.footer_bytes < 2 * (uint64_t)vm_page_size ||            // at least control + metadata page
+        tr.footer_offset != tr.data_capacity ||
+        (off_t)(tr.data_capacity + tr.footer_bytes) != file_size) {
+        goto done;  // mismatch → CORRUPT (page-size / record-size / endian / out-of-range rejected here)
```
fnewberg, line 317: there are a lot of things that fall under "corrupt". should we consider at least adding logs to show what was corrupt?

Sources/EmbraceProfiling/ProfilingRecovery.swift
```diff
@@ -0,0 +1,165 @@
...
+            }
+            let keys: [URLResourceKey] = [.contentModificationDateKey, .fileSizeKey]
+            guard let entries = try? fm.contentsOfDirectory(
+                at: directory, includingPropertiesForKeys: keys) else {
+                return []
+            }
+
+            // Snapshot the active file name under the gate so we don't race start()/teardown.
+            var activeName: String?
+            if acquireGate() { activeName = activeSessionFileName; releaseGate() }
```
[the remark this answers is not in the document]
fnewberg: +1 to this

Sources/EmbraceProfiling/ProfilingRecovery.swift
```diff
@@ -0,0 +1,165 @@
...
+    /// `true` if the file was removed.
+    @discardableResult
+    public func delete(_ session: RecoverableSession) -> Bool {
+        #if os(watchOS)
+            return false
+        #else
+            var activeName: String?
+            if acquireGate() { activeName = activeSessionFileName; releaseGate() }
+            guard session.url.lastPathComponent != activeName else { return false }
+            return (try? FileManager.default.removeItem(at: session.url)) != nil
```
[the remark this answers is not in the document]
fnewberg: +1 to this

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+        if (rsize > write_pos - pos) { break; }                // record overruns committed data
+        if (emit != NULL) {
+            const uintptr_t *frames = (const uintptr_t *)((const uint8_t *)h + hdr_size);
+            emit(ctx, h->timestamp_ns, h->thread_state, h->flags, frames, h->frame_count);
+        }
+        pos += rsize;
+    }
+
+    munmap(base, map_reserve);
+    status = EMB_PROFILE_RECOVER_OK;
```
fnewberg, line 379: we can get here after having hit `break`s in the loop above

- should we be continuing for all those break checks, e.g. `h->frame_count > EMB_MAX_STACK_FRAMES)`
- should we be returning `OK` if some of those `break`s have been hit?

Sources/EmbraceProfiling/ProfilingRecovery.swift
```diff
@@ -0,0 +1,165 @@
...
+}
+
+/// Outcome of recovering a single ``RecoverableSession``.
+public enum SessionRecoveryResult: Sendable {
+    /// Parsed successfully; the (possibly empty) samples for this session.
+    case recovered(ProfilingResult)
+    /// The file was cleanly finalized — nothing to recover.
+    case finalized
+    /// The file could not be parsed (not ours / unsupported / corrupt / I/O error).
+    case unreadable(reason: String)
```
fnewberg: `emb_profile_recover`  returns precise, typed statuses — `EMB_PROFILE_RECOVER_NOT_OURS` / `_UNSUPPORTED` / `_CORRUPT` / `_IO_ERROR` (`emb_profile_store.h:117-124`) — but `describe()`
(line 70) collapses them into a human-readable String before they reach the caller. So a caller can't distinguish the cases that warrant different actions: "corrupt → delete() it," "transient I/O error → leave it, maybe retry next launch," "unsupported version → keep it for a future build that understands the format." All they get is a log string.

Could we surface a typed reason instead — `.unreadable(RecoveryFailure)` where `RecoveryFailure` mirrors the C statuses (`.notOurs` / `.unsupportedVersion` / `.corrupt` / `.ioError`)? The mapping from
`emb_profile_recover_status_t` is already centralized in `describe()`, so this is a localized change — swap that function from → `String` to → `RecoveryFailure`.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+}
+
+// MARK: - Recovery (read-only)
+
+static bool pread_exact(int fd, void *buf, size_t n, off_t off)
+{
+    return pread(fd, buf, n, off) == (ssize_t)n;
+}
+
+emb_profile_recover_status_t emb_profile_recover(const char *path,
```
fnewberg, line 267: `emb_profile_store_create` takes an `errno_out` and threads the real `errno` all the way up to `.storageSetupFailed(reason: "... errno N")`. `emb_profile_recover` thread this through so every open / fstat / pread / mmap failure collapses to an `EMB_PROFILE_RECOVER_IO_ERROR`, which `describe()` turns into the static string "I/O error". When recovery fails, the actual cause (EACCES, ENOMEM, EIO, EINTR) is gone, making debugging harder.

Two related issues in the same path:

- `pread_exact` (line 221) conflates two different failures. A short read (truncated/torn file — genuinely CORRUPT) and pread returning -1 (a real I/O error). both funnel to `IO_ERROR`. A truncated crash file — the artifact this feature exists to handle — gets misclassified as an I/O error rather than corruption.
- No EINTR retry. A signal landing during the launch-time recovery read yields a spurious IO_ERROR and the session is reported .unreadable, even though a retry would have succeeded.

Suggestions: 

- add an `errno_out` to `emb_profile_recover` (mirror create)
- have `pread_exact` distinguish short-read → CORRUPT from -1 → `IO_ERROR`, retry on `EINTR`, and thread the `errno` into the `.unreadable` reason. This pairs with the typed-`RecoveryFailure` suggestion — the `errno` rides along on the `.ioError` case.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+
+    munmap(base, map_reserve);
+    status = EMB_PROFILE_RECOVER_OK;
+
+done:
+    close(fd);
+    return status;
+}
+
+emb_profile_peek_status_t emb_profile_peek(const char *path)
```
fnewberg, line 386: `emb_profile_peek` maps every failure to `NOT_OURS`: 
  - open failure (line 333)
  - via the default case: fstat failure, non-regular file, too-small file, pread failure, and a genuine magic mismatch

this is consumed at `ProfilingRecovery.swift`:112-116: a `NOT_OURS` result hits `default: continue`, so the file is silently skipped.

[from claude -- not sure how realistic it is to see these read failures]

a transient read failure on a file that is actually ours and recoverable — briefly locked under data-protection (Class B, device locked at launch), an EINTR, a momentary EIO — is indistinguishable from "not our file." Because peek is the only gate into the recoverable list, that session is dropped, never retried, and never recovered. 

Suggest distinguishing "definitely not ours" (regular file, readable, magic mismatch → NOT_OURS) from "couldn't determine" (open/fstat/pread failed → a separate indeterminate status). `recoverableSessions` should not permanently skip the indeterminate case — either keep/re-list it or at minimum log() it so a dropped-but-recoverable session leaves a breadcrumb. Retrying EINTR here (same as the recover path) would also help.

Sources/EmbraceProfiling/ProfilingEngine.swift
fnewberg, line 224: The C layer types this exactly — `session_id[16]` with a `_Static_assert(sizeof(...) == 16)` but the Swift wrapper loosens it to [UInt8]?, so "16 bytes" survives only as a doc comment plus a runtime sid.count == 16 guard that surfaces as .storageSetupFailed(reason: "sessionId must be 16 bytes") (line ~353). A wrong-length id is caught at start time, not at the call site.

Two more rough edges on the same parameter:

- Input/output type asymmetry. start takes `[UInt8]`, but `RecoverableSession.sessionId` comes back as a hex String. There's no first-class "same session id" type, so a caller can't round-trip the id they started with to the handle they recover without hand-rolling hex encode/decode.

[claude added the following, but i'm torn about it. should we make the core SDK pass an ID to the profile engine and put the burden of not having an ID on the core SDK rather than the profiling engine?]
- The nil default defeats correlation. nil → random 16 bytes (line ~352) produces a session whose id the caller can never learn — it lives only in the internal `activeSessionFileName` and is never returned. So a nil id yields an un-addressable session you can recover generically but can't tie back to the app session that produced it.

would a  `ProfilingSessionID` value type (fixed 16-byte payload; `init(bytes:)`, `init(uuid:)`, `hexString`, `withUnsafeBytes`) used for both `start(sessionId:)` and `RecoverableSession.sessionId` make sense? that makes the length invariant unrepresentable-if-violated (removes the runtime guard and the `.storageSetupFailed` length case) and unifies the two sides.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+    store_set_version(store, EMB_PROFILE_FORMAT_INVALID);
+}
+
+void emb_profile_store_flush(emb_profile_store_t *store)
+{
+    if (store == NULL) { return; }
+    // Async flush of dirty pages to the file ahead of a possible background Jetsam kill. The lower data
+    // half covers the file data region (the upper half aliases the same file pages); the footer holds the
+    // control block + metadata. MS_ASYNC: schedule writeback, don't block.
+    msync(store->base, store->data_capacity, MS_ASYNC);
```
fnewberg, line 217: should we add some logs for failures in these calls?

fnewberg (commented):
@kstenerud this PR is a beast 🙂 this is what i have so far. will continue tomorrow

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: It's now defined in `sessionFileExtension` in `ProfilingEngine.swift`

Sources/EmbraceProfilingSampler/include/emb_profile_store.h
kstenerud: Yes, sorry, those files are included now. They describe the on-disk formats.

Sources/EmbraceProfilingSampler/include/emb_profile_store.h
kstenerud: This allows the file classifier code to read the magic/version data without knowing or caring anything about the file's format. It frees up the format to place alignment-sensitive data at the front (such as the circular buffer) without worry and keep the handling code simple.

Sources/EmbraceProfilingSampler/include/emb_profile_store.h
kstenerud: A future version of the on-disk format could have a different layout, and then we'd need to keep an inventory of these sizes in order to find the offsets we need. It reduces the amount of machinery we need to read older formats.

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
kstenerud: Should be present now!

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: Yes, this is already how it operates. You provide a directory when calling `start()` for it to store its data in, and the files are keyed by the unique 16-byte session id. You can list them (without having to start the profiler) by calling `recoverableSessions()`, and then delete them individually using `delete(_ session: RecoverableSession)`, or if you want everything gone you can just delete the whole directory before starting the profiler.

.gitignore
```diff
@@ -86,3 +86,4 @@ Examples/Benchmarks/Benchmarks.xcworkspace/xcuserdata
 # Claude
 .claude/
 claude/
+/embrace/
```
fnewberg, line 91: @kstenerud why was this added?

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: Once the engine is started, it will be using the sessionId for file naming, so it can't be changed.

Sources/EmbraceProfiling/ProfilingRecovery.swift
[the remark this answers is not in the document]
kstenerud: Gate failure now bails to `[]` instead of over-listing.

Sources/EmbraceProfiling/ProfilingRecovery.swift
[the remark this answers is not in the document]
kstenerud: Gate failure is now a hard refuse-to-delete.

Sources/EmbraceProfilingSampler/source/emb_stack_walker.c
```diff
@@ -61,12 +65,20 @@ bool emb_stack_walk(thread_t thread,
     // Frame pointers must be aligned to sizeof(void *) * 2 bytes
     // (16 bytes on the supported 64-bit architectures).
     const uintptr_t fp_align_mask = sizeof(void *) * 2 - 1;
-    while (count < max_frames && fp >= stack_bottom && fp < stack_top
+    while (fp >= stack_bottom && fp < stack_top
            && ((uintptr_t)fp & fp_align_mask) == 0) {
         void **record = (void **)fp;
-        void *next_fp = record[0];
         void *ret_addr = ptrauth_strip(record[1], ptrauth_key_return_address);
```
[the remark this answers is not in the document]
kstenerud: Bound now covers the full 16-byte record, not just `fp`.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
```diff
@@ -0,0 +1,350 @@
...
+    const size_t page = (size_t)vm_page_size;
+    const size_t data_capacity = round_up_to_page(capacity_bytes, page);
+    // Footer = one RW control-block page + one write-once metadata page.
+    const size_t footer_bytes = 2 * page;
+    const size_t guard_bytes = page;
+    const off_t  file_size = (off_t)(data_capacity + footer_bytes);
+    const size_t reserve_size = 2 * data_capacity + guard_bytes + footer_bytes;
+
+    // --- open + size the file ---
+    int fd = open(path, O_RDWR | O_CREAT | O_TRUNC, 0600);
```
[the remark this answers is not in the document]
kstenerud: `create` now matches recovery/peek.

Tests/EmbraceProfilingTests/ProfilingEngineFileBackedTests.swift
```diff
@@ -0,0 +1,160 @@
...
+            let sessions = ProfilingEngine.shared.recoverableSessions(in: dir)
+            XCTAssertEqual(sessions.count, 1)
+            let handle = try! XCTUnwrap(sessions.first)
+            XCTAssertEqual(handle.sessionId, "aabb")
+            XCTAssertEqual(handle.status, .recoverable)
+            XCTAssertGreaterThan(handle.byteSize, 0)
+
+            guard case let .recovered(result) = ProfilingEngine.shared.recover(handle) else {
+                return XCTFail("expected .recovered, got \(ProfilingEngine.shared.recover(handle))")
+            }
```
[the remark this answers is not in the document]
kstenerud: It's now captured once, reused.

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
kstenerud: PROFILING-DISK-FORMAT.md was supposed to be added to the top-level of EmbraceProfiling originally. It's there now, along with PROFILING-THREAD-STATE.md.

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
kstenerud: Doc is now committed (PROFILING-THREAD-STATE.md)

Sources/EmbraceProfilingSampler/include/emb_ring_buffer.h
kstenerud: Document added.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
kstenerud: Document added.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
kstenerud: Document added.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
kstenerud: Document added.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
kstenerud: It's been renamed to `fail`.

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: Added

Sources/EmbraceProfiling/ProfilingEngine.swift
kstenerud: Added important note to `start()`.

Sources/EmbraceProfiling/ProfilingSample.swift
kstenerud: ThreadStateTests.swift now verifies this.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
kstenerud: `file_size` is `data_capacity + footer_bytes`, so with the default `data_capacity` of 1MB, you'll have that + 8kb `footer_bytes`.

Sources/EmbraceProfilingSampler/include/emb_profile_store.h
kstenerud: Yes, it's to allow for an image table to be embedded directly with the profile data if you want such a thing in future (to ensure it's synchronized across launches, crashes, and app upgrades). Setting both to 0 now would translate to "no table included" if the machinery gets added.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
kstenerud: There are 4 protection classes, defined in "sys/content_protection.h" (inaccessible to user code so the convention is to define it outside like we are doing here).

    #define PROTECTION_CLASS_A 1  // Complete Protection
    #define PROTECTION_CLASS_B 2  // Protected Unless Open
    #define PROTECTION_CLASS_C 3  // Protected Until First User Authentication
    #define PROTECTION_CLASS_D 4  // No Protection

We use class B here so that we can continue to write to the file after the phone is locked.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
kstenerud: It's purely defensive. This should never actually fail in practice.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
kstenerud: We protect this region to ensure that the version data is protected against rogue writes, and the structure always gets stored without any tears. `mprotect()` isn't expected to ever fail in practice on a block we ourselves already successfully protected, but the API allows for failure (no matter how impossible), so we handle it.

Worst case scenario is a duplicate of sample data you already ingested.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
kstenerud: I've updated the comment. What I meant is that the samples had already been retrieved by you while the app was live.

Sources/EmbraceProfilingSampler/source/emb_profile_store.c
kstenerud: I've added `EMB_PROFILE_PEEK_INDETERMINATE` for the case where we can't tell if it's ours or not. `EMB_PROFILE_RECOVER_NOT_OURS` means "Definitely not ours" now.

Sources/EmbraceProfilingSampler/include/emb_profile_store.h
kstenerud: I've updated the comments to better explain what it's for. `emb_profile_store_flush()` is a hedge against a small edge case where we've just updated the memmapped region, then the app gets put to sleep (so the kernel hasn't flushed it yet), then the app stays suspended for a time, then the phone's battery dies, or there's a kernel panic or something.

Calling `emb_profile_store_flush` from willResignActive/didEnterBackground/willTerminate covers that potential hole.

The OS is supposed to cleanly shut down all apps when it detects critically low battery, but my philosophy is that this is likely the least tested path in the phone, so expect the worst.

fnewberg (approved):
@kstenerud i think we're down to pretty nitpicky comments from me so i've approved the PR

LMK if you'd be okay with me merging `main` into `karl/profiling-engine-api` and then merging that branch into this one so it has the latest from `main`

Sources/EmbraceProfilingSampler/include/emb_stack_walker.h
kstenerud: It's so that a caller doesn't have to create a temp bool when it doesn't care about that value.

PR #665, opened 2026-06-26
[EMBR-12777] Chore: [EMBR-12777] Remove unused code from the core SDK

fnewberg:
## Summary

Removes unused code from the core SDK, identified with [Periphery](https://github.com/peripheryapp/periphery). Because much of the SDK is gated behind `#if os(...)` /  `canImport(...)`, a declaration is only treated as dead if it's unused on **every** platform — this PR removes the intersection of iOS/tvOS/watchOS/macOS scans.

Removed:
  - Non-public declarations flagged unused on all platforms (whole files `NanosecondClock.swift` and `AppInfoPayload.swift`, plus members across `EmbraceCommonInternal` and  `EmbraceCore`).
  - The two per-service MetricKit `Options` types (`MetricKit{Crash,Hang}CaptureService+Options`), superseded by the shared `MetricKitCaptureServiceOptions` that were unreferenced.
- Three `public` `EmbraceCommonInternal` symbols — `EmbraceReadWriteLock`, `StorageMechanism.fileName`, `SessionState.init?(appState:)`. `EmbraceCommonInternal` isn't exposed as a library product, so apps that depend on the SDK via SPM can't import it — these symbols were `public` only so other Embrace modules can use them internally, not as customer-facing API. Nothing in the repo references them, so they're safe to drop.

### Tooling

Adds the detector so this stays maintainable:

- `bin/find-dead-code` — builds + scans each platform and reports only the intersection.
- `.periphery.yml` — retains public API, Obj-C-reachable code, SwiftUI previews, Codable properties, and vendored files to suppress config-addressable false positives.

### Verification

Builds on iOS, tvOS, watchOS, and macOS; full macOS test suite passes. No test references any removed symbol.

## Checklist


ferdraghi (approved):

Sources/EmbraceCore/Embrace.swift
```diff
@@ -43,11 +43,6 @@ package class Embrace {
         return state == .started && remoteConfigEnabled
     }
 
-    /// Returns the version of the Embrace SDK.
-    package class var sdkVersion: String {
-        return EmbraceMeta.sdkVersion
-    }
-
```
NachoEmbrace, line 50: Don't remove

Sources/EmbraceCore/SwiftUI/Internal/EmbraceTraceViewState.swift
```diff
@@ -68,20 +68,4 @@ final class EmbraceTraceViewState<Value: Equatable> {
...
-    /// - Returns: A multi-line string containing all tracked metrics
-    func debugDescription() -> String {
-        return """
-            EmbraceTraceViewState Debug Info:
-            ├─ Initialize: \(initialize) times, first at \(initializeTime?.description ?? "never")
-            ├─ Body: \(body) times, last at \(bodyTime?.description ?? "never")
-            ├─ Appear: \(appear) times, last at \(appearTime?.description ?? "never")
-            └─ Disappear: \(disappear) times, last at \(disappearTime?.description ?? "never")
-            """
-    }
```
NachoEmbrace, line 86: Don't remove

Sources/EmbraceCommonInternal/Storage/StorageMechanism.swift
```diff
@@ -29,16 +29,6 @@ extension StorageMechanism {
...
-    /// URL pointing to the folder where the storage will be saved
-    public var fileName: String? {
-        switch self {
-        case .onDisk(let name, _, _):
-            return name + ".sqlite"
-
-        default: return nil
-        }
-    }
-
```
NachoEmbrace, line 41: Don't remove

NachoEmbrace (changes requested):

NachoEmbrace (approved):

PR #675, opened 2026-06-29
7.0 release

fnewberg:
## Checklist


Package.swift
allenhumphreys, line 22: Since the v7 code no longer has @objc extensions, I think this can be removed. (And that's a good thing 🎉 because requiring the -Objc linker flag has non-trivial implications to consumers)

NachoEmbrace (approved):

Package.swift
fnewberg: addressed in https://github.com/embrace-io/embrace-apple-sdk/pull/749. thanks for the feedback!

PR #681, opened 2026-06-30
[EMBR-12810] Fix: ThreadSanitizer data races in 7.0 tests

fnewberg:
## Summary

Fixes the ThreadSanitizer data races on `7.0` that were **not** production bugs — unsynchronized test
doubles (and one white-box test read), surfaced by a TSan testing on the `7.0` branch.

## Races fixed

| Race | Fix |
|------|-----|
| `MockOTelSignalsHandler` collections | `NSLock`-guard (background span/log callbacks vs main-thread reads) |
| `CapturingSpanProcessor` collections | guard (child-processor forwarding runs on `EmbraceSpanProcessor`'s bg queue) |
| `CapturingSpanExporter` collections | guard (same bg path; low-frequency, caught on verify run 4) |
| `DefaultLogBatcher.batch` (11 white-box read sites) | guard the property at the source with a lock — not `processorQueue.sync`, which would deadlock the queue against its own access |

`MockSpanProcessorDelegate` is intentionally left unguarded: `EmbraceSpanProcessor` calls the delegate
**synchronously**; only child processors/exporters use the background queue.

## Cleanup

Introduces a `@TestLocked` property wrapper in `TestSupport` (NSLock-backed; uses `_modify` so
in-place mutations like `array.append` stay atomic) and adopts it in the single-field test doubles,
removing the duplicated "lock + backing var + accessor" boilerplate. Mocks with cross-field atomic
critical sections keep manual locks. Named `TestLocked` (not `Locked`) to avoid a clash with
`OpenTelemetrySdk.Locked`.

## Verification

Full `EmbraceIO-Package` scheme under TSan (`halt_on_error=0`), iPhone 17 sim: **8/8 runs clean**,
0 races, both before and after the `@TestLocked` refactor. Only production change is the one
`DefaultLogBatcher.batch` lock.

### DIFF
--- a/Sources/EmbraceCore/Internal/Logs/Batcher/DefaultLogBatcher.swift
+++ b/Sources/EmbraceCore/Internal/Logs/Batcher/DefaultLogBatcher.swift
@@ -22,7 +22,9 @@ protocol LogBatcher: AnyObject {
 
     var logBatchLimits: LogBatchLimits { get }
     var delegate: LogBatcherDelegate? { get set }
-    var batch: LogsBatch? { get }
+
+    /// Test-only: the in-flight batch, read synchronized on the batcher's processing queue.
+    func currentBatch() -> LogsBatch?
 }
 
 class DefaultLogBatcher: LogBatcher {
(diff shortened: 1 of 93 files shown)


Sources/EmbraceCore/Internal/Logs/Batcher/DefaultLogBatcher.swift
```diff
@@ -32,7 +32,25 @@ class DefaultLogBatcher: LogBatcher {
...
+            batchLock.lock()
+            defer { batchLock.unlock() }
+            return _batch
+        }
+        set {
+            batchLock.lock()
+            defer { batchLock.unlock() }
+            _batch = newValue
+        }
+    }
```
NachoEmbrace: I don't like that we're changing the SDK code to fix TS errors on tests.
Is it possible to fix these errors by changing the tests instead?

fnewberg: fair. 

i kept batch private and reverted the lock. i added currentBatch(), which reads on processorQueue, to support test access which does not add a lock on the data path used by prod code
 
i looked at making it purely test-side change, but the tests reach the batcher through the LogBatcher protocol, so a test-only accessor would mean dropping private on both batch and processorQueue, which didn't seem great. open to doing that though if you prefer that

NachoEmbrace (approved):

Sources/EmbraceCore/Internal/Logs/Batcher/DefaultLogBatcher.swift
NachoEmbrace: Current approach is better. I don't care too much if it's not pretty as long as it's code that it's only there for tests and doesn't affect the actual logic.

PR #686, opened 2026-06-30
[EMBR-12821] Feature: Remove throws from public api

NachoEmbrace:


fnewberg:
@NachoEmbrace fix failures introduce in example apps / BrandGame app

Sources/EmbraceCore/Internal/Startup/StartupInstrumentation+Customization.swift
```diff
@@ -47,26 +47,16 @@ extension StartupInstrumentation {
     /// If the root span is not found or an attribute fails to be set, a warning is logged.
```
fnewberg: the old code emitted a warning, but the do/catch with the logging was removed. either restore that or update this comment

Sources/EmbraceSemantics/Span/EmbraceSpan.swift
```diff
@@ -76,7 +76,7 @@ public protocol EmbraceSpan {
 
     /// Sets an attribute to the span
     /// Can fail if the attribute limit is reached.
```
fnewberg, line 78: we don't give any feedback now if there is a failure. should we consider changing to something like this?

`@discardableResult func setAttribute(key:value:) -> Bool`

Sources/EmbraceSemantics/Log/StackTrace/EmbraceStackTrace.swift
```diff
@@ -30,14 +30,17 @@ public struct EmbraceStackTrace {
...
-    /// - Throws: An `EmbraceStackTraceError.invalidFormat` if any of the frames are not in the expected format
-    /// - Throws: An `EmbraceStackTraceError.frameIsTooLong` if any of the frames has more than the `maximumFrameLength` (10.000 characters).
-    ///
     /// - Important: A stacktrace can't have more than `maximumAmountOfFrames` (200); if that happens, we'll trim the exceeding frames.
-    public init(frames: [String]) throws {
-        let trimmedStackTrace = EmbraceStackTrace.trimStackTrace(frames)
-        try EmbraceStackTrace.validateStackFrames(trimmedStackTrace)
-        self.frames = trimmedStackTrace
+    /// - Important: The stacktrace can fail to be constructed if any of the frames have an unexpected format
+    /// - Important: The stacktrace can fail to be consutrcted if any of the frames is longer than `maximumFrameLength` (10.000 characters).
```
fnewberg:
```suggestion
    /// - Important: Returns `nil` if any of the frames is longer than `maximumFrameLength` (10,000 characters).
```

Sources/EmbraceSemantics/Log/StackTrace/EmbraceStackTrace.swift
```diff
@@ -30,14 +30,17 @@ public struct EmbraceStackTrace {
...
     /// - Parameter frames: An array of frames strings, following the format of `Thread.callStackSymbols`.
-    /// - Throws: An `EmbraceStackTraceError.invalidFormat` if any of the frames are not in the expected format
-    /// - Throws: An `EmbraceStackTraceError.frameIsTooLong` if any of the frames has more than the `maximumFrameLength` (10.000 characters).
-    ///
     /// - Important: A stacktrace can't have more than `maximumAmountOfFrames` (200); if that happens, we'll trim the exceeding frames.
-    public init(frames: [String]) throws {
-        let trimmedStackTrace = EmbraceStackTrace.trimStackTrace(frames)
-        try EmbraceStackTrace.validateStackFrames(trimmedStackTrace)
-        self.frames = trimmedStackTrace
+    /// - Important: The stacktrace can fail to be constructed if any of the frames have an unexpected format
```
fnewberg:
```suggestion
    /// - Important: Returns `nil` if any of the frames have an unexpected format
```

fnewberg (changes requested):

Sources/EmbraceSemantics/Span/EmbraceSpan.swift
NachoEmbrace: Having throws in the first place was to give feedback and we decided we don't want it.

fnewberg: yeah, fair point

fnewberg (approved):

PR #698, opened 2026-07-03
[EMBR-12850] Feature: [EMBR-12850] Improving code documentation in 7.0

NachoEmbrace:
Lots of code docs changes

Note: Some things that didn't need to be public had their access level changed on this PR as well.


Sources/EmbraceCore/Capture/Network/URLSessionCaptureService+Options.swift
```diff
@@ -23,6 +23,8 @@ extension URLSessionCaptureService {
         /// An empty list means no domains should be captured.
         public let onlyAllowDomains: [String]?
 
+        /// Creates a new `Traceparent` configuration.
+        /// - Parameter onlyAllowDomains: First-party domain allowlist. When not `nil`, only requests to these domains get a `traceparent` header.
```
fnewberg:
```suggestion
        /// - Parameter onlyAllowDomains: First-party domain allowlist. When not `nil`, only requests to these domains get a `traceparent` header. Subdomains will also be matched
```

fnewberg (approved):
i added one minor suggestion

Sources/EmbraceCore/Capture/Network/URLSessionCaptureService+Options.swift
NachoEmbrace:
```suggestion
        /// - Parameter onlyAllowDomains: First-party domain allowlist. When not `nil`, only requests to these domains get a `traceparent` header. Subdomains will also be matched.
```

PR #717, opened 2026-07-17
Guard ProfilingEngine test-only extension behind #if !os(watchOS)

fnewberg:
## What

`Tests/EmbraceProfilingTests/ProfilingEngine+Testing.swift` guarded its imports (`EmbraceProfiling`, `EmbraceProfilingSampler`, `EmbraceProfilingTestSupport`) behind `#if !os(watchOS)`, but left `extension ProfilingEngine { … }` outside the guard. On watchOS the module isn't imported, so the type is out of scope and `EmbraceProfilingTests` failed to compile:

```
Tests/EmbraceProfilingTests/ProfilingEngine+Testing.swift:11:11: error: cannot find type 'ProfilingEngine' in scope
```

This turned the **watchOS test** job red (e.g. PR #652's CI, which is stacked on this branch).

## Fix

Wrap the whole file — imports **and** the extension — in a single `#if !os(watchOS)`, matching every sibling test file in the target (`ProfilingEngineTests.swift`, `ProfilingEnginePauseTests.swift`, `FallbackBenchmarks.swift`). The per-method `#if os(watchOS) return false` stubs became dead once the extension can't exist on watchOS, so they're removed.

## Verification

`xcodebuild build-for-testing -scheme EmbraceIO-Package -only-testing:EmbraceProfilingTests` against a watchOS Simulator → **TEST BUILD SUCCEEDED**. Non-watchOS behavior is unchanged (the extension body is identical to before).

### DIFF
--- a/Tests/EmbraceProfilingTests/ProfilingEngine+Testing.swift
+++ b/Tests/EmbraceProfilingTests/ProfilingEngine+Testing.swift
@@ -3,23 +3,20 @@
 //
 
 #if !os(watchOS)
+
     @testable import EmbraceProfiling
     import EmbraceProfilingSampler
     import EmbraceProfilingTestSupport
-#endif
 
-extension ProfilingEngine {
+    extension ProfilingEngine {
 
-    /// TEST-ONLY: Allocate the ring buffer without starting the sampler.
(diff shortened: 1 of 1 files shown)


kstenerud (approved):
LGTM! I'd actually partially implemented this on my branch, but the conflict will be easy to sort out.

PR #720, opened 2026-07-19
[EMBR-13038] Fix: Make hang FrameRateMonitor injectable to fix span over-count test flake

fnewberg:
## Problem

`HangCaptureServiceTests/test_hangStarted_createsSpan` intermittently fails in CI with `XCTAssertEqual ("2") != ("1")` (`sanitizer none`). Sightings: 2026-06-29 (run 28396508980), and 2026-07-17 on PR #652 (run 29574697849).

## Root cause

`makeInstalledService()` → `install` → `onInstall` builds a real, **live** `FrameRateMonitor` — a `CADisplayLink` on the main runloop. CI has no debugger, so the `isDebuggerAttached()` early-return in `onInstall` never fires, which is why it only reproduces in CI and never locally under Xcode.

The test drives `hangStarted` once, then `wait(...)` spins the runloop. On a loaded runner the main thread genuinely stalls >249 ms, the display link ticks and calls `hangStarted` again → a real **second** `emb-thread-blockage` span (default `hangPerSession` is 20, so it's allowed). A true over-count from a live detector, not a poll-predicate race.

## Fix

Make the monitor injectable via a new designated `init(limits:makeMonitor:)`. Production keeps the real `FrameRateMonitor` (behavior unchanged); span-lifecycle tests inject a no-op factory so no live `CADisplayLink` runs. The detection path stays covered by `FrameRateMonitorTests`.

Also adds two deterministic wiring tests:
- `test_onInstall_wiresMonitorAsObserver_whenHangsEnabled` — monitor built with the configured threshold and the service wired as its observer.
- `test_onInstall_skipsMonitor_whenHangsDisabled` — factory never invoked when `hangPerSession == 0`.

## Testing

All 16 `HangCaptureServiceTests` + `FrameRateMonitorTests` pass on iPhone 17 Pro / Xcode 26.5.

### DIFF
--- a/Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
+++ b/Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
@@ -24,11 +24,22 @@ import OpenTelemetryApi
     @objc(EMBHangCaptureService)
     public final class HangCaptureService: CaptureService {
 
-        public init(
+        public convenience init(
             limits: HangLimits = HangLimits()
+        ) {
+            self.init(limits: limits, makeMonitor: { FrameRateMonitor(threshold: $0) })
+        }
+
+        /// Designated initializer. `makeMonitor` builds the frame-rate monitor for a
+        /// threshold and is substitutable, so tests can pass a factory that returns
+        /// `nil` to run without a live `CADisplayLink`.
+        init(
+            limits: HangLimits,
+            makeMonitor: @escaping (_ threshold: TimeInterval) -> FrameRateMonitor?
         ) {
             dispatchPrecondition(condition: .onQueue(.main))
             self.mainThread = pthread_self()
+            self.makeMonitor = makeMonitor
             self.limitData = EmbraceMutex(MutableLimitData(limits: limits))
             super.init()
         }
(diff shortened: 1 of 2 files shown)


Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
```diff
@@ -24,11 +24,22 @@ import OpenTelemetryApi
     @objc(EMBHangCaptureService)
     public final class HangCaptureService: CaptureService {
 
-        public init(
+        public convenience init(
             limits: HangLimits = HangLimits()
+        ) {
+            self.init(limits: limits, makeMonitor: { FrameRateMonitor(threshold: $0) })
+        }
```
NachoEmbrace, line 31: I would prefer if we could just add a `frameRateMonitor` parameter in the constructor.
This way we could just pass the normal `FrameRateMonitor` on production code, and also build a fake/dummy one for testing.

The current code is essentially doing this, but it's done in a way that is not consistent with the rest of the SDK (I'm mostly referring to having the block member called `makeMonitor` instead of a parameter in the consturctor).

NachoEmbrace (changes requested):

fnewberg:
fixed in a different PR

PR #724, opened 2026-07-23
[EMBR-13022] Feature: New Stacktrace Sampler for Hang Capture Service

ferdraghi:
Implements a new main thread watchdog and stacktrace sampler for the main thread so that when a Hang is confirmed by the Framerate Monitor, we can have a relevant Stacktrace available to include in the span. The current implementation picks up a stacktrace after the hang ended, which is completely useless.

### DIFF
--- a/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
+++ b/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
@@ -98,7 +98,9 @@ extension RemoteConfig: EmbraceConfigurable {
         HangLimits(
             hangThreshold: payload.hangLimitsHangThreshold,
             hangPerSession: payload.hangLimitsHangPerSession,
-            reportsWatchdogEvents: payload.hangLimitsReportsWatchdogEvents
+            reportsWatchdogEvents: payload.hangLimitsReportsWatchdogEvents,
+            sampleTriggerThreshold: payload.hangLimitsSampleTriggerThreshold,
+            samplePollInterval: payload.hangLimitsSamplePollInterval
         )
     }
 
--- a/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
+++ b/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
@@ -70,9 +70,9 @@ public class RemoteConfig {
 
         do {
             let data = try Data(contentsOf: url)
-            try _payload.withLock {
-                $0 = try JSONDecoder().decode(RemoteConfigPayload.self, from: data)
-            }
+            let decoded = try JSONDecoder().decode(RemoteConfigPayload.self, from: data)
+            _payload.withLock { $0 = decoded }
+            logHangLimitCorrections(for: decoded)
         } catch {
             logger.error("Error loading cached remote config!")
         }
@@ -187,6 +187,9 @@ extension RemoteConfig: EmbraceConfigurable {
                 $0 = newPayload
                 return changed
             }
+            if didUpdate {
+                strongSelf.logHangLimitCorrections(for: newPayload)
+            }
             strongSelf.saveToCache(data)
 
             completion(didUpdate, nil)
@@ -195,6 +198,48 @@ extension RemoteConfig: EmbraceConfigurable {
 }
 
 extension RemoteConfig {
+
+    /// Descriptions of any hang-limit payload values that `HangLimits` had to correct — an
+    /// out-of-range clamp, a non-finite/non-positive fallback, or the
+    /// `sampleTriggerThreshold >= hangThreshold` re-derivation. Empty when the payload is already
+    /// valid. `HangLimits` corrects these silently (by design — it can't reach a logger and must not
+    /// trap), so this surfaces them one layer up where a bad remote-config push can be diagnosed.
+    /// Pure, so it can be unit-tested without the fetch/logging plumbing.
+    static func hangLimitCorrections(for payload: RemoteConfigPayload) -> [String] {
+        let limits = HangLimits(
+            hangThreshold: payload.hangLimitsHangThreshold,
+            hangPerSession: payload.hangLimitsHangPerSession,
+            reportsWatchdogEvents: payload.hangLimitsReportsWatchdogEvents,
+            sampleTriggerThreshold: payload.hangLimitsSampleTriggerThreshold,
+            samplePollInterval: payload.hangLimitsSamplePollInterval
+        )
+
+        var corrections: [String] = []
+        if limits.hangThreshold != payload.hangLimitsHangThreshold {
+            corrections.append(
+                "hang_threshold \(payload.hangLimitsHangThreshold) → \(limits.hangThreshold) (must be finite and > 0)")
+        }
+        if limits.sampleTriggerThreshold != payload.hangLimitsSampleTriggerThreshold {
+            corrections.append(
+                "sample_trigger_threshold \(payload.hangLimitsSampleTriggerThreshold) → \(limits.sampleTriggerThreshold) "
+                    + "(clamped into [\(HangLimits.minSampleTriggerThreshold), \(HangLimits.maxSampleTriggerThreshold)] "
+                    + "and kept below hang_threshold)")
+        }
+        if limits.samplePollInterval != payload.hangLimitsSamplePollInterval {
+            corrections.append(
+                "sample_poll_interval \(payload.hangLimitsSamplePollInterval) → \(limits.samplePollInterval) "
+                    + "(clamped into [\(HangLimits.minSamplePollInterval), \(HangLimits.maxSamplePollInterval)])")
+        }
+        return corrections
+    }
+
+    /// Logs a warning for each hang-limit value the incoming payload had to have corrected.
+    func logHangLimitCorrections(for payload: RemoteConfigPayload) {
+        for correction in Self.hangLimitCorrections(for: payload) {
+            logger.warning("[Hang] remote config value out of range, corrected: \(correction)")
+        }
+    }
+
     func isEnabled(threshold: Float) -> Bool {
         return Self.isEnabled(hexValue: deviceIdHexValue, digits: Self.deviceIdUsedDigits, threshold: threshold)
     }
--- a/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
+++ b/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
@@ -199,12 +199,14 @@ extension RemoteConfig: EmbraceConfigurable {
 
 extension RemoteConfig {
 
-    /// Descriptions of any hang-limit payload values that `HangLimits` had to correct — an
-    /// out-of-range clamp, a non-finite/non-positive fallback, or the
-    /// `sampleTriggerThreshold >= hangThreshold` re-derivation. Empty when the payload is already
-    /// valid. `HangLimits` corrects these silently (by design — it can't reach a logger and must not
-    /// trap), so this surfaces them one layer up where a bad remote-config push can be diagnosed.
-    /// Pure, so it can be unit-tested without the fetch/logging plumbing.
+    /// Descriptions of any hang-limit payload values that were *invalid* and had to be corrected — a
+    /// non-finite/non-positive fallback, an out-of-range clamp, or a `sampleTriggerThreshold` at or
+    /// above `hangThreshold`. Empty when the payload is valid. The routine cap that keeps the trigger
+    /// a margin below `hangThreshold` is expected policy, not a misconfiguration, so a valid in-range
+    /// request the cap merely trims is deliberately *not* reported. `HangLimits` applies all of this
+    /// silently (by design — it can't reach a logger and must not trap), so this surfaces the genuine
+    /// misconfigurations one layer up where a bad remote-config push can be diagnosed. Pure, so it can
+    /// be unit-tested without the fetch/logging plumbing.
     static func hangLimitCorrections(for payload: RemoteConfigPayload) -> [String] {
         let limits = HangLimits(
             hangThreshold: payload.hangLimitsHangThreshold,
@@ -232,7 +234,7 @@ extension RemoteConfig {
         } else if rawTrigger < HangLimits.minSampleTriggerThreshold {
             triggerReason = "below the floor of \(HangLimits.minSampleTriggerThreshold); clamped up"
         } else if rawTrigger > HangLimits.maxSampleTriggerThreshold {
-            triggerReason = "above the max of \(HangLimits.maxSampleTriggerThreshold); clamped down"
+            triggerReason = "above the max of \(HangLimits.maxSampleTriggerThreshold) (a seconds/ms mixup?); capped"
         } else if rawTrigger >= limits.hangThreshold {
             triggerReason = "at or above hang_threshold (\(limits.hangThreshold)); capped"
         } else {
--- a/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
+++ b/Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
@@ -215,21 +215,39 @@ extension RemoteConfig {
         )
 
         var corrections: [String] = []
+
         if limits.hangThreshold != payload.hangLimitsHangThreshold {
             corrections.append(
-                "hang_threshold \(payload.hangLimitsHangThreshold) → \(limits.hangThreshold) (must be finite and > 0)")
+                "hang_threshold \(payload.hangLimitsHangThreshold) → \(limits.hangThreshold) "
+                    + "(not finite or <= 0; reset to default)")
         }
+
         if limits.sampleTriggerThreshold != payload.hangLimitsSampleTriggerThreshold {
-            corrections.append(
-                "sample_trigger_threshold \(payload.hangLimitsSampleTriggerThreshold) → \(limits.sampleTriggerThreshold) "
-                    + "(clamped into [\(HangLimits.minSampleTriggerThreshold), \(HangLimits.maxSampleTriggerThreshold)] "
-                    + "and kept below hang_threshold)")
+            let raw = payload.hangLimitsSampleTriggerThreshold
+            let reason: String
+            if !raw.isFinite {
+                reason = "not finite; reset to default"
+            } else {
+                // Mirror HangLimits.init: clamp first, then re-derive if still >= hang_threshold.
+                let clamped = Swift.min(
+                    Swift.max(raw, HangLimits.minSampleTriggerThreshold), HangLimits.maxSampleTriggerThreshold)
+                reason =
+                    clamped >= limits.hangThreshold
+                    ? "must stay below hang_threshold (\(limits.hangThreshold)); re-derived"
+                    : "out of [\(HangLimits.minSampleTriggerThreshold), \(HangLimits.maxSampleTriggerThreshold)]; clamped"
+            }
+            corrections.append("sample_trigger_threshold \(raw) → \(limits.sampleTriggerThreshold) (\(reason))")
         }
+
         if limits.samplePollInterval != payload.hangLimitsSamplePollInterval {
-            corrections.append(
-                "sample_poll_interval \(payload.hangLimitsSamplePollInterval) → \(limits.samplePollInterval) "
-                    + "(clamped into [\(HangLimits.minSamplePollInterval), \(HangLimits.maxSamplePollInterval)])")
+            let raw = payload.hangLimitsSamplePollInterval
+            let reason =
+                raw.isFinite
+                ? "out of [\(HangLimits.minSamplePollInterval), \(HangLimits.maxSamplePollInterval)]; clamped"
+                : "not finite; reset to default"
+            corrections.append("sample_poll_interval \(raw) → \(limits.samplePollInterval) (\(reason))")
         }
+
         return corrections
     }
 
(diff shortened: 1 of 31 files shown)


Sources/EmbraceConfiguration/HangLimits.swift
```diff
@@ -16,14 +16,27 @@ import Foundation
...
+    /// How often (in seconds) the background sampler checks main-thread liveness. The SDK clamps
+    /// this up to a compiled-in minimum before use.
+    public let samplePollInterval: TimeInterval
+
     public init(
         hangThreshold: TimeInterval = 0.249,
         hangPerSession: UInt = 20,
-        reportsWatchdogEvents: Bool = false
+        reportsWatchdogEvents: Bool = false,
+        sampleTriggerThreshold: TimeInterval = 0.15,
```
fnewberg: we defined these values in `Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig/RemoteConfigPayload.swift`. we should reference those, or if that is impractical we should put them in a common place so we don't redefine them in multiple places

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
```diff
@@ -207,11 +207,41 @@ extension EmbraceBacktrace {
 
 extension EmbraceBacktrace {
 
+    /// Number of Embrace capture-plumbing frames sitting on top of a stack that was walked on the
```
fnewberg: this comment seems like it could be a bit more terse

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
```diff
@@ -207,11 +207,41 @@ extension EmbraceBacktrace {
...
+    ///   run that suite in Release). If an internal refactor changes the chain, that test fails and
+    ///   points here — update the constant and keep the `@inline(never)` wrappers intact.
+    static let selfCaptureFrameSkip = 5
+
+    /// Wrapper depth for the Apple/`Thread.callStackReturnAddresses` self-capture path
+    /// (`_takeSnapshotApple`). Smaller than ``selfCaptureFrameSkip`` because that path has no
+    /// `Backtracer` indirection on top. Pinned by the same test; see it for the drift caveat.
+    static let appleSelfCaptureFrameSkip = 3
+
+    // `@inline(never)` here (and on `backtrace(of:threadIndex:)`) is load-bearing: it pins the
```
fnewberg: `load-bearing` is my least favorite term that claude uses

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
```diff
@@ -222,31 +252,49 @@ extension EmbraceBacktrace {
         let machThread = pthread_mach_thread_np(thread)
         let canSuspend = pthread_self() != thread
 
-        // suspend thread if not the current thread.
+        // Drop the SDK's own capture frames (present only on self-capture; see `selfCaptureFrameSkip`).
```
fnewberg, line 249: is self-capture even a usecase that we care about?

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
```diff
@@ -222,31 +252,49 @@ extension EmbraceBacktrace {
...
+        let sdkFrameSkip = canSuspend ? 0 : Self.selfCaptureFrameSkip
+        let entries = 512
```
fnewberg: i'm not a fan of magic numbers. let's define this as a constant, explaining why the value was chosen

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
```diff
@@ -222,31 +252,49 @@ extension EmbraceBacktrace {
...
+
+        let addresses: [UInt]
         if canSuspend {
+            // Suspending another thread to walk it is a #423-class deadlock hazard: if that thread
```
fnewberg: feels like this could be terser, and i'm not a fan of documenting history in comments so i would remove the PR reference

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
```diff
@@ -222,31 +252,49 @@ extension EmbraceBacktrace {
...
-            .dropFirst(5)
-            .prefix(entries)
-            .compactMap { $0 as UInt }
+            addresses =
+                Array(UnsafeBufferPointer(start: buffer, count: max(0, count)))
+                .dropFirst(sdkFrameSkip)
+                .prefix(entries)
+                .compactMap { $0 as UInt }
+        } else {
+            // Self-capture: nothing is suspended, so the allocating array API is safe (and required
```
fnewberg, line 283: again, is this a case we really need to support?

Sources/EmbraceCore/Capture/Hang/Watchdog/MainThreadStackSampler.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  Copyright © 2025 Embrace Mobile, Inc. All rights reserved.
```
fnewberg:
```suggestion
//  Copyright © 2026 Embrace Mobile, Inc. All rights reserved.
```

Sources/EmbraceCore/Capture/Hang/Watchdog/StallTriggeredSampler.swift
```diff
@@ -0,0 +1,232 @@
...
+    ///   `busySince` timestamp doubles as the epoch id: while main is blocked the run loop can't fire
+    ///   the observer again, so the value is stable and we sample the episode exactly once.
+    ///
+    /// The trigger sits below the reported-hang threshold so the snapshot lands *inside* the hang
+    /// window that CADisplayLink later confirms.
+    final class StallTriggeredSampler: MainThreadStackSampler {
+
+        /// Lower bound on the trigger. Prevents a bad/hostile remote value from driving overly
+        /// aggressive main-thread suspension.
+        static let minTriggerThreshold: TimeInterval = 0.05  // 50 ms
```
fnewberg: should we define these constants in a common location along with other constants used in hang subsystem?

Sources/EmbraceCore/Capture/Hang/Watchdog/StallTriggeredSampler.swift
```diff
@@ -0,0 +1,232 @@
...
+        /// them, so this lock is only ever contended by concurrent `start`/`stop`/`deinit` callers.
+        private struct LifecycleState {
+            var observer: CFRunLoopObserver?
+            var thread: Thread?
+            var tokens: [NSObjectProtocol] = []
+        }
+        private let lifecycle = EmbraceMutex(LifecycleState())
+
+        /// - Parameters:
+        ///   - mainThread: the `pthread_t` of the thread to sample (the main thread).
```
fnewberg: why pass this in when we know it's the main thread? can the init not figure out on its own what the reference to the main thread is?

Sources/EmbraceCore/Capture/Hang/Watchdog/StallTriggeredSampler.swift
```diff
@@ -0,0 +1,232 @@
...
+        ///   - triggerThreshold: how long main must be continuously busy before we snapshot it.
+        ///     Clamped up to ``minTriggerThreshold``.
+        ///   - pollInterval: how often the background thread checks liveness. Clamped up to
+        ///     ``minPollInterval``.
+        ///   - bufferCap: max buffered samples (small ring; one per stall episode).
+        init(
+            mainThread: pthread_t,
+            triggerThreshold: TimeInterval,
+            pollInterval: TimeInterval = 0.05,
```
fnewberg: same comment as above. don't duplicate constants

Sources/EmbraceCore/Capture/Hang/Watchdog/StallTriggeredSampler.swift
```diff
@@ -0,0 +1,232 @@
...
+            self.bufferCap = bufferCap
+            self.logger = logger
+        }
+
+        /// Convenience: derive the trigger from the reported-hang threshold (a fraction below it so
+        /// the snapshot lands in-window even for short hangs).
+        convenience init(mainThread: pthread_t, hangThreshold: TimeInterval, logger: InternalLogger?) {
+            self.init(
+                mainThread: mainThread,
+                triggerThreshold: hangThreshold * 0.6,
```
fnewberg: define this constant in a central location

Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
```diff
@@ -46,13 +46,30 @@ import OpenTelemetryApi
...
+            let (monitor, sampler) = makeMonitorAndSampler(for: currentLimits)
+            sampler?.start()
+            limitData.withLock {
+                $0.watchdog = monitor
+                $0.sampler = sampler
+            }
+        }
+
+        /// Builds the detector + during-block sampler pair for `limits`, or `(nil, nil)` when hangs
+        /// are disabled. The sampler is returned un-started; callers own `start()`/`stop()`.
```
fnewberg: > The sampler is returned un-started;

what about the frame rate monitor?

Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
```diff
@@ -181,7 +197,27 @@ import OpenTelemetryApi
...
+            // Reconcile the CADisplayLink-confirmed hang with the sampler's during-block snapshots.
+            // `duration` is a clock-agnostic interval, so subtracting it from a fresh
+            // CLOCK_MONOTONIC_RAW reading (taken here on main, ≈ real hang end) yields the window in
+            // the sampler's own clock. The during-block sample was taken at ≈ start + trigger, so it
+            // lands inside; a small tolerance absorbs callback latency.
+            let tolerance: UInt64 = 20_000_000  // 20 ms
+            let endMono = clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW)
+            let durationNanos = UInt64((duration * 1_000_000_000).rounded())
+            let startMono = endMono > durationNanos + tolerance ? endMono &- durationNanos &- tolerance : 0
+            let sample = limitData.withLock { $0.sampler }?.samples(in: startMono...endMono).last
```
fnewberg: should we take the first one instead of the last one to get closer to the start of the blockage?

Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
```diff
@@ -181,7 +197,27 @@ import OpenTelemetryApi
...
+
             spanQueue.async { [self] in
+                if let sample {
+                    addSamplingSpanEvent(
+                        time: sampleTime,
+                        backtrace: sample.backtrace,
+                        overhead: Int(sample.overhead)
+                    )
+                }
+                // else: emit the span with no stack — an honest "no trace" beats the old misleading one.
```
fnewberg:
```suggestion
                // else: emit the span with no stack
```

fnewberg (commented):
first batch of feedback. more to come

Sources/EmbraceCore/Capture/Hang/Watchdog/StallTriggeredSampler.swift
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        // MARK: - Background sampling loop
+
+        private func run() {
+            // The `busySince` value we last captured for. Poller-local: no synchronization needed.
+            var lastSampledEpoch: UInt64 = 0
+
+            while running.load(order: .acquire) {
+                usleep(pollMicros)
```
fnewberg: according to the [man page](https://man7.org/linux/man-pages/man3/usleep.3.html), `usleep(3)` is marked obsolescent in favor of `nanosleep(2)`.

Sources/EmbraceCore/Capture/Hang/Watchdog/StallTriggeredSampler.swift
```diff
@@ -0,0 +1,232 @@
...
+        init(
+            mainThread: pthread_t,
+            triggerThreshold: TimeInterval,
+            pollInterval: TimeInterval = 0.05,
+            bufferCap: Int = 8,
+            logger: InternalLogger?
+        ) {
+            self.mainThread = mainThread
+            self.triggerNanos = UInt64(max(Self.minTriggerThreshold, triggerThreshold) * 1_000_000_000)
+            self.pollMicros = useconds_t(max(Self.minPollInterval, pollInterval) * 1_000_000)
```
fnewberg: there are no upper clamps on remote-config-driven sampler settings. both conversions only clamp the low side. there's no upper bound anywhere else in the path either — `RemoteConfigPayload` → `RemoteConfig` → `HangLimits` → here all pass`sampleTriggerThreshold`/`samplePollInterval` through unvalidated.

- useconds_t is UInt32 (max ≈ 4295s), so sample_poll_interval: 5000 (a seconds/ms mixup) traps.
- UInt64(...) traps above ≈1.8e10s, so an oversized/garbage sample_trigger_threshold (e.g. 1e11) traps too.

maybe we clamp both sides in `HangLimits.init` instead of only here? that way every consumer of `HangLimits` inherits the guarantee

Sources/EmbraceConfiguration/HangLimits.swift
```diff
@@ -16,14 +16,27 @@ import Foundation
...
+        samplePollInterval: TimeInterval = 0.05
     ) {
         self.hangThreshold = hangThreshold
```
fnewberg: would probably be good to clamp this as well. same gap as `sampleTriggerThreshold` / `samplePollInterval` that i mentioned in another comment above, it's assigned straight through with no > 0/finite check.

Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
fnewberg, line 25: `HangCaptureService` does not override `onStop()`, so if Embrace.client.stop() is called, span recording stops, but the sampler's poll thread and `CFRunLoopObserver` keep running.

Also, the sampler is started from `onInstall()`, `not onStart()`, so a service that's installed but never activated also starts up a live poll thread from install time. Not sure how much this matters, but we may want to change this if that is inconsistent with other capture services

Sources/EmbraceCore/Capture/Hang/Watchdog/StallTriggeredSampler.swift
```diff
@@ -0,0 +1,232 @@
...
+        /// the snapshot lands in-window even for short hangs).
+        convenience init(mainThread: pthread_t, hangThreshold: TimeInterval, logger: InternalLogger?) {
+            self.init(
+                mainThread: mainThread,
+                triggerThreshold: hangThreshold * 0.6,
+                logger: logger
+            )
+        }
+
+        deinit { stop() }
```
fnewberg, line 107: [from claude review]

`deinit { stop() }` can't actually run once started — `self` is retained for the life of the poll loop

```
deinit { stop() }
...
let thread = Thread { [weak self] in self?.run() }
```


`[weak self]` avoids a permanent retain cycle, but it doesn't do what it looks like it does here. Reading a weak var requires ARC to atomically load-and-retain it into a strong temporary — that's what `self?` actually produces. That temporary is then the receiver of `.run()`, and ARC holds it strong for the entire duration of that call. Since `run()` is `while running.load(...) { ... }`, the temporary stays alive for as long as the loop runs — i.e. for as long as `running` is `true`.

So while the sampler is polling, there's always a strong reference to it in scope, `[weak self]` notwithstanding. `deinit` can't fire until `run()` returns, and `run()` only returns once `running == false`, which only happens via an explicit `stop()` call. That makes `deinit { stop() }` circular: it's meant to be the fallback for "nobody called stop()," but it can only run after something else already called `stop()`. 

Net effect on this class: once `start()` has been called, nothing but an explicit external `stop()` can end the sampler. Combined with the missing `onStop()` override above, there's currently no code path that ever provides that explicit stop() outside of `onConfigUpdated`'s swap.

Suggest restructuring so the thread body doesn't call through `self` for the loop itself — e.g. have `start()` capture the atomics/buffer it needs directly into the closure, so the object's lifetime isn't pinned to the loop's lifetime, and `deinit` becomes a real backstop instead of dead code.

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
fnewberg, line 316: since `symbolicated()` (line 154) returns the frame unchanged with `image == nil` when `Embrace.client?.options.symbolicator` is `nil`, this function will drop the frame. 

rather than allowing symbolicator to be `nil`, can we change it so we do something like this

```
symbolicator: Symbolicator? = DarwinImageSymbolicator()
```

so that it is easier for customers to use this feature without custom config. i think we could implement `DarwinImageSymbolicator` using `dladdr` and follow a pattern that KSCrash uses. 

let's discuss to see how big a lift this is

fnewberg:
@ferdraghi we should update [these comments](https://github.com/embrace-io/embrace-apple-sdk/blob/60a99e12262a9eaf6ba468989d37e3a1821fc4b7/Sources/EmbraceCore/Options/Embrace+Options.swift#L48-L51) to reflect what the code does

 - `backtracer` defaults to `nil` in Embrace.Options` — there's no built-in fallback at this layer, only EmbraceIO's convenience options wire one in. 
 - symbolicator — "without it, only raw addresses are shown" isn't what happens today: frames are dropped entirely, not shown as raw addresses

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
ferdraghi: It is for logs that capture a stacktrace for Errors and such.

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
ferdraghi: answered above.

Sources/EmbraceCore/Capture/Hang/Watchdog/StallTriggeredSampler.swift
ferdraghi: It's not always run on mainThread though, when reconstructing the sampler running after `onConfigUpdated` it runs on a background thread. Although we could probably switch to use `CFRunLoopGetMain()` instead...

Sources/EmbraceCore/Capture/Hang/HangCaptureService.swift
ferdraghi: yeah, not sure why I agreed with grabbing the last one...

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
ferdraghi: resolved by using KSCrashBacktracing() as default for both symbolicator and backtracer. Could consider removing these overrides altogether since we think they're pointless and we got rid of them in 7.0

Sources/EmbraceCore/Capture/Hang/Watchdog/StallTriggeredSampler.swift
ferdraghi: resolved by capturing the buffer on `start()` so now `stop()` should work as intended.

ferdraghi:
> @ferdraghi we should update [these comments](https://github.com/embrace-io/embrace-apple-sdk/blob/60a99e12262a9eaf6ba468989d37e3a1821fc4b7/Sources/EmbraceCore/Options/Embrace+Options.swift#L48-L51) to reflect what the code does
> 
>     * `backtracer` defaults to `nil` in Embrace.Options` — there's no built-in fallback at this layer, only EmbraceIO's convenience options wire one in.
> 
>     * symbolicator — "without it, only raw addresses are shown" isn't what happens today: frames are dropped entirely, not shown as raw addresses

replying here as well just for consistency. I added `KSCrashBacktracer()` to be the default for both backtracer and symbolicator and considering removing both overrides anyway since I think they're kinda pointless.

Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
```diff
@@ -70,9 +70,9 @@ public class RemoteConfig {
 
         do {
             let data = try Data(contentsOf: url)
-            try _payload.withLock {
-                $0 = try JSONDecoder().decode(RemoteConfigPayload.self, from: data)
-            }
+            let decoded = try JSONDecoder().decode(RemoteConfigPayload.self, from: data)
+            _payload.withLock { $0 = decoded }
```
NachoEmbrace, line 74: What's the point of this change?

Sources/EmbraceCore/Backtrace/EmbraceBacktrace+Internal.swift
```diff
@@ -172,7 +172,11 @@ extension EmbraceBacktraceFrame {
             ),
             image: result.imageName != nil
                 ? Image(
-                    uuid: NSUUID(uuidBytes: result.imageUUID).uuidString,
+                    // `result.imageUUID` is already the full Mach-O UUID string (built via
+                    // `NSUUID(uuidBytes:).uuidString` in the symbolicator). Passing that String back
+                    // through `NSUUID(uuidBytes:)` reinterprets its first 16 UTF-8 bytes as a raw
+                    // `uuid_t`, truncating the UUID to its first 16 characters. Use it directly.
```
NachoEmbrace: This comment seems unnecessary

NachoEmbrace (approved):

Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig.swift
ferdraghi: Because the decoded data is also needed for the log below, but the log does not need to be under a lock, only the payload assignment. So it decodes first, then creates a lock only for the payload assignment and then kicks off the log without a lock.

PR #734, opened 2026-08-14
[EMBR-12964] Feature: Implementing Experiments and Feature Flags

NachoEmbrace:


Sources/EmbraceCore/Internal/Experiments/ExperimentsHandler.swift
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        // A bulk call rebuilds the value once and writes it once.
+        if let newValue = newValue {
+            persist(newValue)
+        }
+    }
+
+    private func untrack(ids: [String], kind: ExperimentKind, endedAt: Date?) {
+        let now = Date()
```
fnewberg: nitpick

```suggestion
        let end = endedAt ?? Date()
```

Sources/EmbraceCore/Internal/Experiments/ExperimentsHandler.swift
```diff
@@ -0,0 +1,262 @@
...
+                else {
+                    continue
+                }
+
+                // The first end time wins, mirroring the write-once rule for tracking.
+                guard state.records[position].endedAt == nil else {
+                    continue
+                }
+
+                state.records[position].endedAt = endedAt ?? now
```
fnewberg:
```suggestion
                state.records[position].endedAt = end
```

Sources/EmbraceConfigInternal/EmbraceConfigurable/RemoteConfig/RemoteConfigPayload.swift
```diff
@@ -107,6 +111,10 @@ public struct RemoteConfigPayload: Decodable, Equatable {
             case samplePollInterval = "sample_poll_interval"
         }
 
+        case maxExperimentCount = "max_experiment_count"
+        case maxExperimentIdLength = "max_experiment_id_length"
+        case maxExperimentVariantLength = "max_experiment_variant_length"
```
fnewberg:
```suggestion
        case maxExperimentCount = "experiment_max_count"
        case maxExperimentIdLength = "experiment_id_max_length"
        case maxExperimentVariantLength = "experiment_variant_max_length"
        
```

Tests/EmbraceConfigInternalTests/Fixtures/remote_config.json
```diff
@@ -55,5 +55,8 @@
         "reports_watchdog_events": true,
         "sample_trigger_threshold": 0.2,
         "sample_poll_interval": 0.03
-    }
+    },
+    "max_experiment_count": 1000,
+    "max_experiment_id_length": 256,
+    "max_experiment_variant_length": 64
```
fnewberg:
```suggestion
    "experiment_max_count": 1000,
    "experiment_id_max_length": 256,
    "experiment_variant_max_length": 64
```

Sources/EmbraceIO/EmbraceIO+Experiments.swift
```diff
@@ -0,0 +1,144 @@
...
+    ///
+    /// Entries are applied in order. An invalid entry is dropped on its own without affecting the
+    /// rest of the list: an id that is empty once trimmed is dropped, and so is an entry whose id or
+    /// variant is over the allowed length. Values are never truncated, since a truncated id would
+    /// refer to a different experiment. Duplicate ids within one call resolve to the first one.
+    /// An empty list does nothing.
+    ///
+    /// - Parameter experiments: The experiments to track.
+    public func trackExperiments(_ experiments: [TrackedExperiment]) {
+        Embrace.client?.experiments.trackExperiments(experiments)
```
fnewberg, line 45: if we call this before the SDK is started, experiments will be silently dropped. if called before the SDK is started, we should be buffering the experiments until the SDK is started

Sources/EmbraceIO/EmbraceIO+Experiments.swift
```diff
@@ -0,0 +1,144 @@
...
+    /// The records stay in the reported data with their end time recorded, rather than disappearing,
+    /// so the end is known rather than inferred. Only the first call for a given id takes effect.
+    /// Ids that were never tracked, or that belong to a feature flag rather than an experiment, are
+    /// ignored.
+    ///
+    /// - Parameters:
+    ///   - ids: Identifiers of the experiments to end.
+    ///   - endedAt: When the enrollments ended. Defaults to the moment of the call.
+    public func untrackExperiments(ids: [String], endedAt: Date? = nil) {
+        Embrace.client?.experiments.untrackExperiments(ids: ids, endedAt: endedAt)
```
fnewberg, line 79: similar to above, although unlikely to see extensive use, we should cache untrack calls before the SDK is started

Sources/EmbraceIO/EmbraceIO+Experiments.swift
```diff
@@ -0,0 +1,144 @@
...
+    ///
+    /// Entries are applied in order. An invalid entry is dropped on its own without affecting the
+    /// rest of the list: an id that is empty once trimmed is dropped, and so is an entry whose id or
+    /// variant is over the allowed length. Values are never truncated, since a truncated id would
+    /// refer to a different flag. Duplicate ids within one call resolve to the first one.
+    /// An empty list does nothing.
+    ///
+    /// - Parameter flags: The feature flags to track.
+    public func trackFeatureFlags(_ flags: [TrackedFeatureFlag]) {
+        Embrace.client?.experiments.trackFeatureFlags(flags)
```
fnewberg, line 120: same comment as above

Sources/EmbraceIO/EmbraceIO+Experiments.swift
```diff
@@ -0,0 +1,144 @@
...
+    /// The records stay in the reported data with their end time recorded, rather than disappearing,
+    /// so the end is known rather than inferred. Only the first call for a given id takes effect.
+    /// Ids that were never tracked, or that belong to an experiment rather than a feature flag, are
+    /// ignored.
+    ///
+    /// - Parameters:
+    ///   - ids: Identifiers of the feature flags to end.
+    ///   - endedAt: When the exposures ended. Defaults to the moment of the call.
+    public func untrackFeatureFlags(ids: [String], endedAt: Date? = nil) {
+        Embrace.client?.experiments.untrackFeatureFlags(ids: ids, endedAt: endedAt)
```
fnewberg, line 154: same comment as above

Sources/EmbraceCore/Session/DataRecovery/UnsentDataHandler.swift
```diff
@@ -152,13 +163,17 @@ class UnsentDataHandler {
         }
     }
 
+    /// - Parameter processId: Process the crash happened in. Pass `nil` when it can't be determined,
```
fnewberg: seems weird to add only this parameter for the docstring. either build out the whole docstring or make this a comment above the arg on L176

Sources/EmbraceCore/Session/DataRecovery/UnsentDataHandler.swift
```diff
@@ -152,13 +163,17 @@ class UnsentDataHandler {
...
+    ///                        so the experiments of the process building the log are not wrongly
+    ///                        attributed to this crash.
     static public func sendCrashLog(
         report: EmbraceCrashReport,
         reporter: EmbraceCrashReporter?,
         session: EmbraceSession?,
         storage: EmbraceStorage?,
         upload: EmbraceUpload?,
         otel: EmbraceOpenTelemetry?,
+        processId: EmbraceIdentifier? = ProcessIdentifier.current,
```
fnewberg: is this right? isn't this getting the current process's ID, which is exactly what we don't want?

Sources/EmbraceCore/Session/DataRecovery/UnsentDataHandler.swift
```diff
@@ -482,11 +515,18 @@ extension UnsentDataHandler {
         session: EmbraceSession?,
         storage: EmbraceStorage?,
         upload: EmbraceUpload?,
-        otel: EmbraceOpenTelemetry?
+        otel: EmbraceOpenTelemetry?,
+        processId: EmbraceIdentifier? = ProcessIdentifier.current
```
fnewberg: same comment as above

Sources/EmbraceCore/Internal/Experiments/ExperimentsHandler.swift
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+
+    // MARK: - Persistence
+
+    /// Mirrors the encoded value into storage and announces the change.
+    ///
+    /// The write goes through `addRequiredResources` rather than the metadata handler because the
+    /// value is exempt from the standard attribute length limit, and the metadata handler truncates.
+    /// Called outside the lock so it is never held across a database hop.
```
fnewberg:
```suggestion
    /// Called outside the lock, and it has to stay that way. Beyond not holding the lock across a
    /// database hop, the notification below is delivered synchronously, and its observer takes
    /// `SessionController`'s lock — which `startSession` holds while reading `encodedExperiments`.
    /// Persisting under the lock would invert that order and deadlock.    
```

Sources/EmbraceCore/Internal/Experiments/ExperimentsSerializer.swift
```diff
@@ -0,0 +1,63 @@
...
+            escape(record.variant ?? ""),
+            String(record.startedAt.millisecondsSince1970Truncated)
+        ]
+
+        if let endedAt = record.endedAt {
+            fields.append(String(endedAt.millisecondsSince1970Truncated))
+        }
+
+        return fields.joined(separator: fieldSeparator)
+    }
```
fnewberg: i think the current serialization approach leads to a lot more overhead than we need to incur. `serialize` re-encodes every record on every change, so tracking N records across N separate calls does N(N+1)/2 record-encodes instead of N. at the default cap of 500 that's 125,250 encodes for 500 records.  and that work is done inside the lock

the per-record cost isn't trivial either. `escape` runs three `replacingOccurrences` passes over two fields, so ~10 string allocations per record-encode.

bulk calls are already fine since the loop appends everything and serializes once. but repeated single-entry tracking degrades, which is a pattern that some customers have indicated they will use extensively

since `kind`, `id`, `variant` and `startedAt` are all `let`, a record's encoding can only change when `endedAt` is set. encoding once on creation and again on untrack turns this into a join:

```suggestion
    /// Joins the records into the final attribute value, or `nil` when there is nothing to report.
    ///
    /// Each record carries its own encoded form, so this only concatenates — the escaping and
    /// timestamp formatting already happened when the record was created or ended.
    static func serialize(_ records: [ExperimentRecord]) -> String? {
        guard !records.isEmpty else {
            return nil
        }

        return records.map(\.encoded).joined(separator: recordSeparator)
    }

    /// Encodes one record from its parts, so ``ExperimentRecord`` can call it from `init` before
    /// `self` is fully formed.
    static func encode(
        kind: ExperimentKind,
        id: String,
        variant: String?,
        startedAt: Date,
        endedAt: Date?
    ) -> String {
        var fields = [
            kind.rawValue,
            escape(id),
            escape(variant ?? ""),
            String(startedAt.millisecondsSince1970Truncated)
        ]

        if let endedAt {
            fields.append(String(endedAt.millisecondsSince1970Truncated))
        }

        return fields.joined(separator: fieldSeparator)
    }
```

if we adopt this approach `encode(_ record:)` can be removed

Sources/EmbraceCore/Internal/Experiments/ExperimentRecord.swift
```diff
@@ -0,0 +1,37 @@
...
+    let kind: ExperimentKind
+    let id: String
+    let variant: String?
+    let startedAt: Date
+    var endedAt: Date?
+
+    var key: ExperimentRecordKey {
+        ExperimentRecordKey(kind: kind, id: id)
+    }
+}
```
fnewberg, line 81: this is part 1 of a proposed with part 2 in the `serialize` comment below.

holding the encoded form on the record prevents use from having to continually re-encode values when new experiments are added. also, storing this here rather than in `ExperimentsHandler.State`, makes the cache to start drifting. `private(set) var endedAt` plus a single `end(at:)` mutator means no code path can change a record without refreshing its encoding. 

`endedAt` is currently written in exactly one place (`ExperimentsHandler.swift:219`), so `private(set)` costs one call site.

```suggestion
struct ExperimentRecord {
    let kind: ExperimentKind
    let id: String
    let variant: String?
    let startedAt: Date

    /// Only ``end(at:)`` may write this, so ``encoded`` can never fall out of step with the fields.
    private(set) var endedAt: Date?

    /// This record's contribution to the `emb.experiments` attribute, built once.
    private(set) var encoded: String

    var key: ExperimentRecordKey {
        ExperimentRecordKey(kind: kind, id: id)
    }

    init(kind: ExperimentKind, id: String, variant: String?, startedAt: Date, endedAt: Date? = nil) {
        self.kind = kind
        self.id = id
        self.variant = variant
        self.startedAt = startedAt
        self.endedAt = endedAt
        self.encoded = ExperimentsSerializer.encode(
            kind: kind, id: id, variant: variant, startedAt: startedAt, endedAt: endedAt)
    }

    /// Marks the record ended and refreshes its encoded form. Enforcing the write-once rule for the
    /// end time stays with the caller; this applies whatever it is given.
    mutating func end(at date: Date) {
        encoded = ExperimentsSerializer.encode(
            kind: kind, id: id, variant: variant, startedAt: startedAt, endedAt: date)
        endedAt = date
    }
}
```

notes on the shape:

- **Encode in `init`, not via a separate `serialize()` call.**
  - A post-creation call forces `encoded` into one of two bad shapes:
    - Optional — `serialize` can no longer just `map(\.encoded)`, and every reader has to handle nil.
    - Empty string to start — worse, because a record that was never serialized silently contributes nothing to the attribute rather than failing loudly.
  - Nothing would force the caller to actually make the call, which hands back the exact guarantee the encapsulation is here to provide.
  - There is no work to defer anyway. `track` validates empty id, duplicate key, length, and cap all *before* constructing the record, so every `ExperimentRecord` that exists is one that gets encoded.

- **There is a memory cost.**
  - `encoded` duplicates `id` and `variant`, so worst-case footprint for the feature roughly doubles.
  - i think that's still a fair tradeoff for the performance gains

- **Naming.**
  - `end(at:)` over `updateEndTime`, because the policy is first-end-wins and `update` implies the value is re-settable.
  - `encode` over `serialize` at the record level, keeping the serializer's existing split: `serialize` builds the whole attribute value, `encode` handles one record.

Sources/EmbraceCore/Internal/Experiments/ExperimentsHandler.swift
fnewberg:
```suggestion
                state.records[position].end(at: endedAt ?? now)
```

with the `endedAt == nil` guard just above unchanged, first-end-wins behavior is identical — `end(at:)` only keeps `encoded` in step with the new timestamp.

worth considering as an alternative: `end(at:)` could own the write-once rule itself, returning whether it changed anything, so both invariants live in the type and neither can be violated from outside:

```swift
mutating func end(at date: Date) -> Bool {
    guard endedAt == nil else { return false }
    ...
    return true
}
```

which would collapse lines 214-220 to `if state.records[position].end(at: endedAt ?? now) { changed = true }`. the cost is that the "first end time wins" comment moves out of `untrack`, away from the other untrack rules.

Sources/EmbraceCore/Internal/Experiments/ExperimentsHandler.swift
```diff
@@ -0,0 +1,262 @@
...
+    ///
+    /// The record exists so a later process can read back this process's value; it is deliberately
+    /// kept out of the reported resources by `ResourcePayload` and `ResourceStorageExporter`.
+    private func persist(_ value: String) {
+        storage?.addRequiredResources([SpanSemantics.keyExperiments: value])
```
fnewberg, line 264: thoughts on adding some sort of cool-down here so we don't have this written 200 in a few seconds if lots of back-to-back calls are made? even a short cool-down could probably help if there are lots of back-to-back calls

Sources/EmbraceCore/Payload/Builders/SessionPayloadBuilder.swift
```diff
@@ -16,11 +17,21 @@ class SessionPayloadBuilder {
         // fetch properties
         let properties = storage.fetchCustomProperties(sessionId: session.idRaw, processId: session.processIdRaw)
 
+        // fetch the experiments tracked by the process this session belongs to
+        // this may be an earlier process, so it can't be read from the handler in memory
```
fnewberg:
```suggestion
        // Fetch the experiments tracked by the process this session belongs to.
        // This may be an earlier process, so it can't be read from the handler in memory.
```

Sources/EmbraceCore/Session/DataRecovery/UnsentDataHandler.swift
```diff
@@ -186,7 +202,8 @@ class UnsentDataHandler {
                 body: "",
                 attributes: attributes,
                 storage: storage,
-                sessionId: session?.id
+                sessionId: session?.id,
+                processId: processId ?? ProcessIdentifier.current
```
fnewberg: similar to the L176`comment, but a distinct issue.

at L176 the problem is a wrong *default*. here the problem is that a `nil` we deliberately computed gets thrown away. `sendCrashReports` (118-125) works to resolve the crashed process — session first, then the report — and lands on `nil` when neither knows. we should keep that `nil` rather than override with a default that is wrong.

```suggestion
                processId: processId
```

this needs `LogPayloadBuilder.build`'s `processId` to become optional first — see the comment there.

Sources/EmbraceCore/Session/DataRecovery/UnsentDataHandler.swift
```diff
@@ -228,13 +246,28 @@ class UnsentDataHandler {
             initialAttributes: [:]
         )
 
+        // The crash belongs to a process that has usually already ended, so its experiments come from
```
fnewberg:
```suggestion
        // The crash belongs to a process that has already ended, so its experiments come from
```

Sources/EmbraceCore/Session/DataRecovery/UnsentDataHandler.swift
```diff
@@ -228,13 +246,28 @@ class UnsentDataHandler {
...
+        // storage rather than from the handler in memory. When the process can't be determined the
+        // attribute is left off entirely, which is better than reporting this process's enrolments.
```
fnewberg:
```suggestion
        // attribute is left off entirely.
```

Sources/EmbraceCore/Payload/Builders/LogPayloadBuilder.swift
```diff
@@ -27,13 +27,22 @@ struct LogPayloadBuilder {
...
+    ///                describe that process rather than the one building the payload.
     static func build(
         timestamp: Date,
         severity: LogSeverity,
         body: String,
         attributes: [String: String],
         storage: EmbraceStorage?,
-        sessionId: EmbraceIdentifier?
+        sessionId: EmbraceIdentifier?,
+        processId: EmbraceIdentifier = ProcessIdentifier.current
```
fnewberg: this needs to be nil-able so sendCrashLog can express "couldn't determine it" with `nil`, which is a state `sendCrashReports:118-125` can produce.
```suggestion
        processId: EmbraceIdentifier?
```

Sources/EmbraceCore/Payload/Builders/LogPayloadBuilder.swift
```diff
@@ -49,8 +58,8 @@ struct LogPayloadBuilder {
                 metadata.append(contentsOf: properties)
                 metadata.append(contentsOf: tags)
             } else {
```
fnewberg: needed since we made `processId` optional

```suggestion
            } else if let processId {
```

Sources/EmbraceCore/Capture/CaptureServices.swift
```diff
@@ -60,7 +60,8 @@ final class CaptureServices {
                     session: nil,
                     storage: storage,
                     upload: upload,
-                    otel: Embrace.client
+                    otel: Embrace.client,
+                    processId: ProcessIdentifier.current
```
fnewberg: the crashlytics report should not be associated with the process sending it

```suggestion
                    processId: nil
```

Sources/EmbraceIO/EmbraceIO+Experiments.swift
```diff
@@ -0,0 +1,144 @@
...
+/// data produced by this process be segmented and compared by experiment and variant.
+///
+/// The declared state belongs to the process, not to a single session: it is held for as long as the
+/// app runs and is attached to everything the SDK reports during that time. It is not persisted
+/// across launches, so it has to be declared again on every start.
+///
+/// A record is identified by its kind together with its id, so an experiment and a feature flag may
+/// share an id and stay independent of each other.
+///
+/// - Note: These methods do nothing until the SDK has been set up. Calls made before
```
fnewberg, line 24: the spec calls for buffering of calls made before SDK startup

Sources/EmbraceSemantics/Span/SpanSemantics.swift
```diff
@@ -13,4 +13,7 @@ public struct SpanSemantics {
     public static let keyNSErrorCode = "error.code"
 
     public static let keyAutoTerminationCode = "emb.auto_termination.code"
+
+    /// Experiments and feature flags the user is enrolled in during the current process.
+    public static let keyExperiments = "emb.experiments"
```
fnewberg, line 18: we should have a shared location when spans and logs need the same value to ensure there is no way for them to drift

fnewberg (changes requested):

Sources/EmbraceCore/Internal/Experiments/ExperimentsHandler.swift
```diff
@@ -0,0 +1,262 @@
...
+
+                if let variant = variant, variant.count > limits.maxVariantLength {
+                    droppedTooLong = true
+                    continue
+                }
+
+                guard state.records.count < limits.maxCount else {
+                    droppedForCap = true
+                    continue
+                }
```
fnewberg, line 167: we should assume that the experiment data that can have invalid timestamps. 

thus we need to code defensively around `startedAt` and `endedAt` values which currently reach `millisecondsSince1970Truncated` unvalidated, which is `EMBInt(trunc(...))` — an unchecked conversion that traps (`EmbraceCommonInternal/Extensions/Date.swift:21-23`). `Date(timeIntervalSince1970: .nan)` constructs happily and carries the NaN all the way down, so this crashes the host app:

```
Fatal error: Double value cannot be converted to Int64 because it is either infinite or NaN
exit=133   (SIGTRAP)
```

`EMBInt(exactly:)` gives us the check for free — nil for non-finite and out-of-range, and it accepts everything legitimate including pre-1970 dates and `Date.distantPast`/`distantFuture`:

```suggestion
                // `startedAt` reaches `millisecondsSince1970Truncated`, whose `EMBInt` conversion
                // traps on a non-finite or out-of-range value. Drop the entry rather than crash the
                // host app, and do it before the cap so bad input can't consume a slot.
                if let startedAt = entry.startedAt, !startedAt.isEncodableForExperiments {
                    droppedBadTimestamp = true
                    continue
                }

                guard state.records.count < limits.maxCount else {
                    droppedForCap = true
                    continue
                }
```

three companion edits this needs:

- `var droppedBadTimestamp = false` alongside the other flags at `:114-116`, and a warning next to the existing three at `:181-191`

- a `Date+Experiments.swift` next to `String+Experiments.swift`:

  ```swift
  extension Date {
      /// Whether this date can be encoded as epoch milliseconds without trapping.
      ///
      /// `millisecondsSince1970Truncated` performs an unchecked `EMBInt` conversion, and
      /// `Date(timeIntervalSince1970:)` accepts `.nan` and both infinities without complaint, so a
      /// caller-supplied date can reach it unvalidated.
      var isEncodableForExperiments: Bool {
          EMBInt(exactly: trunc(millisecondsSince1970)) != nil
      }
  }
  ```

- the same check in `untrack`. it takes one `endedAt` for the whole call, so it validates once before the loop rather than per record:

  ```swift
  if let endedAt, !endedAt.isEncodableForExperiments {
      logger?.warning("Experiments untracked with an unrepresentable end date were dropped!")
      return
  }
  let end = endedAt ?? Date()
  ```

Sources/EmbraceIO/EmbraceIO+Experiments.swift
Tischel: This makes no sense with how our SDK works.

Sources/EmbraceIO/EmbraceIO+Experiments.swift
ferdraghi: I'd argue that's a misuse of the SDK since we require that the first thing your app has to do is start the SDK. Same thing could be said about anything, really... Someone trying to log something before starting the SDK would result in the log being dropped.

Sources/EmbraceSemantics/Span/SpanSemantics.swift
NachoEmbrace: I'm fine with this but this should be done in a separated PR as a clean up task.
Lots of repeated things already.

Sources/EmbraceCore/Internal/Logs/EmbraceLogAttributesBuilder.swift
```diff
@@ -133,6 +133,21 @@ class EmbraceLogAttributesBuilder {
...
+    func addExperiments(_ experiments: String?) -> Self {
+        guard let experiments = experiments,
+            attributes[LogSemantics.keyExperiments] == nil
+        else {
+            return self
+        }
+
+        attributes[LogSemantics.keyExperiments] = experiments
+        return self
+    }
```
fnewberg, line 149: for OTel logs, we have the following call tree

```
UnsentDataHandler.createLogCrashAttributes
  → resolves the dead process's value — nil when the process is unknown
  → .addExperiments(nil)   → early return, key not written
  → .build()               → attributes dict with no emb.experiments
  → otel?.log(attributes: dict)
       ↓
LogController.createLog
  → EmbraceLogAttributesBuilder(initialAttributes: dict)
  → .addExperiments(experiments?.encodedExperiments)  ← current process's experiments; key is free, so it writes
```

in `UnsentDataHandler.createLogCrashAttributes`, if we could not find experiment data, `addExperiments` does not write a key if there had been no experiment data found for the crash. but then when we hit `addExperiments` in `LogController.createLog`, we will populate it with the current process's experiment data if the key was not present. 

to cover this scenario in tests, a test would need a real client — `MockEmbraceOpenTelemetry` means `otel?.log` never reaches `LogController`, which is why tests currently pass.

Sources/EmbraceCore/Session/DataRecovery/UnsentDataHandler.swift
```diff
@@ -228,13 +243,28 @@ class UnsentDataHandler {
...
             attributesBuilder
             .addLogType(.crash)
             .addApplicationProperties()
             .addApplicationState()
             .addSessionIdentifier()
             .addCrashReportProperties()
+            .addExperiments(experiments)
             .build()
 
         otel?.log(
```
fnewberg, line 270: i think we have the same issue here as described in the other comment. a fix to `addExperiments` will probably fix things though

Sources/EmbraceCore/Internal/Experiments/ExperimentsHandler.swift
NachoEmbrace: addressed

Sources/EmbraceCore/Internal/Experiments/ExperimentsHandler.swift
NachoEmbrace: addressed

Sources/EmbraceCore/Internal/Logs/EmbraceLogAttributesBuilder.swift
NachoEmbrace: This is a real but pre-existing issue.
It affects more than just experiments.
Will be addressed in another PR.

NachoEmbrace: Also just as a sidenote, the crash log that is uploaded and seen in our dashboard is correct.
The log that the user's custom processors/exporters receive can potentially have wrong data, and this is the bug.

fnewberg (approved):

PR #738, opened 2026-08-19
[EMBR-13709] Feature: Dropping log batches that are missing required metadata

NachoEmbrace:
Also sends a private log with the amount of logs dropped

### DIFF
--- a/Sources/EmbraceCommonInternal/Protocols/EmbracePrivateLogger.swift
+++ b/Sources/EmbraceCommonInternal/Protocols/EmbracePrivateLogger.swift
@@ -1,0 +1,15 @@
+//
+//  Copyright © 2026 Embrace Mobile, Inc. All rights reserved.
+//
+
+/// Allows sending logs meant for Embrace's own diagnostics.
+///
+/// These logs are uploaded right away instead of going through the regular log pipeline,
+/// which means they are never persisted nor forwarded to processors or exporters set by
+/// the user of the SDK.
+package protocol EmbracePrivateLogger: AnyObject {
+
+    /// Sends a log meant for Embrace's own diagnostics.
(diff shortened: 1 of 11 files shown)


Sources/EmbraceCommonInternal/Protocols/EmbracePrivateLogger.swift
```diff
@@ -0,0 +1,15 @@
+//
+//  Copyright © 2026 Embrace Mobile, Inc. All rights reserved.
+//
+
+/// Allows sending logs meant for Embrace's own diagnostics.
+///
+/// These logs are uploaded right away instead of going through the regular log pipeline,
```
fnewberg, line 7: i don't think we need them that urgently. i'd rather they go in batches to avoid excessive calls to deliver them being made if that doesn't make the implementation a lot more complex

fnewberg (approved):

Sources/EmbraceCore/Internal/Logs/LogController.swift
```diff
@@ -221,6 +222,15 @@ extension LogController {
             }
             let resourcePayload = try createResourcePayload(sessionId: resolvedSessionId)
             let metadataPayload = try createMetadataPayload(sessionId: resolvedSessionId)
+
+            // the backend drops payloads that are missing the required metadata,
+            // so we discard these logs instead of uploading them
+            guard resourcePayload.hasRequiredMetadata else {
+                Embrace.logger.warning("Dropped \(logs.count) logs due to missing metadata!")
```
ferdraghi, line 229: non `DEBUG` builds drop all logs below `.error`.

ferdraghi (approved):

PR #745, opened 2026-09-01
[EMBR-13825] Feature: Implementing fragments handling in the WebViewCaptureService

NachoEmbrace:
Note that even though I could've kept the old logic for `stripQueryParams = true `and `fragmentsHandling = .keep`, I've decided to use the new sanitizer to cover all cases because it is more performant than the Apple API we were using.

### DIFF
--- a/Sources/EmbraceCore/Capture/WebView/WebViewCaptureService+Options.swift
+++ b/Sources/EmbraceCore/Capture/WebView/WebViewCaptureService+Options.swift
@@ -12,7 +12,7 @@
         /// the server, so it is commonly used to carry values that are not meant to leave the browser, such as
         /// the access tokens returned by the OAuth 2.0 implicit grant. It is also used to carry information that
         /// is useful when debugging, such as the hash routes of a single page application.
-        public enum FragmentHandling: Int {
+        public enum FragmentHandling {
             /// The fragment is captured exactly as it appears in the URL.
             case keep
 
--- a/Sources/EmbraceCore/Capture/WebView/WebViewCaptureService+Options.swift
+++ b/Sources/EmbraceCore/Capture/WebView/WebViewCaptureService+Options.swift
@@ -6,15 +6,45 @@
     import Foundation
 
     extension WebViewCaptureService {
+        /// Defines how the SDK should treat the fragment component of a URL captured from a web view.
+        ///
+        /// The fragment (everything after the first `#`) is resolved by the user agent alone and never sent to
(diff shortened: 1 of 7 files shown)


Tests/EmbraceCoreTests/Capture/WebView/WebViewURLSanitizerTests.swift
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        // MARK: - Rule 2: hash routes
+
+        func test_redact_keepsHashRoutes() {
+            XCTAssertEqual(redactFragment("/orders/123"), "/orders/123")
+            XCTAssertEqual(redactFragment("!/legacy/orders/123"), "!/legacy/orders/123")
+        }
+
+        func test_redact_recursesIntoTheQueryOfAHashRoute() {
```
fnewberg, line 163: this checks that recursion happens, but not where it stops. since a terrible URL -- unlikely, but possible -- could cause us to hit the recursion limit, we should verify that it is handled gracefully.

can we add a case that nests past the limit and asserts that the limit is respected and that the return value is what we expect it to be?

Sources/EmbraceCore/Capture/WebView/WebViewCaptureService+Options.swift
```diff
@@ -6,15 +6,45 @@
     import Foundation
 
     extension WebViewCaptureService {
+        /// Defines how the SDK should treat the fragment component of a URL captured from a web view.
+        ///
+        /// The fragment (everything after the first `#`) is resolved by the user agent alone and never sent to
+        /// the server, so it is commonly used to carry values that are not meant to leave the browser, such as
+        /// the access tokens returned by the OAuth 2.0 implicit grant. It is also used to carry information that
+        /// is useful when debugging, such as the hash routes of a single page application.
+        public enum FragmentHandling: Int {
```
fnewberg: [from claude w/ some changes by me]

- nothing reads those numbers today, but `: Int` publishes them as part of the public API and reordering the cases silently changes what a stored 1 means. should we just get rid of `: Int`?

- should we add  `Sendable`, here and on `Options`?  Public types don't get it automatically, and customers can't add it themselves from outside the module. options gets read from the swizzled `WKNavigationDelegate` callbacks, so it does move between threads — that's fine in this package today, but it'll show up as an error in customers' Swift 6 builds.

fnewberg (approved):
just some minor comments

PR #754, opened 2026-09-10
[EMBR-13862] Feature: Enabling MetricKit when the CrashlyticsReporter is active

NachoEmbrace:


fnewberg:
@NachoEmbrace any test coverage we should update and/or add?

NachoEmbrace:
> @NachoEmbrace any test coverage we should update and/or add?

Yes, good catch.
Tests added.

fnewberg (approved):

PR #755, opened 2026-09-10
[EMBR-10910] Feature: Screen Tracking: SwiftUI Manual Tracking

ferdraghi:


Sources/EmbraceCore/Capture/State/StateSpanToken.swift
```diff
@@ -89,6 +88,24 @@ final class StateSpanToken {
...
+    /// counter keys are omitted when zero, so a forged `emb.state.not_in_session` would survive on
+    /// any event without counts.
+    private static func bounded(_ callerAttributes: EmbraceAttributes) -> EmbraceAttributes {
+        let allowed = callerAttributes.filter { !SpanSemantics.State.isReserved($0.key) }
+        return sanitizer.sanitizeSpanEventAttributes(allowed)
+    }
+
+    /// Default-constructed, which is the only way the handler ever builds one. If its limits become
+    /// configurable, thread that instance in here rather than letting this one keep its own.
+    private static let sanitizer = DefaultOtelSignalsSanitizer()
```
NachoEmbrace: This seems out of place here and not testable since the sanitizer is hardcoded.
Should probably be injected somehow, either directly in the constructor (like our DefaultOTelSignalsHandler), or through some other pattern (like the delegate in DefaultEmbraceSpan).

ferdraghi: good catch, I'll fix it.

NachoEmbrace (approved):

PR #776, opened 2026-09-16
[EMBR-14055] Fix: Span parenting and trace ids

NachoEmbrace:
Three related fixes to span parenting and trace ids, all cases where the bridged path (`EmbraceIO.Options.otel`
  configured) behaved differently from the non-bridged one. In each case the span Embrace uploaded and the span the
  customer's exporter received disagreed, with nothing logged either way.

  ## 1. Child spans not inheriting the parent's trace (`ab67234b`)

  `EmbraceOTelBridge.startSpan` only set an OTel parent when the parent was still live in `spanCache`. On a miss it
  fell through to the builder's default, so the child got a **new trace id** while Core still recorded
  `parentSpanId` — an orphan in the payload and a root in the customer's exporter.

  Affected any child whose parent had already ended, was created with an `endTime`, or came from outside the bridge.
  Reachable from `createStartupChildSpan`, `CaptureServices+UIViewController.createChildSpan`, and the public
  `createSpan(parentSpan:)`.

  Parenting in OTel is a relationship between identifiers, not object references — an ended span is a perfectly
  valid parent. The bridge now reconstructs a `SpanContext` from the parent's ids on a cache miss, matching both
  `DefaultOTelSignalBridge` and the OTel SDK's own behaviour.

  ## 2. Link identifier validation (`9bbb9854`)

  `addLink(spanId:traceId:)` takes raw strings from customers. Core stored them unexamined while the bridge
  validated and silently discarded the link — so the Embrace payload carried a link nothing could resolve, and the
  exporter carried none.

  Adds a shared validator on `EmbraceSpanContext` in `EmbraceSemantics` (exactly 32/16 ASCII hex characters, not all
  zeros) used by both Core and the bridge. `createLink` now rejects malformed identifiers with a new
  `EmbraceOTelError.invalidSpanLinkIdentifiers` and lowercases valid ones, so the same id written in either case is
  always stored the same way.

  **Also fixes a crash.** `SpanId(fromHexString:)` builds its string indices *before* checking the length, so it
  traps on any string shorter than 16 characters. Both id→OTel conversions in the bridge now validate first. The
  links path had this latent in production: `addLink(spanId: "abc", …)` with the bridge enabled would have taken the
  app down.

  ## 3. Root spans adopting the ambient active span (`6c4952da`)

  `SpanBuilderSdk` defaults to `parentType = .currentSpan`, and the active span lives in a **process-global** store
  shared with every OTel user in the app (on Apple platforms it's `os_activity`-backed, so it propagates across
  dispatched work too). A parentless Embrace span was therefore only a root by accident.

  An app instrumenting its own startup with `withActiveSpan { … }` and calling `EmbraceIO.start()` inside it would
  have its **session span** silently parented to that span, taking `session.traceId` with it — and consecutive
  session parts could end up sharing one trace id. Every Embrace root was exposed: session part, startup root,
  `emb-process-launch`, network, hang, low-power, and view-load roots.

  The bridge now states parentage explicitly in both directions, calling `setNoParent()` when no parent was
  requested.


Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
```diff
@@ -368,9 +368,25 @@ extension DefaultOTelSignalsHandler: EmbraceSpanDataSource {
             throw EmbraceOTelError.spanLinkLimitReached("Links limit reached for span \(spanName)")
         }
 
+        // A link is nothing but a pair of identifiers pointing at another span, so identifiers that
+        // can't refer to one make the link meaningless: nothing downstream is able to resolve them.
+        // Rejecting here keeps the link out of the payload entirely, rather than recording one that
+        // only looks valid until something tries to follow it.
+        guard EmbraceSpanContext.isValidSpanId(spanId) else {
+            throw EmbraceOTelError.invalidSpanLinkIdentifiers(
```
fnewberg, line 385: This error never reaches customers in a release build.

```
throw invalidSpanLinkIdentifiers
 --> caught by DefaultEmbraceSpan.addLink
      --> Embrace.logger.warning(...)
           --> console  DROPPED  EmbraceLogLevel.default == .error outside DEBUG,
                                 BaseInternalLogger:85 gates level.rawValue <= level.rawValue
           --> backend  DROPPED  InternalLogLimits(warning: 0),
                                 send path returns early on guard limit > 0
      --> returns nil, but addLink is @discardableResult --> usually unread
```

Thus a malformed id silently drops the link, and the only way to find out is to pass `logLevel: .warning` to `EmbraceIO.Options`,  which a customer is unlikely to do

All that said, this should probably be an  `.error`. It's a caller mistake, permanently reproducible and fixable by the customer, so it belongs in the band that's on by default (`InternalLogLimits(error: 3)`). `spanLinkLimitReached` is different though and can stay at `.warning`.

Sources/EmbraceCore/Internal/OTel/EmbraceOTelError.swift
```diff
@@ -9,6 +9,7 @@ package enum EmbraceOTelError: Error, Equatable {
     case spanLimitReached
     case spanEventLimitReached(_ description: String)
     case spanLinkLimitReached(_ description: String)
+    case invalidSpanLinkIdentifiers(_ description: String)
```
fnewberg, line 12: This new case adds a third reason `addLink` can return `nil`, and the public doc names only one.

`EmbraceSpan.swift:69-70` and `:124-125` both say the return is `nil` "if the per-span link limit was reached and the link was dropped (a warning is logged in that case)". As of this PR `nil` also means the identifiers were rejected.

Those are `///` docs on customer-facing API, so they ship in generated documentation. Suggest enumerating the three causes and saying which of them log.

(Commenting here because `EmbraceSpan.swift` isn't in this PR's diff.)

fnewberg:
@NachoEmbrace we don't have the trace ID/span ID validation that is added in this PR on the default bridge path. i don't think this is worth addressing, but i'm bringing it in case you feel differently. AFAICT the only way this could become an issue is if a customer implemented the public `EmbraceSpan` protocol themselves, got the ID generation wrong, and passed instances created by it to `createSpan()`. that seems unlikely to happen

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
```diff
@@ -368,9 +368,25 @@ extension DefaultOTelSignalsHandler: EmbraceSpanDataSource {
...
+                "Invalid span id '\(spanId)' for a link on span \(spanName). Expected \(EmbraceSpanContext.spanIdLength) hexadecimal characters."
```
```suggestion
                "Invalid span id '\(spanId)' for a link on span \(spanName). Expected \(EmbraceSpanContext.spanIdLength) hexadecimal characters that are not all zeros."
```

Sources/EmbraceSemantics/Span/EmbraceSpanContext.swift
```diff
@@ -22,3 +22,61 @@ public class EmbraceSpanContext {
...
+        return isValidIdentifier(spanId, length: spanIdLength)
+    }
+
+    /// Returns wether both identifiers in this context are well formed.
+    public var isValid: Bool {
+        return Self.isValidSpanId(spanId) && Self.isValidTraceId(traceId)
+    }
+
+    /// Identifiers are compared and stored in lowercase, so the same value written in either
+    /// case always ends up as the same identifier.
```
fnewberg:
```suggestion
    /// Returns the identifier in its canonical lowercase form. Call this 
    /// before storing or comparing an identifier that came from outside 
    /// the SDK.
```

Sources/EmbraceSemantics/Span/EmbraceSpanContext.swift
```diff
@@ -22,3 +22,61 @@ public class EmbraceSpanContext {
...
+// MARK: Validation
+extension EmbraceSpanContext {
+
+    /// Amount of hexadecimal characters in a well formed trace identifier.
+    public static let traceIdLength = 32
+
+    /// Amount of hexadecimal characters in a well formed span identifier.
+    public static let spanIdLength = 16
+
+    /// Returns wether the given string can be used as a trace identifier.
```
fnewberg:
```suggestion
    /// Returns whether the given string can be used as a trace identifier.
```

Sources/EmbraceSemantics/Span/EmbraceSpanContext.swift
```diff
@@ -22,3 +22,61 @@ public class EmbraceSpanContext {
         self.traceId = traceId
     }
 }
+
+// MARK: Validation
+extension EmbraceSpanContext {
+
+    /// Amount of hexadecimal characters in a well formed trace identifier.
+    public static let traceIdLength = 32
```
fnewberg:
```suggestion
    package static let traceIdLength = 32
```

Sources/EmbraceSemantics/Span/EmbraceSpanContext.swift
```diff
@@ -22,3 +22,61 @@ public class EmbraceSpanContext {
...
+
+    /// Amount of hexadecimal characters in a well formed span identifier.
+    public static let spanIdLength = 16
```
fnewberg:
```suggestion
    package static let spanIdLength = 16
```

Sources/EmbraceSemantics/Span/EmbraceSpanContext.swift
```diff
@@ -22,3 +22,61 @@ public class EmbraceSpanContext {
...
+
+    /// Returns wether both identifiers in this context are well formed.
+    public var isValid: Bool {
+        return Self.isValidSpanId(spanId) && Self.isValidTraceId(traceId)
+    }
+
+    /// Identifiers are compared and stored in lowercase, so the same value written in either
+    /// case always ends up as the same identifier.
+    /// - Parameter identifier: The identifier to normalize.
+    public static func normalize(_ identifier: String) -> String {
```
fnewberg:
```suggestion
    package static func normalize(_ identifier: String) -> String {
```

Sources/EmbraceOTelBridge/EmbraceOTelBridge.swift
```diff
@@ -121,25 +121,44 @@ extension EmbraceOTelBridge: EmbraceOTelSignalBridge {
...
+                // relationship between identifiers: a parent that already ended, or one that was
+                // never created through this bridge, is still a valid parent. Rebuilding the
+                // context from the identifiers keeps the child in the parent's trace.
+                //
+                // The identifiers are validated first: `SpanId(fromHexString:)` traps on strings
+                // shorter than 16 characters, so they can't be handed to it unchecked.
+                if let parentContext = Self.otelContext(from: parentSpan.context) {
+                    builder = builder.setParent(parentContext)
+
+                } else {
```
fnewberg, line 141: Core still records the parent the bridge just refused.

`_createSpan` passes `parentSpanId: parentSpan?.context.spanId` through to `newSpan` regardless, so the uploaded span claims a `parentSpanId` while carrying a freshly minted trace id that contains no such parent. That's the same "two records of the same span disagree, nothing logged" shape this PR sets out to fix, one level down.

NachoEmbrace:
> @NachoEmbrace we don't have the trace ID/span ID validation that is added in this PR on the default bridge path. i don't think this is worth addressing, but i'm bringing it in case you feel differently. AFAICT the only way this could become an issue is if a customer implemented the public `EmbraceSpan` protocol themselves, got the ID generation wrong, and passed instances created by it to `createSpan()`. that seems unlikely to happen

@fnewberg Are you referring to EmbraceIO.createSpan? That method can't receive a constructed span. There's no way for the customer to send us a custom made span by implementing the protocol.
The only way for us to receive a span that was not created through our public API, is directly from the OTel SDK through our bridge. And in this scenario the span context was already validated by the OTel SDK.

Sources/EmbraceSemantics/Span/EmbraceSpanContext.swift
```diff
@@ -22,3 +22,61 @@ public class EmbraceSpanContext {
         self.traceId = traceId
     }
 }
+
+// MARK: Validation
+extension EmbraceSpanContext {
+
+    /// Amount of hexadecimal characters in a well formed trace identifier.
```
fnewberg:
```suggestion
    /// Number of hexadecimal characters in a well-formed trace identifier.
```

Sources/EmbraceSemantics/Span/EmbraceSpanContext.swift
```diff
@@ -22,3 +22,61 @@ public class EmbraceSpanContext {
...
+    public static let traceIdLength = 32
+
+    /// Amount of hexadecimal characters in a well formed span identifier.
```
fnewberg:
```suggestion
    /// Number of hexadecimal characters in a well-formed span identifier.
```

Sources/EmbraceOTelBridge/EmbraceOTelBridge.swift
```diff
@@ -180,6 +199,31 @@ extension EmbraceOTelBridge: EmbraceOTelSignalBridge {
...
+    /// indices before validating the length, so it traps on anything shorter than 16 characters.
+    private static func otelContext(from context: EmbraceSpanContext) -> SpanContext? {
+        guard context.isValid else {
+            return nil
+        }
+
+        let otelTraceId = TraceId(fromHexString: context.traceId)
+        let otelSpanId = SpanId(fromHexString: context.spanId)
+
+        guard otelTraceId.isValid, otelSpanId.isValid else {
```
fnewberg: is this necessary given the `guard` on L208 and that we derive the `otelTraceId` and `otelSpanId` values from `context`?

Sources/EmbraceOTelBridge/EmbraceOTelBridge.swift
```diff
@@ -121,25 +121,44 @@ extension EmbraceOTelBridge: EmbraceOTelSignalBridge {
...
                 builder = builder.setParent(otelParent)
+
+            } else {
+                // The cache only holds spans that are still open, but parenting in OTel is a
+                // relationship between identifiers: a parent that already ended, or one that was
+                // never created through this bridge, is still a valid parent. Rebuilding the
+                // context from the identifiers keeps the child in the parent's trace.
+                //
+                // The identifiers are validated first: `SpanId(fromHexString:)` traps on strings
+                // shorter than 16 characters, so they can't be handed to it unchecked.
```
fnewberg: nitpick: redundant comment. already stated on L205-206

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
```diff
@@ -368,9 +368,25 @@ extension DefaultOTelSignalsHandler: EmbraceSpanDataSource {
             throw EmbraceOTelError.spanLinkLimitReached("Links limit reached for span \(spanName)")
         }
 
+        // A link is nothing but a pair of identifiers pointing at another span, so identifiers that
+        // can't refer to one make the link meaningless: nothing downstream is able to resolve them.
+        // Rejecting here keeps the link out of the payload entirely, rather than recording one that
+        // only looks valid until something tries to follow it.
+        guard EmbraceSpanContext.isValidSpanId(spanId) else {
```
fnewberg, line 384: Consider moving both guards above the limit check at `:366-369`. Knowing that something is invalid is probably more useful than knowing the limit has been hit.

```
now:       limit? --> spanLinkLimitReached      (masks a malformed id at the cap)
              +--> valid? --> invalidSpanLinkIdentifiers

suggested: valid? --> invalidSpanLinkIdentifiers
              +--> limit? --> spanLinkLimitReached
```

Tests/EmbraceOTelBridgeTests/EmbraceOTelBridgeTests.swift
```diff
@@ -299,6 +299,276 @@ final class EmbraceOTelBridgeTests: XCTestCase {
...
+        )
+
+        let childMock = MockEmbraceSpan(spanId: childCtx.spanId, traceId: childCtx.traceId)
+        bridge.endSpan(childMock, endTime: Date())
+
+        wait(timeout: .defaultTimeout) { self.spanProcessor.endedSpans.count == 1 }
+        XCTAssertNil(spanProcessor.endedSpans.first?.parentSpanId)
+    }
+
+    func test_startSpan_withUnusableParentIds_startsNewTrace() {
```
fnewberg, line 550: This test and `test_startSpan_withParentShorterThanASpanId_doesNotTrap` (`:529`) both pass against pre-fix source.

They assert `parentSpanId == nil`, but they run outside any active scope, where that holds whether or not `setNoParent()` is there. 

Two-line fix: wrap the `bridge.startSpan` call in `withActiveSpan(ambientSpan) { }` the way `:404` already does, and add `XCTAssertNotEqual(childCtx.traceId, ambientSpan.context.traceId.hexString)`.

fnewberg (approved):
added comments, but all of them are polish

Sources/EmbraceOTelBridge/EmbraceOTelBridge.swift
NachoEmbrace: good catch

fnewberg (approved):

PR #778, opened 2026-09-16
[EMBR-14059] Fix: Preventing ended spans from being mutated

NachoEmbrace:


Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
```diff
@@ -99,7 +99,28 @@ class DefaultEmbraceSpan: EmbraceSpan {
...
 
+    /// Returns whether this span has already ended, and therefore no longer accepts changes.
+    ///
+    /// A span that ends becomes a completed record: its status, attributes, events and links are
+    /// fixed from that point on. This mirrors the OpenTelemetry behavior, where a span stops
+    /// recording once it ends. Without it, a late change would reach the copy Embrace stores and
+    /// uploads while being silently refused by the OpenTelemetry pipeline, leaving the two
+    /// descriptions of the same span disagreeing.
+    var hasEnded: Bool {
+        state.safeValue.endTime != nil
```
ferdraghi, line 111: if the end condition is based on the span having an end date, maybe we could make the property setter be accessed only through a method like `endSpan(date)` or something so that we prevent accidentally setting an end date before adding some other properties and we make it clear that any modifications that could be added in the future should come *before* the `endSpan` call

fnewberg:
`EmbraceIO.addAttributesToStartupTrace(_:)` could be called late prior to this change.

**The startup root span ends about one frame after the app becomes active.**

The span's own lifetime is the startup, so it has to end early:

```
UIApplicationDidBecomeActive
└─ EMBStartupTracker.onAppDidBecomeActive:                      EMBStartupTracker.m:85
   └─ EMBDisplayLinkProxy.trackNextTick: { … }                  :98-101   ← next display-link tick
      └─ EMBStartupTracker.firstFrameTime = [NSDate date]       :100
         └─ setFirstFrameTime: → self.onFirstFrameTimeSet(…)    :29-35
            └─ StartupInstrumentation.endSpans(_:)              StartupInstrumentation.swift:28-30, 37
               ├─ $0.firstFrameSpan?.end(endTime:)              :39
               └─ $0.rootSpan?.end(endTime:)                    :40
```

So the window in which the API can work closes roughly 16 ms after `didBecomeActive`. Anything
driven by an async completion — a remote config fetch, a feature-flag SDK, a keychain read, a Core
Data load — runs after it. Only a call made synchronously during SDK setup or `didFinishLaunching`
lands inside the window.

@NachoEmbrace would be curious to get your thoughts on this. i'm not sure how we get around this in a way that doesn't feel hack-ish

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
```diff
@@ -312,6 +316,14 @@ extension DefaultOTelSignalsHandler: EmbraceSpanDelegate {
     }
 
     func onSpanEnded(_ span: any EmbraceSpan, endTime: Date) {
+        // A span that ended on its own must not be auto terminated when the session ends: that
+        // would stamp an error code and an error status on a span that already completed, and
+        // would report it as failed. Dropping it here also keeps the cache from holding every
+        // auto terminating span for the whole session.
+        cache.withLock {
+            $0.autoTerminationSpans[span.context.spanId] = nil
```
fnewberg, line 350: This eviction breaks auto-termination code inheritance. `_createSpan` resolves a child's code by looking its parent up in this same dictionary (`:121`), so once a parent ends, a child created afterwards inherits nothing. It's never auto-terminated, endTime stays nil, and it ships as a span snapshot instead of a completed span carrying `emb.error_code`.

The cache is doing two jobs: the worklist of spans to terminate at session end, and the lookup table for inheriting a parent's code. Evicting on end is right for the first and problematic for the second. Reading the code off the parent object instead fixes it and lets the cache stay single-purpose:

Reading the code off the parent object instead fixes it and lets the cache stay single-purpose. This can be fixed by changing line 121:

``` 
    // get auto termination code from parent if needed
    var code = autoTerminationCode
    if let parentSpan, code == nil {
  -     code = cache.safeValue.autoTerminationSpans[parentSpan.context.spanId]?.autoTerminationCode
  +     code = (parentSpan as? DefaultEmbraceSpan)?.autoTerminationCode
    }
```

`newSpan` already stores the resolved code on the span (`:138`), so chains still work, and this also covers parents that were never in the cache.

`test_createSpan_autoTermination_parentCode` creates the child while the parent is still open, so it does not cover this scenario.

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
```diff
@@ -99,7 +99,28 @@ class DefaultEmbraceSpan: EmbraceSpan {
...
+    }
+
+    /// Reports a change that was dropped because the span had already ended.
+    func logIgnoredMutation(_ description: String) {
+        Embrace.logger.warning("Ignoring \(description) on span '\(self.name)': the span already ended.")
```
fnewberg, line 116: should we make this an error so that it is visible in production or would you worry that this could get really chatty?

Sources/EmbraceCore/Public/OTel/EmbraceSpan+EndErrorCode.swift
```diff
@@ -16,6 +16,13 @@ extension EmbraceSpan {
     ///   - errorCode: Error code for the span
     ///   - endTime: Time when the span ended
     public func end(errorCode: EmbraceSpanErrorCode? = nil, endTime: Date = Date()) {
+        // A span that already ended keeps the status and error code it ended with. Returning here
+        // rather than letting each step be refused on its own keeps a single call from reporting
+        // the same thing three times.
+        guard self.endTime == nil else {
+            return
```
fnewberg, line 34: should this emit at least a debug log?

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
NachoEmbrace: endTime doesnt have a public setter

Sources/EmbraceCore/Public/OTel/EmbraceSpan+EndErrorCode.swift
```diff
@@ -16,6 +16,13 @@ extension EmbraceSpan {
     ///   - errorCode: Error code for the span
     ///   - endTime: Time when the span ended
     public func end(errorCode: EmbraceSpanErrorCode? = nil, endTime: Date = Date()) {
+        // A span that already ended keeps the status and error code it ended with. Returning here
+        // rather than letting each step be refused on its own keeps a single call from reporting
+        // the same thing three times.
+        guard self.endTime == nil else {
```
fnewberg, line 33: should this be wrapped in a lock?

Sources/EmbraceCore/Public/OTel/EmbraceSpan+EndErrorCode.swift
```diff
@@ -16,6 +16,13 @@ extension EmbraceSpan {
     ///   - errorCode: Error code for the span
     ///   - endTime: Time when the span ended
     public func end(errorCode: EmbraceSpanErrorCode? = nil, endTime: Date = Date()) {
+        // A span that already ended keeps the status and error code it ended with. Returning here
```
fnewberg: should this, or some of it, be in the docstring above?

NachoEmbrace:
> `EmbraceIO.addAttributesToStartupTrace(_:)` could be called late prior to this change.
> 
> **The startup root span ends about one frame after the app becomes active.**
> 
> The span's own lifetime is the startup, so it has to end early:
> 
> ```
> UIApplicationDidBecomeActive
> └─ EMBStartupTracker.onAppDidBecomeActive:                      EMBStartupTracker.m:85
>    └─ EMBDisplayLinkProxy.trackNextTick: { … }                  :98-101   ← next display-link tick
>       └─ EMBStartupTracker.firstFrameTime = [NSDate date]       :100
>          └─ setFirstFrameTime: → self.onFirstFrameTimeSet(…)    :29-35
>             └─ StartupInstrumentation.endSpans(_:)              StartupInstrumentation.swift:28-30, 37
>                ├─ $0.firstFrameSpan?.end(endTime:)              :39
>                └─ $0.rootSpan?.end(endTime:)                    :40
> ```
> 
> So the window in which the API can work closes roughly 16 ms after `didBecomeActive`. Anything driven by an async completion — a remote config fetch, a feature-flag SDK, a keychain read, a Core Data load — runs after it. Only a call made synchronously during SDK setup or `didFinishLaunching` lands inside the window.
> 
> @NachoEmbrace would be curious to get your thoughts on this. i'm not sure how we get around this in a way that doesn't feel hack-ish

I feel like that's the intended behavior, and why I never really liked this API in the first place.
Adding attributes to an ended span is not very OTel-like behavior.
The fact that it worked that way before was an oversight on 7.0 (in 6.x the behavior is the same as in this PR).

This is the current 6.x implementation:
Note that rootSpan.SetAttribute does nothing silently if the span already ended
```
    public func addAttributesToTrace(_ attributes: [String: String]) -> Bool {

        return state.withLock {
            guard let rootSpan = $0.rootSpan else {
                return false
            }

            // rootSpan is an OTel SDK span
            // setAttribute does nothing silently if the span is ended
            attributes.forEach {
                rootSpan.setAttribute(key: $0.key, value: .string($0.value))
            }

            // TODO: Clean up reference to client! There's currently no other way to trigger a flush!
            Embrace.client?.flush(rootSpan)

            return true
        }
    }
```

@fnewberg

ferdraghi (approved):

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
```diff
@@ -312,6 +316,14 @@ extension DefaultOTelSignalsHandler: EmbraceSpanDelegate {
     }
 
     func onSpanEnded(_ span: any EmbraceSpan, endTime: Date) {
+        // A span that ended on its own must not be auto terminated when the session ends: that
+        // would stamp an error code and an error status on a span that already completed, and
+        // would report it as failed. Dropping it here also keeps the cache from holding every
+        // auto terminating span for the whole session.
```
```suggestion
        // Auto-terminating spans are dropped from the cache when they end on their own, so the cache
        // doesn't hold every one of them for the whole session.        
```

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
```diff
@@ -99,7 +99,28 @@ class DefaultEmbraceSpan: EmbraceSpan {
         }
     }
 
+    /// Returns whether this span has already ended, and therefore no longer accepts changes.
+    ///
+    /// A span that ends becomes a completed record: its status, attributes, events and links are
+    /// fixed from that point on. This mirrors the OpenTelemetry behavior, where a span stops
+    /// recording once it ends. Without it, a late change would reach the copy Embrace stores and
+    /// uploads while being silently refused by the OpenTelemetry pipeline, leaving the two
+    /// descriptions of the same span disagreeing.
```
fnewberg: nitpick

```suggestion
    /// Whether the span has ended. An ended span no longer accepts changes to its status,
    /// attributes, events or links.
    ///
    /// A span exists in two places: the record Embrace stores and uploads, and the span the OTel
    /// pipeline exports. Storage accepts any mutation it is handed. The OTel side accepts none once
    /// the span has ended, because the bridge drops it from its span cache and the OTel SDK stops
    /// recording. Refusing the change up front keeps a late write from landing in the Embrace instance
    /// while being omitted from the OTel one.    
```

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
fnewberg: @NachoEmbrace don't `endTime`, `events`, `links` and `attributes` all have unguarded setters declared  here `DefaultEmbraceSpan.swift:32-50`)? if so, should we just delete them?

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
fnewberg: This caches regardless of `endTime` and `createSpan(endTime:autoTerminationCode:)` is a valid public combination. `end()` never fires for a span that is created with end time set, so the new eviction in `onSpanEnded` (`:324`) never runs for these and they sit here until `autoTerminateSpans`'s `removeAll()` is called.

```suggestion
        if code != nil && endTime == nil {
```

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
```diff
@@ -99,7 +99,28 @@ class DefaultEmbraceSpan: EmbraceSpan {
...
+    var hasEnded: Bool {
+        state.safeValue.endTime != nil
+    }
+
+    /// Reports a change that was dropped because the span had already ended.
+    func logIgnoredMutation(_ description: String) {
```
fnewberg:
```suggestion
    private func logIgnoredMutation(_ description: String) {
```

Tests/EmbraceCoreTests/Internal/OTel/Signals/DefaultEmbraceSpanTests.swift
```diff
@@ -494,4 +504,106 @@ class DefaultEmbraceSpanTests: XCTestCase {
...
+        span.end(endTime: endTime)
+
+        // when ending it again
+        span.end(endTime: Date(timeIntervalSince1970: 20))
+
+        // then the original end time is kept and the handler is not notified again
+        XCTAssertEqual(span.endTime, endTime)
+        XCTAssertEqual(handler.onSpanEndedCallCount, 1)
+    }
+
```
fnewberg, line 523: Add a test that makes concurrent calls to `end()`

```
func test_end_concurrent_notifiesTheHandlerOnce() throws {
    // given a span
    let span = testSpan
    let endTime = Date(timeIntervalSince1970: 9)

    // when many threads end it at once
    DispatchQueue.concurrentPerform(iterations: 100) { _ in
        span.end(endTime: endTime)
    }

    // then only the first end is written through
    XCTAssertEqual(span.endTime, endTime)
    XCTAssertEqual(handler.onSpanEndedCallCount, 1)
}
```

Also, should we make `MockEmbraceSpanHandler.onSpanEndedCallCount` an `EmbraceMutex<Int>` instead of being an unsynchronized `Int` (:101-103) so that any future regression fails the assertion instead of tripping TSan?

Tests/EmbraceCoreTests/Public/OTel/DefaultOTelSignalsHandlerTests.swift
```diff
@@ -227,6 +227,33 @@ class DefaultOTelSignalsHandlerTests: XCTestCase {
...
+
+        // when the session ends and the auto termination is triggered
+        handler.autoTerminateSpans()
+
+        // then the span keeps the outcome it ended with, instead of being reported as failed
+        XCTAssertEqual(span.status, .ok)
+        XCTAssertEqual(span.endTime, endTime)
+        XCTAssertNil(span.attributes["emb.error_code"])
+    }
+
```
fnewberg, line 322: add a test that verifies a late `setAttribute` does not reach the stored copy

```
func test_setAttribute_afterEnd_doesNotReachStorage() throws {
     // given a span with an attribute set before it ended
     let span = try XCTUnwrap(handler.createSpan(name: "test"))
     span.setAttribute(key: "early", value: "value")
     span.end()

     // when setting an attribute afterwards
     span.setAttribute(key: "late", value: "value")

     // then the stored record is unchanged
     let record = try XCTUnwrap(storage.fetchSpan(id: span.context.spanId, traceId: span.context.traceId))
     XCTAssertEqual(record.attributes["early"] as? String, "value")
     XCTAssertNil(record.attributes["late"])
 }

File: Tests/EmbraceCoreTests/Internal/OTel/Signals/DefaultEmbraceSpanTests.swift
Comment (fnewberg) [inline Tests/EmbraceCoreTests/Internal/OTel/Signals/DefaultEmbraceSpanTests.swift:480]: add this test

```
func test_internalMutations_afterEnd_areIgnored() throws {
    // given a span that ended
    let span = testSpan
    span.end()

    // when the SDK stamps it afterwards
    span.setInternalAttribute(key: "emb.key", value: "value")
    let event = try span.addSessionEvent(name: "event", isInternal: true)

    // then nothing is recorded and nothing is written through
    XCTAssertNil(span.attributes["emb.key"])
    XCTAssertNil(event)
    XCTAssertEqual(handler.onSpanAttributesUpdatedCallCount, 0)
    XCTAssertEqual(handler.onSpanEventAddedCallCount, 0)
}
```

File: Tests/EmbraceCoreTests/Public/OTel/DefaultOTelSignalsHandlerTests.swift
Comment (fnewberg) [inline Tests/EmbraceCoreTests/Public/OTel/DefaultOTelSignalsHandlerTests.swift:230]: add a test that ends the parent first and verifies the parent's code still propagates:


```
func test_createSpan_autoTermination_parentCode_afterParentEnded() throws {
    // given a parent with an auto termination code that ended on its own
    let parentSpan = try XCTUnwrap(handler.createSpan(name: "parent", autoTerminationCode: .userAbandon))
    parentSpan.end()

    // when creating a child afterwards
    let child = try XCTUnwrap(handler.createSpan(name: "child", parentSpan: parentSpan))

    // and the session ends
    handler.autoTerminateSpans()

    // then the child inherits the parent's code
    XCTAssertNotNil(child.endTime)
    XCTAssertEqual(child.attributes["emb.error_code"] as? String, "user_abandon")
}
```

**Comment by fnewberg** [comment]
@NachoEmbrace  we should make `MockSpan` refuse post-end mutations, matching what the implementation now does, e.g.



```
  public var hasEnded: Bool { endTime != nil }

  public func setStatus(_ status: EmbraceSpanStatus) {
      guard !hasEnded else { return }
      self._status = status
  }

  public func addEvent(...) -> EmbraceSpanEvent? {
      guard !hasEnded else { return nil }
      ...
  }

  public func addLink(...) -> EmbraceSpanLink? {
      guard !hasEnded else { return nil }
      ...
  }

  public func end(endTime: Date) {
      guard !hasEnded else { return }
      self.endTime = endTime
      delegate?.onSpanEnded(self)
  }

  public func setAttribute(key: String, value: EmbraceAttributeValue?) {
      guard !hasEnded else { return }
      attributes[key] = value
  }
```

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
NachoEmbrace: Those are internal.
Regardless, I already removed them in https://github.com/embrace-io/embrace-apple-sdk/pull/784 since they are not used.

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
NachoEmbrace: I think errors should only be used for more important things / unrecoverable situations.
A warning log is already a lot for this IMO.

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
NachoEmbrace: Will fix this in a different way.
Casting to DefaultEmbraceSpan is fragile given that parentSpan can be of any type that follows the EmbraceSpan protocol (and autoTerminationCode is not part of the protocol).

Sources/EmbraceCore/Internal/OTel/DefaultOTelSignalsHandler+Internal.swift
NachoEmbrace: Will fix in a different way.

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
fnewberg: fair. maybe we take a pass through the SDK at some point and add counters for certain behaviors like this and just get counts. that should suffice. i mainly care about knowing what actually happens in the wild and don't need any details that a log would give you

NachoEmbrace: Given the very low limit on our internal logs, its very unlikely we get the data that way.

fnewberg:
> Adding attributes to an ended span is not very OTel-like behavior.

yeah, agreed.

fnewberg (approved):

PR #780, opened 2026-09-16
[EMBR-14064] Fix: Clearing internal log id cache

NachoEmbrace:


Sources/EmbraceOTelBridge/EmbraceOTelBridge.swift
```diff
@@ -41,9 +41,21 @@ final package class EmbraceOTelBridge {
...
+    // be empty between `createLog` calls — a non-empty set outside a `createLog` frame is a bug.
     private let internalLogIds = EmbraceMutex(Set<String>())
 
+    /// Test-only view of the IDs of the logs currently being emitted outbound.
+    ///
+    /// Outside of an in-flight `createLog` call this is always empty; tests use it to verify that
+    /// entries do not outlive the emit window they are needed for.
+    var inFlightInternalLogIds: Set<String> {
+        internalLogIds.withLock { $0 }
+    }
```
ferdraghi, line 57: kind of an odd pattern but... sure

ferdraghi (approved):

Tests/EmbraceIOTests/BridgeInternalLogIdsTests.swift
```diff
@@ -0,0 +1,83 @@
...
+/// synchronously inside `emit()`. This test drives the full public logging path to confirm that
+/// nothing is retained afterwards, and that scoping the IDs to the emit window did not break the
+/// deduplication they exist for — each log must still reach the pipeline exactly once.
+final class BridgeInternalLogIdsTests: XCTestCase {
+
+    private let logCount = 10
+    private var exporter: ThreadSafeLogExporter!
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
```
fnewberg, line 27: probably worth clearing the client here as well as in `tearDown` to avoid issues if other tests did not clear it.  not reachable today given the current class ordering, but `EmbraceIOExperimentsTests` and `PerformanceTests` both guard this in `setUp`

```suggestion
        try super.setUpWithError()
        Embrace.client = nil
```

Tests/EmbraceOTelBridgeTests/EmbraceOTelBridgeTests.swift
```diff
@@ -189,6 +189,67 @@ final class EmbraceOTelBridgeTests: XCTestCase {
...
     }
 
+    // MARK: - createLog internal ID lifetime
+
+    func test_createLog_doesNotRetainInternalLogId() {
+        bridge.createLog(MockEmbraceLog())
+        XCTAssertTrue(bridge.inFlightInternalLogIds.isEmpty)
+    }
+
+    func test_createLog_repeated_doesNotAccumulateInternalLogIds() {
```
fnewberg, line 194: does this really test anything that `test_createLog_doesNotRetainInternalLogId`  doesn't? or maybe we keep this one and drop `test_createLog_doesNotRetainInternalLogId`

Sources/EmbraceOTelBridge/EmbraceOTelBridge.swift
```diff
@@ -41,9 +41,21 @@ final package class EmbraceOTelBridge {
     // isInternalSpan correctly returns true when onStart fires synchronously during startSpan().
     private let pendingSpanIds = EmbraceMutex(Set<String>())
 
-    // Cache of IDs for logs emitted outbound so the inbound processor can skip them.
+    // IDs of logs currently being emitted outbound, so the inbound processor can skip them.
+    // Entries live only for the duration of the `emit()` call that creates them: the OTel
+    // `LogRecordProcessor` chain is invoked synchronously from `emit()`, so `isInternalLog` has
+    // already run by the time `createLog` removes the ID again. The set is therefore expected to
+    // be empty between `createLog` calls — a non-empty set outside a `createLog` frame is a bug.
```
fnewberg:
```suggestion
     // IDs of the logs the bridge is currently emitting, so `isInternalLog` can tell them apart from
     // logs created by OTel code in the host app. `EmbraceLogProcessor` is the root processor
     // registered on `loggerProvider`, and it consults `isInternalLog` at the top of its `onEmit`,
     // which the OTel SDK calls synchronously from `emit()`. An ID therefore only has to be present
     // across the `emit()` call in `createLog`, whatever the child processors do with the record.    
```

fnewberg (approved):
just some minor feedback

PR #782, opened 2026-09-17
[EMBR-14022] Feature: Exposing Embrace OTel providers

NachoEmbrace:


Examples/BrandGame/BrandGame/View/Menu/OpenTelemetry/OpenTelemetryView.swift
```diff
@@ -83,14 +83,20 @@ extension OpenTelemetryView {
...
         guard !name.isEmpty else { throw CreateTracerError.nameCannotBeEmpty }
-        return OpenTelemetry.instance.tracerProvider.get(instrumentationName: name)
+        guard let tracer = EmbraceIO.shared.tracer(instrumentationName: name) else {
+            throw CreateTracerError.embraceTracerUnavailable
+        }
+        return tracer
     }
 
+    /// Spans from this tracer are only captured by Embrace if the app enabled
+    /// `registersGlobalProviders` in its `EmbraceIO.OTelOptions`. Otherwise the process-wide
```
fnewberg, line 98: this should be added to the docs, right?

Sources/EmbraceIO/Options/EmbraceIO+OTelOptions.swift
```diff
@@ -36,6 +36,26 @@ extension EmbraceIO {
...
+        /// Determines whether the Embrace tracer and logger providers are installed as the
+        /// process-wide OpenTelemetry providers.
+        ///
+        /// When `true`, instrumentation that resolves `OpenTelemetry.instance` — including
+        /// third-party OpenTelemetry libraries that you do not configure yourself — emits through
+        /// the Embrace pipeline. When `false`, the process-wide providers are left untouched and
+        /// the Embrace providers are reachable only through `EmbraceIO.shared`.
+        ///
+        /// Defaults to `false`. Three things are worth knowing before enabling it:
+        ///
```
fnewberg, line 48:
```suggestion
        ///
        /// - It only takes effect when this `start(options:)` call is the one that creates the
        ///   SDK client. The value passed to any later call is ignored.
```

Sources/EmbraceIO/Options/EmbraceIO+OTelOptions.swift
```diff
@@ -36,6 +36,26 @@ extension EmbraceIO {
...
 
+        /// Determines whether the Embrace tracer and logger providers are installed as the
+        /// process-wide OpenTelemetry providers.
+        ///
+        /// When `true`, instrumentation that resolves `OpenTelemetry.instance` — including
+        /// third-party OpenTelemetry libraries that you do not configure yourself — emits through
+        /// the Embrace pipeline. When `false`, the process-wide providers are left untouched and
+        /// the Embrace providers are reachable only through `EmbraceIO.shared`.
+        ///
+        /// Defaults to `false`. Three things are worth knowing before enabling it:
```
fnewberg:
```suggestion
        /// Defaults to `false`. These things are worth knowing before enabling it:
```

Sources/EmbraceIO/EmbraceIO.swift
```diff
@@ -80,8 +85,15 @@ public class EmbraceIO {
...
+        //
+        // `Embrace.setup` keeps the client created by the first successful call and ignores the
+        // options — and therefore the bridge — passed to any later one. Building a bridge when a
+        // client already exists would produce a complete OTel pipeline that nothing ever emits
+        // through, so skip it and leave the existing one in place. Note this is about a client
+        // existing at all, not about it running: a client that failed to start, or was stopped,
+        // or was disabled remotely still owns the bridge the SDK emits through.
         var bridge: EmbraceOTelBridge?
-        if let otelOptions = options.otel {
+        if let otelOptions = options.otel, Embrace.client == nil {
```
fnewberg, line 96: if somebody messes up integration and calls `start` multiple times, some things will silently fail. should we

- at least log that this happened
- short circuit things above and return early if start has already been called
- add something to ensure that `start` can't be called multiple times concurrently
- something else?

Sources/EmbraceIO/EmbraceIO.swift
```diff
@@ -103,6 +115,20 @@ public class EmbraceIO {
                 metadataProvider: otel,
                 criticalResourceGroup: Embrace.client?.captureServicesGroup
             )
+
+            // Publish the bridge only now that it is fully wired. Everything reachable through
+            // the public provider accessors therefore always has a delegate attached, which
+            // closes the window in which a span could be created against a bridge that would
+            // silently drop it.
```
fnewberg:
```suggestion
            // Publish the bridge only now that it is fully wired. Child processors and exporters
            // run whether or not a delegate is attached, so publishing earlier would let a span
            // reach the app's own exporters un-stamped, with EmbraceCore never offered it.            
```

Tests/EmbraceIOTests/EmbraceIOOTelProvidersTests.swift
```diff
@@ -0,0 +1,244 @@
...
+        originalLoggerProvider = OpenTelemetry.instance.loggerProvider
+
+        mockDelegate = ProvidersMockOTelDelegate()
+        mockMetadata = ProvidersMockMetadataProvider()
+        bridge = EmbraceOTelBridge()
+        bridge.setup(delegate: mockDelegate, metadataProvider: mockMetadata)
+    }
+
+    override func tearDownWithError() throws {
+        Embrace.client = nil
```
fnewberg, line 47: since we actually start the SDK in `test_startTwice_keepsThePublishedBridgeTiedToTheClient`, do we need to stop it?
```suggestion
    override func tearDownWithError() throws {
        _ = try? Embrace.client?.stop()
        Embrace.client = nil
```

Sources/EmbraceIO/Public/EmbraceIO+OTelProviders.swift
```diff
@@ -0,0 +1,79 @@
...
+
+import Foundation
+import OpenTelemetryApi
+
+/// Access to the OpenTelemetry providers backing the Embrace pipeline.
+///
+/// Signals created through these providers are captured by Embrace and forwarded to any custom
+/// processors and exporters supplied through `EmbraceIO.OTelOptions`, exactly like the signals the
+/// SDK generates itself. They also count against the same per-session limits as
+/// `createSpan(name:)` and `log(_:severity:)`.
```
fnewberg:
```suggestion
/// SDK generates itself. They count against the same per-session limits as `EmbraceIO.shared.createSpan`
/// and `EmbraceIO.shared.log`, so they are dropped once that budget is exhausted.
```

Sources/EmbraceIO/Public/EmbraceIO+OTelProviders.swift
```diff
@@ -0,0 +1,79 @@
...
+///
+/// Every accessor here returns `nil` when the SDK was started without `EmbraceIO.OTelOptions`,
+/// because in that configuration no OpenTelemetry SDK instance is created at all. Nothing is
+/// substituted in its place: a `nil` return makes the missing configuration visible at the call
+/// site instead of silently discarding the telemetry.
```
fnewberg:
```suggestion
/// Every accessor returns `nil` when there is no Embrace OTel pipeline to hand back:
///
/// - `start(options:)` has not been called yet.
/// - The SDK was started without `EmbraceIO.OTelOptions`.
/// - `start(options:)` threw.
/// - A previous `start(options:)` already created the SDK client, so the `OTelOptions` passed to
///   any later call were ignored.
///
/// Nothing is substituted in place of the missing pipeline. An absent pipeline shows up at the
/// call site rather than quietly swallowing the telemetry.
///
/// A non-`nil` provider means the options were supplied, not that the SDK is running. The
/// providers outlive `EmbraceIO.stop()`, and signals created afterwards are stamped with an empty
/// session id and belong to no session. Consult `state` or `isSDKEnabled` if you need to know.
```

fnewberg (commented):
comment and test suggestions and one design question.

Examples/BrandGame/BrandGame/View/Menu/OpenTelemetry/OpenTelemetryView.swift
NachoEmbrace: Yes, will do once this is merged

Sources/EmbraceIO/EmbraceIO.swift
NachoEmbrace: What silently fails?
There are logs already in Embrace.setup/start
Also these already have guards so they can only be called from the main thread

fnewberg: must have missed that. so if somebody starts the SDK a second time, a log would be emitted?

NachoEmbrace: Yes. Check Embrace.setup and start (which is what EmbraceIO.start calls)

fnewberg (approved):

PR #783, opened 2026-09-17
[EMBR-14038] Fix: Fixing span id generation race condition

NachoEmbrace:
The bridge creates OTel spans on Core's behalf, then has to recognise those same spans coming back through its own `SpanProcessor` so it doesn't record them twice. It does that by pre-reserving the ID the OTel SDK is about to assign, through a custom `IdGenerator`.

That reservation lived in a **single shared slot**. Two threads creating spans at the same time overwrote each other's reservation, and three things fell out of it:

  - One span ends up with an ID nobody registered, so the bridge treats its own span as third-party: `emb.type` is overwritten with `performance` — destroying the type for session, network, view and
  startup spans — a custom-span limit slot is consumed, and a sanitized duplicate is written over the real record.
  - The other span silently takes the first one's ID.
  - The unused ID is never cleaned up, so the tracking set grows for the life of the process.

The reservation is now held per thread. A thread with no reservation of its own draws a fresh random ID and never waits, so nothing is serialized and no lock is held while calling into the OTel SDK.


Sources/EmbraceOTelBridge/EmbraceOTelBridge.swift
```diff
@@ -237,6 +247,23 @@ extension EmbraceOTelBridge: EmbraceOTelSignalBridge {
...
+    /// The reservation protocol depends on `startSpan()` consuming the reserved ID synchronously,
+    /// on the calling thread. A mismatch means that assumption no longer holds, and spans created
+    /// here would be recorded twice: once directly, and once again as if they came from outside
+    /// the SDK.
+    ///
+    /// Non-recording spans are exempt. When a sampler drops a span the SDK discards the IDs it
+    /// generated and returns a non-recording span carrying an unrelated context, so a mismatch
+    /// there is expected rather than a broken assumption.
+    private func assertReservationWasConsumed(_ reserved: SpanId, by span: any OpenTelemetryApi.Span) {
+        assert(
```
fnewberg, line 262: do we really want an `assert`?

Tests/EmbraceOTelBridgeTests/EmbraceSpanIdGeneratorTests.swift
```diff
@@ -39,75 +58,111 @@ final class EmbraceSpanIdGeneratorTests: XCTestCase {
...
-    func test_concurrentReserveAndGenerate_doesNotCrash() {
-        let iterations = 1000
-        let expectation = expectation(description: "concurrent access completes")
-        expectation.expectedFulfillmentCount = iterations * 2
+    // MARK: - Reservations are per-thread
 
-        let queue = DispatchQueue(label: "test.concurrent", attributes: .concurrent)
+    func test_reservation_isNotVisibleToOtherThreads() {
+        let otherThreadDidGenerate = expectation(description: "other thread generated an id")
+        var idFromOtherThread: SpanId?
```
fnewberg: would this be safer given access to this from different threads?

```suggestion
        let idFromOtherThread = EmbraceMutex<SpanId?>(nil)
```

Sources/EmbraceOTelBridge/EmbraceOTelBridge.swift
NachoEmbrace: Doesn't trigger in production and this assert will let us know if the OTel SDK starts behaving differently; which we need to know since this logic depends on it.

fnewberg (approved):

PR #784, opened 2026-09-17
[EMBR-14089] Fix: Concurrency clean up in DefaultEmbraceSpan

NachoEmbrace:


ferdraghi (approved):

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
```diff
@@ -30,25 +30,27 @@ class DefaultEmbraceSpan: EmbraceSpan {
     }
 
     var endTime: Date? {
-        get { state.safeValue.endTime }
-        set { state.safeValue.endTime = newValue }
+        state.safeValue.endTime
```
fnewberg: do we want this to be settable directly?

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
```diff
@@ -231,41 +242,43 @@ class DefaultEmbraceSpan: EmbraceSpan {
...
-        state.withLock {
-            $0.attributes[attribute.0] = attribute.1
+            data.attributes[attribute.0] = attribute.1
 
             if isInternal {
-                $0.internalAttributeCount += value != nil ? 1 : -1
+                data.internalAttributeCount += value != nil ? 1 : -1
             }
+
+            return (attribute, data.attributes)
```
fnewberg: `snapshot` is captured inside the lock but delivered after it, and `onSpanAttributesUpdated` reaches `storage.setSpanAttributes`, which replaces the record's entire attribute set (`EmbraceStorage+Span.swift:180`) on an async Core Data queue.

So if thread A writes `x` and snapshots `{x}` and thread B writes `y` and snapshots `{x,y}`, toth are out of the lock at that point. Then if B enqueues first and A second, the stored span ends up as `{x}` and `y` is gone.

fnewberg:
@NachoEmbrace i think we should add some test coverage that stresses the concurrency fix, e.g. tests that check for this

  - N threads calling `addLink` with the limit raised, asserting every link is present. 
  - N threads racing a limit-enforcing mock, asserting the count lands on exactly the limit and never `limit + k`. One for `addLink` and one for `addEvent`,  since the internal-event branch takes a different path.
  - A mixed test interleaving `setStatus`, `end`, `addEvent`, `addLink` and `setAttribute`, asserting none of them clobbers another.

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
```diff
@@ -148,35 +150,38 @@ class DefaultEmbraceSpan: EmbraceSpan {
...
+            }
+
         } else {
             // No handler means the span was constructed without one (e.g. read-only adapter / record).
             // Match today's silent skip — return nil.
             guard let handler else { return nil }
 
-            let currentCount = state.withLock {
-                $0.events.count - $0.internalEventCount
+            // The count the limit is checked against and the append that consumes a slot share a
```
fnewberg, line 163: this comment is largely duplicate on lines 200 and 244. we should remove two of the three places it exists in this file. not sure which location is most appropriate to keep

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
```diff
@@ -148,35 +150,38 @@ class DefaultEmbraceSpan: EmbraceSpan {
...
+            // single critical section. Reading the count under its own lock would let two concurrent
+            // callers both pass the check and push the event count past the limit.
+            event = try state.withLock {
+                let currentCount = $0.events.count - $0.internalEventCount
+
+                let event = try handler.createEvent(
```
fnewberg, line 169: does this need to be inside the lock or does only the count need to be?

fnewberg (changes requested):

Sources/EmbraceCore/Internal/OTel/Signals/DefaultEmbraceSpan.swift
NachoEmbrace: Fixed by storing 1 attribute change at a time instead of the whole dictionary

fnewberg (approved):

